PR #1, opened 2017-05-01
Add tool used to build official Docker images and manifests

MichaelSimons:
Sample repo-info.json file can be seen [here](https://github.com/MichaelSimons/dotnet-docker-nightly/blob/multi-arch/repo-info.json)
Sample output can be seen [here](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/index?buildId=715206&_a=summary)


src/image-builder/CommandType.cs
```diff
@@ -0,0 +1,8 @@
+namespace ImageBuilder
```
dagood: Copyright headers?

src/image-builder/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+WORKDIR $GOPATH
+
+# Install manifest-tool
+RUN cd $GOPATH/src \
+    && mkdir -p github.com/estesp \
+    && cd github.com/estesp \
+    && git clone https://github.com/estesp/manifest-tool \
+    && cd manifest-tool \
+    && make binary \
+    && make install
```
dagood: Is it worthwhile to clean up stuff here after the build to reduce image size?

src/image-builder/ExecuteHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+        public static void ExecuteWithRetry(
+            string fileName,
+            string args,
+            bool isDryRun,
+            string errorMessage = null,
+            string executeMessageOverride = null)
+        {
+            Execute(
+                new ProcessStartInfo(fileName, args),
+                (info) => ExecuteWithRetry(info, ExecuteProcess),
```
dagood: \<nit> Unnecessary parens around `info`.

src/image-builder/ImageBuilder.cs
```diff
@@ -0,0 +1,178 @@
...
+
+                    switch (Options.Command)
+                    {
+                        case CommandType.Build:
+                            ExecuteBuild();
+                            break;
+                        case CommandType.Manifest:
+                            ExecuteManifest();
+                            break;
+                    };
```
dagood: Unnecessary `;`

src/image-builder/ImageBuilder.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            WriteHeading("GENERATING MANIFESTS");
+            foreach (ImageInfo imageInfo in RepoInfo.Images)
+            {
+                foreach (string tag in imageInfo.Model.SharedTags)
+                {
+                    string manifestYml =
+$@"image: {RepoInfo.Model.DockerRepo}:{tag}
+manifests:
+";
```
dagood: Annoying how these break the indentation in nested `foreach`es, but not sure if it's worth avoiding.

src/image-builder/ImageBuilder.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            if (!Options.IsTestRunDisabled)
+            {
+                WriteHeading("TESTING IMAGES");
+                foreach (string command in RepoInfo.TestCommands)
+                {
+                    string[] parts = command.Split(' ');
+                    ExecuteHelper.Execute(
+                        parts[0],
+                        command.Substring(parts[0].Length + 1),
```
dagood: In VersionTools we did this kind of thing with `IndexOf` and `Substring`. Minor difference.

src/image-builder/Options.cs
```diff
@@ -0,0 +1,108 @@
...
+    public class Options
+    {
+        public const string Usage = @"Docker Image Builder
+
+Summary:  Builds all Dockerfiles detected in the current folder and sub-folders in the correct order to satisfy cross dependencies.
+
+Usage:  image-builder [options]
+
+Options:
+      --command                         Build command to execeute (build/manifest)
```
dagood: sp. `execeute`

src/image-builder/Options.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            return options;
+        }
+
+        private static string GetArgValue(string[] args, ref int i, string argName)
+        {
+            if (!IsNextArgValue(args, i))
+            {
+                throw new ArgumentException($"A '{argName}' value was not specified.{Environment.NewLine}{Usage}");
```
dagood: `$"No value specified for option '{argName}'.{Environment.NewLine}{Usage}"`?

src/image-builder/image-builder.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
+    <RootNamespace>ImageBuilder</RootNamespace>
```
dagood: Base Msft namespace?

src/image-builder/ViewModel/RepoInfo.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        private void InitializeDockerOS()
+        {
+            ProcessStartInfo startInfo = new ProcessStartInfo("docker", "version -f \"{{ .Server.Os }}\"");
+            startInfo.RedirectStandardOutput = true;
+            Process process = ExecuteHelper.Execute(startInfo, false, $"Failed to detect Docker Mode");
+            DockerOS = process.StandardOutput.ReadToEnd().Trim();
+        }
+
+        public override string ToString()
```
dagood: Are the ViewModel ToStrings formatted for something in particular, or just readability?

src/image-builder/ViewModel/RepoInfo.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.IO;
+using System.Linq;
+
+namespace ImageBuilder.ViewModel
+{
+    public class RepoInfo
+    {
+        private string DockerOS { get; set; }
+        public IEnumerable<ImageInfo> Images { get; set; }
+        public Repo Model { get; set; }
```
dagood: Looks like the `set` access modifiers are inconsistent across the viewmodels.

src/image-builder/ViewModel/PlatformInfo.cs
```diff
@@ -0,0 +1,53 @@
...
+    public class PlatformInfo
+    {
+        private static Regex FromRegex { get; } = new Regex(@"FROM\s+(?<fromImage>\S+)");
+
+        public string FromImage { get; private set; }
+        public bool IsExternalFromImage { get; private set; }
+        public Platform Model { get; private set; }
+        public IEnumerable<string> Tags { get; private set; }
+
+        public static PlatformInfo Create(Platform model, Repo repo)
```
dagood: `Create` without a private constructor? `ImageInfo` too

src/image-builder/Model/Repo.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace ImageBuilder.Model
+{
+    public class Repo
+    {
+        public string DockerRepo { get; set; }
+        public Image[] Images { get; set; }
+        public IDictionary<string, string[]> TestCommands {get;set;}
```
dagood: Whitespace in `{get;set;}`

.gitignore
```diff
@@ -0,0 +1,7 @@
+# Build output
+[Bb]in/
+[Oo]bj/
+[Oo]ut/
+
+# Visual Studio Code cache/options directory
+.vscode/
```
dagood: Is it worthwhile to add the Visual Studio ignores too?

src/image-builder/Options.cs
```diff
@@ -0,0 +1,108 @@
...
+
+Usage:  image-builder [options]
+
+Options:
+      --command                         Build command to execeute (build/manifest)
+      --dry-run                         Dry run of what images get built and order they would get built in
+  -h, --help                            Show help information
+      --password                        Password for the Docker registry the images are pushed to
+      --push                            Push built images to Docker registry
+      --repo-info                       path to json file which describes the repo
```
dleeapho: "repo-info" doesn't seem very apt. The json is more like a Docker build definition so maybe a name like "build-def" is better?

MichaelSimons: I'm not in love with `repo-info` but I don't see it as a `build-def`.  I see this as being the "source" for many things in the repo.  For example, we can use this data to generate the tag info in the readme.  The test-script can read this info to know things like the docker repo name.  Any other suggestion?

src/image-builder/CommandType.cs
MichaelSimons: Added the following to all CS files

```
// Licensed to the .NET Foundation under one or more agreements.
```

src/image-builder/Dockerfile
MichaelSimons: I am going to wait until we pickup the version of Docker that supports multi-stage builds.  This will make it so much easier.

src/image-builder/ExecuteHelper.cs
MichaelSimons: removed

src/image-builder/ImageBuilder.cs
MichaelSimons: removed

src/image-builder/Options.cs
MichaelSimons: fixed

src/image-builder/Options.cs
MichaelSimons: fixed

src/image-builder/ViewModel/RepoInfo.cs
MichaelSimons: Ensured all public properties have private setters.

src/image-builder/ViewModel/RepoInfo.cs
MichaelSimons: Nothing in particular.  I found it to be very helpful to emit this when debugging both the code and the repo-info.json.

src/image-builder/Model/Repo.cs
MichaelSimons: fixed

src/image-builder/Options.cs
dagood: `manifest` sounds reasonable to use somehow, `--manifest`, `--image-manifest`, `--image-build-manifest`? Just `--manifest` might make sense because the tool's already called `image-builder`, although you'd probably want to name the file something more specific.

src/image-builder/ViewModel/PlatformInfo.cs
MichaelSimons: added

.gitignore
MichaelSimons: I am going to wait on this one.  I don't want to make the change here and in .dockerignore without verifying the end-to-end.  If someone edits with Visual Studio I will welcome the change.

src/image-builder/image-builder.csproj
MichaelSimons: Changed to `Microsoft.DotNet.ImageBuilder` everywhere.

.gitignore
dagood: Sounds good, that can happen incrementally.

src/image-builder/ImageBuilder.cs
MichaelSimons: Agreed - I wish C# had better support for this scenario.  Per our offline discussion, I changed it to use a StringBuilder.

src/image-builder/Options.cs
dagood: Remembered that `manifest` is already used even in this code, so never mind my suggestion. I can't think of any other names at the moment.

src/image-builder/ImageBuilder.cs
MichaelSimons: changed

src/image-builder/Options.cs
MichaelSimons: Yes - I had thought about manifest but thought it added too much confusion with the docker manifest tool.

MichaelSimons: What do you think about renaming `manifest` command to `publish-manifest` and `--repo-info repo-info.json` to `--manifest manifest.json`?  The idea is that publish-manifest transforms the manifest.json into the appropriate Docker manifest format prior to publishing.

dagood: That sounds reasonable to me--they are both manifests, just different kinds. As long as it's clear when a Docker manifest is being generated/published, I think it makes sense for `manifest` to mean "a repo manifest for image-builder" in the context of this tool.

I do wonder if the filename should be more descriptive (`imagebuilder-manifest.json` or something) to distinguish from other types of manifest. That could be done on a case-by-case basis, though.

dleeapho: I think the `publish-manifest` rename scheme is better to the point where it is good enough.

dleeapho:
LGTM.

src/image-builder/Options.cs
MichaelSimons: Great dialog @dagood and @dleeapho - thanks

PR #3, opened 2017-05-12
Add support for updating the Full Description in Docker Hub.

### LINKED ISSUE #2: Add support for updating the `Full Description` in Docker Hub
The image-builder tool should provide functionality to update the `Full Description` in Docker Hub.  Without this the `Full Description` is surely going to get out of sync with the README.md stored in the GH repo.

MichaelSimons:
Fixes #2

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
@@ -44,7 +44,7 @@ namespace Microsoft.DotNet.ImageBuilder
                             PublishManifest();
                             break;
                         case CommandType.UpdateReadme:
-                            UpdateReadme().Wait();
+                            UpdateReadmeAsync().Wait();
                             break;
                     }
                 }
@@ -160,7 +160,7 @@ namespace Microsoft.DotNet.ImageBuilder
             }
         }
 
-        private static async Task UpdateReadme()
+        private static async Task UpdateReadmeAsync()
         {
             WriteHeading("UPDATING README");
 
@@ -173,7 +173,7 @@ namespace Microsoft.DotNet.ImageBuilder
             string credentials = Convert.ToBase64String(ASCIIEncoding.ASCII.GetBytes($"{Options.Username}:{Options.Password}"));
             request.Headers.Add("Authorization", $"Basic {credentials}");
 
-            JObject jsonContent = new JObject(new JProperty("full_description", new JValue(Manifest.GetReadme())));
+            JObject jsonContent = new JObject(new JProperty("full_description", new JValue(Manifest.GetReadmeContent())));
             request.Content = new StringContent(jsonContent.ToString(), Encoding.UTF8, "application/json");
 
             if (!Options.IsDryRun)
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
```diff
@@ -155,9 +160,33 @@ private static void PushImages()
             }
         }
 
+        private static async Task UpdateReadme()
```
dagood: Typical C# naming convention is that this should be `UpdateReadmeAsync`.

src/Microsoft.DotNet.ImageBuilder/Options.cs
```diff
@@ -15,7 +15,7 @@ public class Options
 Usage:  image-builder [options]
 
 Options:
-      --command                         Build command to execute (build/publishManifest)
+      --command                         Build command to execute (build/publishManifest/UpdateReadme)
```
dagood: Casing is inconsistent, I'd go with `Build/PublishManifest/UpdateReadme` and let people lowercase if they want.

src/Microsoft.DotNet.ImageBuilder/Model/Manifest.cs
```diff
@@ -10,6 +10,7 @@ public class Manifest
     {
         public string DockerRepo { get; set; }
         public Image[] Images { get; set; }
+        public string Readme { get; set; }
```
dagood: Hmm, not clear to me that this is a path to a readme file until I read the `ManifestInfo` getter/reader. `ReadmePath`?

src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestInfo.cs
```diff
@@ -51,6 +51,11 @@ public IEnumerable<string> GetPlatformTags()
                 .SelectMany(image => image.Platform.Tags);
         }
 
+        public string GetReadme()
```
dagood: Maybe this could be `GetReadmeContent` to distinguish from the path `Readme` in the model.

src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
MichaelSimons: Of course - fixed.

src/Microsoft.DotNet.ImageBuilder/Options.cs
MichaelSimons: I like your suggestion - changed

src/Microsoft.DotNet.ImageBuilder/Model/Manifest.cs
MichaelSimons: Good point - changed.

src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestInfo.cs
MichaelSimons: Changed

PR #4, opened 2017-05-18
Changes needed to build the aspnet-docker images

MichaelSimons:
- Refactored to support building images that are from multiple Docker Hub repos.
- Refactored manifest model to validate required properties.
- Refactored code to handle optional manifest properties.
- Fixed issue related to test commands that have no arguments.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestInfo.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestInfo.cs
@@ -14,8 +14,10 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
     public class ManifestInfo
     {
         private string DockerOS { get; set; }
-        public IEnumerable<RepoInfo> Repos { get; private set; }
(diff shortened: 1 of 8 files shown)


MichaelSimons:
Related to https://github.com/aspnet/aspnet-docker/pull/235

src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestInfo.cs
```diff
@@ -37,23 +37,23 @@ public static ManifestInfo Create(string repoJsonPath)
...
-                .Select(image => ImageInfo.Create(image, manifestInfo.DockerOS, manifestInfo.Model))
+            manifestInfo.Repos = manifestInfo.Model.Repos
+                .Select(image => RepoInfo.Create(image, manifestInfo.DockerOS))
                 .ToArray();
 
             return manifestInfo;
         }
 
-        public IEnumerable<string> GetPlatformTags()
+        public IEnumerable<ImageInfo> GetAllImages()
```
MichaelSimons: Changed to a property and put the initialization logic in the factory method.

PR #29, opened 2017-07-20
Refactor pull logic and rename view model `Tag` properties

### LINKED ISSUE #17: ImageBuilder tag terminology should be more clear
Within ImageBuilder the same tag nomenclature is used for the Model/ViewModel members however they behave slightly different which causes some confusion.  Within the Model, the tag is the short name while within the ViewModel the tag is the fully qualified name.  The members should be renamed within the ViewModel to reflect this difference.

### LINKED ISSUE #27: Pull logic should be refactored to pull each image at most once during a build
Image-Builder has logic that will explicitly pull the FROM image when building.  This ensures the images get built with the latest base image.  There is a potential problem with this logic.

Dockerfile1 -> FROM official-image-xyz
Dockerfile2 -> FROM official-image-xyz

Currently if you use Image-Builder to build these the following happens

docker pull official-image-xyz
docker build Dockerfile1
docker pull official-image-xyz
docker build Dockerfile2

Normally the docker pull resolves to the same image but it is possible that official-image-xyz could get updated during the build in which the second pull would get a different image.  This is bad as a single build should get built with a common set of dependencies.  

Image-Builder should be modified to compute all dependent images into a unique list and pull them once.

MichaelSimons:
This fixes #27 and fixes #17

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestInfo.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestInfo.cs
@@ -51,8 +51,7 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
             manifestInfo.ActiveImages = manifestInfo.Images
                 .Where(image => image.ActivePlatform != null)
                 .ToArray();
-            manifestInfo.ActivePlatformFullyQualifiedTags = manifestInfo.Images
-                .Where(image => image.ActivePlatform != null)
+            manifestInfo.ActivePlatformFullyQualifiedTags = manifestInfo.ActiveImages
                 .SelectMany(image => image.ActivePlatform.FullyQualifiedTags)
                 .ToArray();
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestInfo.cs
```diff
@@ -47,9 +48,12 @@ private ManifestInfo()
             manifestInfo.Images = manifestInfo.Repos
                 .SelectMany(repo => repo.Images)
                 .ToArray();
-            manifestInfo.ActivePlatformTags = manifestInfo.Images
+            manifestInfo.ActiveImages = manifestInfo.Images
                 .Where(image => image.ActivePlatform != null)
-                .SelectMany(image => image.ActivePlatform.Tags)
+                .ToArray();
+            manifestInfo.ActivePlatformFullyQualifiedTags = manifestInfo.Images
+                .Where(image => image.ActivePlatform != null)
```
dagood: Why not base this on `manifestInfo.ActiveImages`?

MichaelSimons: Good point - changed.

PR #30, opened 2017-07-24
Add support for changing the repo owner via options

MichaelSimons:
Looking for feedback on the new option.  I can imagine a scenario in which specifying an entirely different repo may be useful.  That would require the mapping of old repo to new repo since the manifest can contain multiple repos.  The repo owner functionality I added satisfies the current requirements from the dotnet core repos.  Having to specify a map makes it clunky for this scenario which is why I didn't add it.  Adding another option down the road to specify a completely different repo is certainly possible.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
@@ -12,6 +12,7 @@ using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
 namespace Microsoft.DotNet.ImageBuilder
@@ -62,6 +63,7 @@ namespace Microsoft.DotNet.ImageBuilder
 
         private static void Build()
         {
+            PullBaseImages();
             BuildImages();
             RunTests();
             PushImages();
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
```diff
@@ -70,54 +72,26 @@ private static void Build()
...
+                finally
+                {
+                    if (createdPrivateDockerfile)
                     {
-                        filename = command.Substring(0, firstSpaceIndex);
-                        args = command.Substring(firstSpaceIndex + 1);
+                        File.Delete(dockerfilePath);
                     }
-
-                    ExecuteHelper.Execute(filename, args, Options.IsDryRun);
```
dagood: Are all the changes in the file for this PR? There are some things here that look like unrelated improvements.

MichaelSimons: Apologies, there were some methods that were not ordered correctly so I reordered them. The BuildImages method was getting too busy to I split it into separate methods but I consider those to be related.

src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
```diff
@@ -191,6 +181,72 @@ private static void PushImages()
...
+            dockerfilePath = Path.Combine(image.ActivePlatform.Model.Dockerfile, "Dockerfile");
+
+            // If an alternative repo owner was specified, update the intra-repo FROM commands.
+            bool updateDockerfile = !string.IsNullOrWhiteSpace(Options.RepoOwner)
+                && !image.ActivePlatform.FromImages.All(Manifest.IsExternalImage);
+            if (updateDockerfile)
+            {
+                string dockerfileContents = File.ReadAllText(dockerfilePath);
+
+                foreach (string fromImage in image.ActivePlatform.FromImages.SkipWhile(Manifest.IsExternalImage))
```
dagood: `SkipWhile` doesn't seem like what's intended here... I think you'll need `Where(x => !Manifest.IsExternalImage(x))`.

src/Microsoft.DotNet.ImageBuilder/Options.cs
```diff
@@ -18,6 +18,7 @@ public class Options
 Usage:  image-builder [options]
 
 Options:
+      --repo-owner                      An alternative repo owner which overrides what is specified in the manifest
```
dagood: Alphabetize?

src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
```diff
@@ -70,54 +72,26 @@ private static void Build()
...
             foreach (ImageInfo image in Manifest.ActiveImages)
             {
-                ExecuteHelper.Execute(
-                    "docker",
-                    $"build -t {string.Join(" -t ", image.ActiveFullyQualifiedTags)} {image.ActivePlatform.Model.Dockerfile}",
-                    Options.IsDryRun);
-            }
-        }
+                string dockerfilePath;
+                bool createdPrivateDockerfile = UpdateDockerfileFromCommands(image, out dockerfilePath);
```
MichaelSimons: I agree passing the Dockerfile as stdin would be ideal.  Unfortunately if you pass in the Dockerfile as stdin then there is no build context.  ASPNETCORE has files that must be included in the context therefore -f is the only option I can think of that would work.

src/Microsoft.DotNet.ImageBuilder/Options.cs
MichaelSimons: Yes - I renamed this and forgot the alphabetize - fixed.

src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
MichaelSimons: you are correct - changed

src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
dagood: Ah, unfortunate but it makes sense.

src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
```diff
@@ -191,6 +181,74 @@ private static void PushImages()
...
+                        filename = command.Substring(0, firstSpaceIndex);
+                        args = command.Substring(firstSpaceIndex + 1);
+                    }
+
+                    ExecuteHelper.Execute(filename, args, Options.IsDryRun);
+                }
+            }
+        }
+
+        private static bool UpdateDockerfileFromCommands(ImageInfo image, out string dockerfilePath)
```
dagood: The `bool` being the return value throws me off a little--makes it look like a `TryX` method but it's really not at all. Swapping them would help me, but I might have bad expectations. A factoring that doesn't use `out` would be appealing... I regret using it so much in VersionTools because it made some additions very awkward.

Also, isn't very clear to me what the return value means from the method name.

MichaelSimons: I apologize for my premature merge 😞.  I acknowledge the point you bring up and agree avoiding the `out` would be appealing.  IMO swapping the args doesn't address the problem.  I had thought about returning the dockerfile path only when re-written but that makes the calling code more complex in that it has to either re-calculate the path or conditionally specify the -f option.  Thinking...

dagood: Np, I think it's clear enough what it's doing to maintain it, just could be clearer.

Going a bit crazy with it, this could be solved by returning something `IDisposable` that provides a dockerfile path, and not even revealing to the caller whether or not it's a temp one.

PR #31, opened 2017-07-28
ToLower arch when creating manifest for publish

MichaelSimons:
This change is needed because the Docker manifest tool used for publishing the multi-arch tags is case sensitive.


src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
```diff
@@ -115,7 +115,7 @@ private static void PublishManifest()
                             manifestYml.AppendLine($"  -");
                             manifestYml.AppendLine($"    image: {repo.Name}:{platformTag}");
                             manifestYml.AppendLine($"    platform:");
-                            manifestYml.AppendLine($"      architecture: {platform.Architecture}");
+                            manifestYml.AppendLine($"      architecture: {platform.Architecture.ToString().ToLower()}");
```
dagood: `ToLowerInvariant`?

MichaelSimons: Would be safer in general but probably never needed in practice for this usage.  Changing.

PR #32, opened 2017-07-31
Refactor commands into separate components

### LINKED ISSUE #18: ImageBuilder command line args should be linked to the command they are for
Currently the ImageBuilder command line parsing is pretty simple.  As a consequence the options are not linked to the commands they are for.  For example the --skip-test option is only valid for the build command.

```
dotnet run -- -h
Docker Image Builder

Summary:  Builds all Dockerfiles detected in the current folder and sub-folders in the correct order to satisfy cross dependencies.

Options:
      --command                         Build command to execute (Build/PublishManifest/UpdateReadme)
      --dry-run                         Dry run of what images get built and order they would get built in
  -h, --help                            Show help information
      --manifest                        path to json file which describes the repo
      --password                        Password for the Docker registry the images are pushed to
      --push                            Push built images to Docker registry
      --skip-pulling                    Skip explicitly pulling the base images of the Dockerfiles
      --skip-test                       Skip running the tests
      --username                        Username for the Docker registry the images are pushed to
```

Using System.CommandLine (CoreFxLab) or Microsoft.Extensions.CommandLine (ASPNET common) would help solve this problem.

MichaelSimons:
The benefit of this refactoring is two fold.
1. This architecture makes it easier to add/remove commands.  This is because the logic and corresponding options for each command are self-contained within command specific classes.
1. The help text displayed to the end user is now in terms of commands.  The options that are display are displayed only in the context of the commands they are for.

Fixes #18


src/Microsoft.DotNet.ImageBuilder/NuGet.config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab/" />
```
dagood: Should there be a `ProjectReference` in this PR to get System.CommandLine from here?

src/Microsoft.DotNet.ImageBuilder/Commands/BuildOptions.cs
```diff
@@ -0,0 +1,75 @@
...
+        protected override string CommandName { get; } = "build";
+
+        public Architecture Architecture { get; set; }
+        public bool IsPushEnabled { get; set; }
+        public bool IsSkipPullingEnabled { get; set; }
+        public bool IsTestRunDisabled { get; set; }
+        public string Path { get; set; }
+        public IDictionary<string, string> TestVariables { get; set; }
+
+        public BuildOptions() : base()
```
dagood: None of the commands do anything in the explicit constructor, so they don't seem worthwhile to add to all of them.

src/Microsoft.DotNet.ImageBuilder/Commands/BuildOptions.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.CommandLine;
+using System.Collections.ObjectModel;
+using System;
+using System.Linq;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class BuildOptions : Options
+    {
+        protected override string CommandHelp { get; } = "Builds and Tests Dockerfiles";
```
dagood: `protected override string CommandHelp => "Builds and Tests Dockerfiles";` instead?

src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifesFilter.cs
```diff
@@ -0,0 +1,42 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.ImageBuilder.Model;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.DotNet.ImageBuilder.ViewModel
+{
+    public class ManifestFilter
```
dagood: File name has a typo.

src/Microsoft.DotNet.ImageBuilder/ViewModel/ImageInfo.cs
```diff
@@ -41,12 +35,7 @@ private ImageInfo()
                     .ToArray();
             }
 
-            imageInfo.ActivePlatform = model.Platforms
-                .Where(platform => platform.OS == dockerOS && platform.Architecture == dockerArchitecture)
-                .Where(platform => string.IsNullOrWhiteSpace(includePath) || platform.Dockerfile.StartsWith(includePath))
-                .Select(platform => PlatformInfo.Create(platform, manifest, repoName))
-                .SingleOrDefault();
-
+            imageInfo.ActivePlatform =  PlatformInfo.Create(manifestFilter.GetPlatform(model), manifest, repoName);
```
dagood: Doesn't look like this preserves behavior: originally, if no platforms were found, `ActivePlatform = null`, but now `ActivePlatform = PlatformInfo.Create(null, ...)` which looks like it'll always throw a null ref exception.

I think it would make sense to change `GetPlatform` to `GetActivePlatforms` (more similar to the other `ManifestFilter` methods that way, too) and enforce the `SingleOrDefault` out here.

src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
```diff
@@ -2,53 +2,41 @@
...
             {
-                Options = Options.ParseArgs(args);
-                if (Options.IsHelpRequest)
-                {
-                    Console.WriteLine(Options.Usage);
-                }
-                else
-                {
-                    ReadManifest();
+                ICommand[] commands = new ICommand[] {
```
dagood: I think `new ICommand[]` is redundant.

src/Microsoft.DotNet.ImageBuilder/Commands/BuildOptions.cs
MichaelSimons: Changed

src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifesFilter.cs
MichaelSimons: Fixed

src/Microsoft.DotNet.ImageBuilder/Commands/BuildOptions.cs
MichaelSimons: My preference is to explicitly declare ctors.  IMO this is similar to explicitly using an access modifier.  It helps call out what the API is of for every type.

src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
MichaelSimons: Removed

src/Microsoft.DotNet.ImageBuilder/ViewModel/ImageInfo.cs
MichaelSimons: Good catch.  IMO the SingleOrDefault better belongs in the filter because it is the one that knows there is always at most one Platform (in a valid manifest) given the parameters it filters on.  Having the caller know this would break the encapsulation of the filter logic within a single class.

Given this, I changed the logic to:
```
Platform activePlatformModel = manifestFilter.GetPlatform(model);
if (activePlatformModel != null)
{
    imageInfo.ActivePlatform = PlatformInfo.Create(activePlatformModel, manifest, repoName);
    imageInfo.ActiveFullyQualifiedTags = imageInfo.SharedFullyQualifiedTags
        .Concat(imageInfo.ActivePlatform.FullyQualifiedTags);
}
```

dagood: Makes sense, that sounds good.

Issue #10, opened 2017-06-09
Create tool that generates the tag section of the readme from the manifest


MichaelSimons:
1. Will help ensure the tag section of the various readmes are the same.  
1. Will help ensure all of the tags are documented and the links are valid.
1. Could be automated to run as part of the build to ensure the readme is always up to date.

PR #34, opened 2017-08-03
Add command to generate tags section of readme

MichaelSimons:
This required a change to the manifest schema in order to capture the required metadata.  I applied the needed [changes](https://github.com/MichaelSimons/dotnet-docker-nightly/commit/39741d5a206239d795e732087f367d6cf18ce472#diff-4b1eb3dc48c4e16d49db5b42298fe654) to the microsoft/dotnet-nightly repo if you want to take a look.  I'm not crazy about the verbosity of the tags element but I can see benefits to being able to add additional metadata over time.  I'm certainly open to feedback.

related #10

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
@@ -1,0 +1,79 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.ImageBuilder.Model;
+using Microsoft.DotNet.ImageBuilder.ViewModel;
+using System;
+using System.Linq;
+using System.Text;
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class GenerateTagsReadmeCommand : Command<GenerateTagsReadmeOptions>
+    {
+        public GenerateTagsReadmeCommand() : base()
+        {
+        }
+
+        public override void Execute()
+        {
+            WriteHeading("GENERATING TAGS README");
+            foreach (RepoInfo repo in Manifest.Repos)
+            {
+                StringBuilder tagsReadme = new StringBuilder();
+
+                var platformGroups = repo.Images
+                    .OrderBy(image => image.Model.ReadmeOrder)
+                    .SelectMany(image => image.Platforms.Select(platform => new { Image = image, Platform = platform }))
+                    .GroupBy(tuple => new { tuple.Platform.Model.OS, tuple.Platform.Model.Architecture });
+
+                foreach (var platformGroup in platformGroups)
+                {
+                    string os = platformGroup.Key.OS.Substring(0, 1).ToUpper() + platformGroup.Key.OS.Substring(1);
+                    string arch = GetArchitectureDisplayName(platformGroup.Key.Architecture);
+                    tagsReadme.AppendLine($"# Supported {os} {arch} tags");
+                    tagsReadme.AppendLine();
+                    foreach (var tuple in platformGroup)
+                    {
+                        string tags = GetDocumentedTags(tuple.Platform.Tags)
+                            .Concat(GetDocumentedTags(tuple.Image.SharedTags))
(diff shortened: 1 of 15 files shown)


dagood:
My first impression of the schema is that this:

```json
"tags": [
  {
    "name": "1.0.5-runtime-deps-jessie"
  },
  {
    "name": "1.0.5-core-deps",
    "isUndocumented": true
  },
  {
    "name": "1.0-core-deps",
    "isUndocumented": true
  }
]
```

Could be this:

```json
"tags": {
  "1.0.5-runtime-deps-jessie": {},
  "1.0.5-core-deps": {
    "isUndocumented": true
  },
  "1.0-core-deps": {
    "isUndocumented": true
  }
}
```

Or will there ever be a reason to add the same tag name twice?

MichaelSimons:
@dagood - interesting suggestion - I see tag names as needing to be unique - let me play with it.

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+
+        private IEnumerable<string> GetDocumentedTags(IEnumerable<TagInfo> tagInfos)
+        {
+            return tagInfos.Where(tag => !tag.Model.IsUndocumented)
+                .Select(tag => tag.Name);
+        }
+
+        private string GetArchitectureDisplayName(Architecture architecture)
```
dagood: Can these helpers be `static`?

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+            base.ParseCommandLine(syntax);
+
+            string sourceUrl = null;
+            string sourceUrlName = "source-url";
+            Argument sourceUrlArg = syntax.DefineOption(sourceUrlName, ref sourceUrl, true, "Base URL of the Dockerfile sources");
+            SourceUrl = sourceUrl;
+
+            if (!sourceUrlArg.IsSpecified)
+            {
+                syntax.ReportError($"`{sourceUrlName}` must be specified.");
```
MichaelSimons: There is a reason this library is still in corefxlab 😄  The isRequired arg indicates is the option is specified, then a value must be specified.  There is no built in behavior to enforce a required option or parameter.

src/Microsoft.DotNet.ImageBuilder/Model/Image.cs
```diff
@@ -3,17 +3,18 @@
...
 namespace Microsoft.DotNet.ImageBuilder.Model
 {
     public class Image
     {
-        public string[] SharedTags { get; set; }
-
         [JsonProperty(Required = Required.Always)]
         public Platform[] Platforms { get; set; }
 
+        public int ReadmeOrder { get; set; }
```
MichaelSimons: The order of the images is the order they get built in.  This order is necessary for intra-repo references.  This order could be calculated with the from statements and utilizing a t-sort.  I didn't implement the ordering logic originally as the manifest order seemed sufficient and I preferred using the manifest to control the order.  My concern with implementing the build order logic and then using the order for the readme, is what will be the next order related requirement that arises?

dagood: Ah, I'm fine with this then, didn't realize the order already had meaning.

MichaelSimons:
@dagood, I refactored the tag schema as you suggested - https://github.com/dotnet/docker-tools/pull/34/commits/5547af26582b5a075d40baabef1d055cb0775f79.  You can take a look at the update manifest [here](https://github.com/MichaelSimons/dotnet-docker-nightly/commit/f17f91f342d301d4020cd46cb28b34cee6c82b17).

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeOptions.cs
MichaelSimons: I followed up on this with corefxlab.  Options are by definition supposed to be optional.  That being said, there is no built in functionality to enforce parameters are required as well - https://github.com/dotnet/corefxlab/issues/1689.  Given that Options are intended to be optional, I changed this to a parameter.

PR #35, opened 2017-08-07
Fixup manifest generation regression to include fully qualified image name

MichaelSimons:


MichaelSimons:
Merging to fixup blocked builds.

Issue #16, opened 2017-06-09
Utilize multi-step Dockerfiles for ImageBuilder


MichaelSimons:
This will greatly reduce the size of the image because it is currently a an 'sdk' based image.

PR #43, opened 2017-08-17
Update Dockerfiles to be multi-staged

MichaelSimons:
Related to #16 

This reduces the image sizes substantially.
Linux => 1.81GB to 398 MB
Windows => 2.83GB to 1.14 GB

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/Dockerfile
+++ b/src/Microsoft.DotNet.ImageBuilder/Dockerfile
@@ -2,12 +2,29 @@
 #     docker build -t image-builder .
 #     docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v <local path to build>:/repo -w /repo image-builder <image-build args>
 
-FROM microsoft/dotnet:2.0-sdk
+FROM microsoft/dotnet:2.0-sdk AS build-env
 
-# Install Docker
+# build Microsoft.DotNet.ImageBuilder
+WORKDIR image-builder
+
+# copy csproj and restore as distinct layers
+COPY *.csproj ./
+COPY NuGet.config ./
+RUN dotnet restore
+
+# copy everything else and build
+COPY . ./
+RUN dotnet publish -c Release -o out
+
+
+# build runtime image
+FROM microsoft/dotnet:2.0-runtime
+
+# install Docker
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/Dockerfile
```diff
@@ -2,12 +2,29 @@
...
 #     docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v <local path to build>:/repo -w /repo image-builder <image-build args>
 
-FROM microsoft/dotnet:2.0-sdk
+FROM microsoft/dotnet:2.0-sdk AS build-env
 
-# Install Docker
+# build Microsoft.DotNet.ImageBuilder
+WORKDIR image-builder
+
+# copy csproj and restore as distinct layers
```
dagood: When I read this at first I thought it meant copying the csproj being distinct from the `dotnet restore`, but it really means this group of three commands being distinct from copying the source files, right? (For dev rebuild efficiency.) Seems the comment could be clarified.

src/Microsoft.DotNet.ImageBuilder/Dockerfile
```diff
@@ -2,12 +2,29 @@
...
+
+
+# build runtime image
+FROM microsoft/dotnet:2.0-runtime
+
+# install Docker
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         apt-transport-https \
+        gnupg \
```
dagood: Just curious what this is for.

MichaelSimons: This is a docker-ce prereq.  Before this change it was coming from the base image chain.  specifically it is part of the buildpack-deps image which is the base image for the .NET Core SDK image

src/Microsoft.DotNet.ImageBuilder/Dockerfile
MichaelSimons: Is this better?  `restore packages before coping entire source - provides optimizations when rebuilding`

dagood: Perfect, with `coping` -> `copying`. :smile:

PR #48, opened 2017-09-21
Display version of Docker being used.

ravimeda:
Display version of Docker using -

```C#
ExecuteHelper.ExecuteWithRetry(fileName: "docker", args: "version", isDryRun: false);
```

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
@@ -46,6 +46,7 @@ namespace Microsoft.DotNet.ImageBuilder
 
                 if (argSyntax.ActiveCommand != null)
                 {
+                    ExecuteHelper.ExecuteWithRetry("docker", "version", false);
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.ImageBuilder/ImageBuilder.cs
```diff
@@ -46,6 +46,7 @@ public static int Main(string[] args)
 
                 if (argSyntax.ActiveCommand != null)
                 {
+                    ExecuteHelper.ExecuteWithRetry(fileName: "docker", args: "version", isDryRun: false);
```
MichaelSimons: Execute would be preferable over ExecuteWithRetry.  `docker version` doesn't make any web calls and therefore is not prone to intermittent failures.  If `docker version` fails it would mean docker is not configured correctly and retrying isn't going to pass rather it will just waste time.

It would be valuable to capture why this call exists.
`// Capture the docker version used in the output`

It might help to also add a blank line after to separate this from the logic that executes the command.

PR #49, opened 2017-10-04
Add regex support on path argument in ImageBuilder

ravimeda:
These changes add regex support on _path_ argument in _ImageBuilder_. Wildcard characters `*` and `?` are supported. For example, passing `build --repo microsoft/dotnet-nightly --path "*sdk*"` would build all images that contain the word `sdk` in Dockerfile path.

Changes also include -

- A solution file to open ImageBuilder project in Visual Studio
- Remove and sort _using_ across the solution

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/Model/Tag.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/Model/Tag.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-
 namespace Microsoft.DotNet.ImageBuilder.Model
 {
     public class Tag
--- a/src/Microsoft.DotNet.ImageBuilder/Model/Tag.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/Model/Tag.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Newtonsoft.Json;
 
 namespace Microsoft.DotNet.ImageBuilder.Model
 {
--- a/src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestFilter.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestFilter.cs
@@ -31,7 +31,7 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
         {
             bool isPathBlank = string.IsNullOrWhiteSpace(IncludePath);
             string pattern = isPathBlank ? 
-                string.Empty : "^" + Regex.Escape(IncludePath).Replace(@"\*", ".*").Replace(@"\?", ".");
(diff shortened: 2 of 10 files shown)


src/Microsoft.DotNet.ImageBuilder/Model/Tag.cs
```diff
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Newtonsoft.Json;
```
MichaelSimons: Extra blank line should be removed.

src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestFilter.cs
```diff
@@ -29,7 +30,11 @@ public Platform GetActivePlatform(Image image)
         public IEnumerable<Platform> GetPlatforms(Image image)
         {
             return image.Platforms
-                .Where(platform => string.IsNullOrWhiteSpace(IncludePath) || platform.Dockerfile.StartsWith(IncludePath));
+                .Where(platform => string.IsNullOrWhiteSpace(IncludePath)
+                    || Regex.IsMatch(
+                        platform.Dockerfile,
+                        "^" + Regex.Escape(IncludePath).Replace(@"\*", ".*").Replace(@"\?", "."),
```
MichaelSimons: This should be pulled outside of the where predicate so that it is not executed for each platform.  When pulled out you will need to account for the IncludePath not being specified.

src/Microsoft.DotNet.ImageBuilder/Model/Tag.cs
ravimeda: Updated.

src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestFilter.cs
ravimeda: Good one! Updated as -

```C#
        public IEnumerable<Platform> GetPlatforms(Image image)
        {
            bool isPathBlank = string.IsNullOrWhiteSpace(IncludePath);
            string pattern = isPathBlank ? 
                string.Empty : "^" + Regex.Escape(IncludePath).Replace(@"\*", ".*").Replace(@"\?", ".");

            return image.Platforms
                .Where(platform => isPathBlank || Regex.IsMatch(platform.Dockerfile, pattern, RegexOptions.IgnoreCase));
        }
```

MichaelSimons: Consider renaming `pattern` to `includePattern` as this is more descriptive.  Also I think setting the pattern to null when the path is not specified would be more appropriate as that would cause an exception when used which would help flush out any coding bugs.

ravimeda: Thanks Michael. Updated accordingly.

ravimeda:
Thanks!

PR #50, opened 2017-10-05
Include end of line anchor in regex for path.

ravimeda:
Adding `$` anchor in regex pattern for path ensures full match of Dockerfile path string.


MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker-nightly/issues/399

PR #51, opened 2017-10-05
Add support for multiple os images per multi-arch tag

MichaelSimons:
These changes allow for manifest tag to reference to more than one image per platform which will be for different OS versions.

This functionality is need as part of https://github.com/dotnet/dotnet-docker/issues/298.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/Commands/BuildCommand.cs
@@ -39,8 +39,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             {
                 foreach (PlatformInfo platform in image.ActivePlatforms)
                 {
-                    string dockerfilePath;
-                    bool createdPrivateDockerfile = UpdateDockerfileFromCommands(platform, out dockerfilePath);
+                    bool createdPrivateDockerfile = UpdateDockerfileFromCommands(platform, out string dockerfilePath);
 
                     try
                     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.ImageBuilder/Commands/BuildCommand.cs
```diff
@@ -35,22 +37,34 @@ private void BuildImages()
...
-                    ExecuteHelper.Execute(
-                        "docker",
-                        $"build {tagArgs} -f {dockerfilePath} {image.ActivePlatform.BuildContextPath}",
-                        Options.IsDryRun);
-                }
-                finally
-                {
-                    if (createdPrivateDockerfile)
+                    string dockerfilePath;
+                    bool createdPrivateDockerfile = UpdateDockerfileFromCommands(platform, out dockerfilePath);
```
ravimeda: Inline the declaration - `out string dockerfilePath`

MichaelSimons: Fixed

src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestFilter.cs
```diff
@@ -20,11 +21,11 @@ public ManifestFilter()
...
         }
 
-        public Platform GetActivePlatform(Image image)
+        public IEnumerable<Platform> GetActivePlatforms(Image image)
         {
             return GetPlatforms(image)
-                .Where(platform => platform.OS == DockerOS && platform.Architecture == DockerArchitecture)
-                .SingleOrDefault();
+                .Where(platform => string.Equals(platform.OS, DockerOS, StringComparison.OrdinalIgnoreCase)
+                    && platform.Architecture == DockerArchitecture);
```
ravimeda: Ignore case on `platform.Architecture == DockerArchitecture`?

MichaelSimons: Architecture is an enum 😄

ravimeda: I need some coffee I guess :)

PR #53, opened 2017-10-13
Update Docker login logic and add verbose option

### LINKED ISSUE #52: Docker login logic should be updated to use the password-stdin option
The current Docker login logic is using the `password` option which is insecure.  At the time the logic was written, it was the only supported option.  The `password-stdin` option is now the preferred way.

MichaelSimons:
Fixes #52

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/Commands/Options.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/Commands/Options.cs
@@ -15,6 +15,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         protected abstract string CommandName { get; }
 
         public bool IsDryRun { get; set; }
+        public bool IsVerbose { get; set; }
         public string Manifest { get; set; }
         public string Repo { get; set; }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.ImageBuilder/Commands/Options.cs
```diff
@@ -46,6 +47,10 @@ public virtual void ParseCommandLine(ArgumentSyntax syntax)
             string repo = null;
             syntax.DefineOption("repo", ref repo, "Repo to operate on (Default is all)");
             Repo = repo;
+
+            bool isVerbose = false;
```
ravimeda: Minor - since the default for a bool is `false`, this initialization is not needed.

ravimeda:
Thanks for finding `--password-stdin`

src/Microsoft.DotNet.ImageBuilder/Commands/Options.cs
MichaelSimons: Because the variable is passed as a ref parameter it must be explicitly initialized.

PR #54, opened 2017-10-19
Add support for OS Version in manifest

MichaelSimons:
Fixes https://github.com/dotnet/dotnet-docker-samples/issues/90

This [PR](https://github.com/dotnet/dotnet-docker-samples/compare/nano-1709?expand=1#diff-4b1eb3dc48c4e16d49db5b42298fe654) illustrates what the manifest would look like with this change.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestFilter.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestFilter.cs
@@ -15,6 +15,7 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
         public Architecture DockerArchitecture { get; set; } = DockerHelper.Architecture;
         public string DockerOS { get; } = DockerHelper.GetOS();
         public string IncludeRepo { get; set; }
+        public string IncludeOsVersion { get; set; }
         public string IncludePath { get; set; }
 
         public ManifestFilter()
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.ImageBuilder/ViewModel/ManifestFilter.cs
```diff
@@ -28,15 +29,22 @@ public IEnumerable<Platform> GetActivePlatforms(Image image)
                     && platform.Architecture == DockerArchitecture);
         }
 
+        private string GetFilterRegexPattern(string pattern)
+        {
+            return pattern == null ? null : $"^{Regex.Escape(pattern).Replace(@"\*", ".*").Replace(@"\?", ".")}$";
```
dagood: The `.*` stand in doesn't look quite like typical file globbing `*` because `.*` will accept `foo/bar`. No problem for os version, though.

Just noticed that this is only moved, not new, anyway.

MichaelSimons: yes you are correct - for better or worse the * was intended as any characters including path separators.  It was the simplest way that came to mind to support matching on all 2.0 paths yet allow things like 2.0/*/nanoserver-1709 where * matches all paths to nanoserver-1709 which may occur at multiple depths.

dagood: If it was `**` being replaced, it would be more typical, but then it'd be weird not to support `*`, which is somewhat more complicated here. I really doubt anyone would rely on `*` *not* going to subdirs, so I'm fine with how it is. :smile:

MichaelSimons: It is a balance with keeping it simple, following "typical" conventions, and meeting the requirements.  In this case 1 and 3 won out over 2 😃

PR #55, opened 2017-10-20
Add support for logging in across multiple versions of Docker

MichaelSimons:
The [changes ](https://github.com/dotnet/docker-tools/commit/1452ac24c24f7dae94e4990792c269574ad66719) recently made to make use of the newish Docker login `--password-stdin` option can't be the only login option as image-builder is being used on machines with older versions of docker.  This change adds support to detect which version exists and use the appropriate login options.


MichaelSimons:
@ravimeda - merging to unblock work in https://github.com/dotnet/dotnet-docker-samples.  If you have feedback, please let me know and I will make the appropriate changes.

Issue #58, opened 2017-11-07
Update GenerateTagsReadme command to update the README.md


MichaelSimons:
Currently the GenerateTagsReadme command emits the tags documentation section of the readme to the console.  Logic should be added to optionally update the actual readme.  This would make the workflow for updating the readme easier/quicker.

The only reason this wasn't done originally was due to time constraints.  This functionality has proven to be pretty useful therefore it would be nice to enhance it a little more.

PR #60, opened 2017-11-21
Added --update-readme option to generateTagsReadme

MichaelSimons:
Related to #58

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
@@ -6,8 +6,10 @@ using Microsoft.DotNet.ImageBuilder.Model;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
 namespace Microsoft.DotNet.ImageBuilder.Commands
@@ -23,37 +25,16 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             Utilities.WriteHeading("GENERATING TAGS README");
             foreach (RepoInfo repo in Manifest.Repos)
             {
-                StringBuilder tagsReadme = new StringBuilder();
+                string tagsDoc = GetTagsDocumentation(repo);
 
-                var platformGroups = repo.Images
-                    .OrderBy(image => image.Model.ReadmeOrder)
-                    .SelectMany(image => image.Platforms.Select(platform => new { Image = image, Platform = platform }))
-                    .GroupBy(tuple => new { tuple.Platform.Model.OS, tuple.Platform.Model.OsVersion, tuple.Platform.Model.Architecture })
-                    .OrderByDescending(platformGroup => platformGroup.Key.Architecture)
-                    .ThenBy(platformGroup => platformGroup.Key.OS)
-                    .ThenByDescending(platformGroup => platformGroup.Key.OsVersion);
+                Console.WriteLine($"-- {repo.Name} Tags Documentation:");
+                Console.WriteLine();
+                Console.WriteLine(tagsDoc);
 
-                foreach (var platformGroup in platformGroups)
+                if (Options.UpdateReadme)
                 {
-                    string os = GetOsDisplayName(platformGroup.Key.OS, platformGroup.Key.OsVersion);
-                    string arch = GetArchitectureDisplayName(platformGroup.Key.Architecture);
-                    tagsReadme.AppendLine($"# Supported {os} {arch} tags");
-                    tagsReadme.AppendLine();
-                    foreach (var tuple in platformGroup)
-                    {
-                        string tags = GetDocumentedTags(tuple.Platform.Tags)
-                            .Concat(GetDocumentedTags(tuple.Image.SharedTags))
-                            .Select(tag => $"`{tag}`")
-                            .Aggregate((working, next) => $"{working}, {next}");
-                        string dockerfile = tuple.Platform.DockerfilePath.Replace('\\', '/');
-                        tagsReadme.AppendLine($"- [{tags} (*{dockerfile}*)]({Options.SourceUrl}/{dockerfile})");
-                    }
-
-                    tagsReadme.AppendLine();
+                    UpdateReadme(tagsDoc, repo);
                 }
-
-                Console.WriteLine($"-- {repo.Name} Tags Readme:");
-                Console.WriteLine(tagsReadme);
             }
 
             return Task.CompletedTask;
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
```diff
@@ -82,6 +63,40 @@ private static IEnumerable<string> GetDocumentedTags(IEnumerable<TagInfo> tagInf
...
+        {
+            StringBuilder tagsDoc = new StringBuilder();
+
+            var platformGroups = repo.Images
+                .OrderBy(image => image.Model.ReadmeOrder)
+                .SelectMany(image => image.Platforms.Select(platform => new { Image = image, Platform = platform }))
+                .GroupBy(tuple => new { tuple.Platform.Model.OS, tuple.Platform.Model.OsVersion, tuple.Platform.Model.Architecture })
+                .OrderByDescending(platformGroup => platformGroup.Key.Architecture)
+                .ThenBy(platformGroup => platformGroup.Key.OS)
+                .ThenByDescending(platformGroup => platformGroup.Key.OsVersion);
```
ravimeda: Observation - Adding a `ToList()` as in `.ThenByDescending(platformGroup => platformGroup.Key.OsVersion).ToList();` would increase the time to prepare `platformGroups`. However, it would decrease the time to finish `foreach (var platformGroup in platformGroups)` below. Net result would be the same i.e., around 15ms to finish `GetTagsDocumentation`.

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
```diff
@@ -82,6 +63,40 @@ private static IEnumerable<string> GetDocumentedTags(IEnumerable<TagInfo> tagInf
...
+
+            foreach (var platformGroup in platformGroups)
+            {
+                string os = GetOsDisplayName(platformGroup.Key.OS, platformGroup.Key.OsVersion);
```
ravimeda: Collapse as - 

```C#
tagsDoc.AppendLine(
                    $"# Supported {GetOsDisplayName(platformGroup.Key.OS, platformGroup.Key.OsVersion)} " +
                    $"{GetArchitectureDisplayName(platformGroup.Key.Architecture)} tags");
```

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
```diff
@@ -104,5 +119,17 @@ private static string GetOsDisplayName(OS os, string osVersion)
 
             return displayName;
         }
+
+        private void UpdateReadme(string tagsDocumentation, RepoInfo repo)
+        {
+            Utilities.WriteHeading("UPDATING README");
+
+            string readme = File.ReadAllText(repo.Model.ReadmePath);
+            string updatedReadme = Regex.Replace(readme, "(# .*\\s*(- \\[.*\\s*)+)+", tagsDocumentation);
```
ravimeda: Nice job on preparing the regex pattern! How about adding RegexOptions - IgnoreCase? If existing bad readme had `AMD64` we can match and replace it with `amd64`.

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
```diff
@@ -104,5 +119,17 @@ private static string GetOsDisplayName(OS os, string osVersion)
 
             return displayName;
         }
+
+        private void UpdateReadme(string tagsDocumentation, RepoInfo repo)
```
ravimeda: In `repo` the only property used is `repo.Model.ReadmePath`, would it be cleaner to update the signature to accept a string as opposed to `RepoInfo`.

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
```diff
@@ -23,37 +25,16 @@ public override Task ExecuteAsync()
...
 
-                var platformGroups = repo.Images
-                    .OrderBy(image => image.Model.ReadmeOrder)
-                    .SelectMany(image => image.Platforms.Select(platform => new { Image = image, Platform = platform }))
-                    .GroupBy(tuple => new { tuple.Platform.Model.OS, tuple.Platform.Model.OsVersion, tuple.Platform.Model.Architecture })
-                    .OrderByDescending(platformGroup => platformGroup.Key.Architecture)
-                    .ThenBy(platformGroup => platformGroup.Key.OS)
-                    .ThenByDescending(platformGroup => platformGroup.Key.OsVersion);
+                Console.WriteLine($"-- {repo.Name} Tags Documentation:");
+                Console.WriteLine();
```
ravimeda: Minor - `Console.WriteLine();` why not remove this by adding a `\n` to the previous statement as in -
 `Console.WriteLine($"-- {repo.Name} Tags Documentation:\n");`

MichaelSimons: Xplat applications shouldn't use `\n`.  `Environment.Newline` should be used instead.  I don't see much difference in `Console.Writeline()` vs appending `Environment.Newline` in this scenario.  Leaving as is.

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
MichaelSimons: `ToList` is valuable when you are going to access the Linq query multiple times.  In this case it is only being accessed once therefore it doesn't add value.  Leaving as is.

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
MichaelSimons: I find the current easier to visualize what the resulting string will look like, in other words it is more readable. 
 Leaving as is.

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
MichaelSimons: `repo.Model.ReadmePath` has to be accessed someplace.  Either here or by the caller.  Making use of the model makes sense in a lot of cases because it is a nice abstraction and multiple members are typically accessed.  In this case it isn't strictly necessary.  Leaving as is.

src/Microsoft.DotNet.ImageBuilder/Commands/GenerateTagsReadmeCommand.cs
MichaelSimons: I don't think IgnoreCase is necessary.  The regex itself doesn't match on any letters - just symbols.  Leaving as is.

Issue #63, opened 2017-12-13
Add support in ImageBuilder to substitute global variables in manifest


ravimeda:
Consider a scenario where a user would like to improve the traceability of images.  User would like the 
images built using _ImageBuilder_ to be tagged with a specific time-stamp.  User can update the manifest.json to include a time-stamp variable in image tag.  An example of such tag is `alpine-3.6--$(timeStamp)`.  _ImageBuilder_ should be able to replace all such `$(timeStamp)` variables with a time-stamp. This means the example tag should be like `alpine-3.6-20173413043446`, using a `yyyymmddhhmmss` DateTime format. All images built during the _ImageBuilder_ session would have the same "_global_" time-stamp.

Another scenario to further improve the traceability of images is to include the commit SHA of the Dockerfile that was used to build the image.  For this purpose, say, `(dockerfileSha)` variable is included in image tag. An example of such tag is `alpine-3.6-$(dockerfileSha)-$(timeStamp)`.  _ImageBuilder_  should replace both the variables so that the example tag would be like `alpine-3.6-3148f11-20173413043446`, where `3148f11` is the commit SHA of the Dockerfile.

PR #64, opened 2017-12-13
Support for global variable substitution

ravimeda:
Adds the ability to _ImageBuilder_ to substitute global variables with corresponding values in manifest.json. Dockerfile commit SHA and timestamp variable substitution is enabled.

Refer to https://github.com/dotnet/docker-tools/issues/63

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,6 @@
 # Visual Studio Code cache/options directory
 .vscode/
 .vs
+
+# Visual Studio debug profile
 /Microsoft.DotNet.ImageBuilder/src/Properties/launchSettings.json
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 # Visual Studio Code cache/options directory
 .vscode/
 .vs
+/Microsoft.DotNet.ImageBuilder/src/Properties/launchSettings.json
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@
 .vs
 
 # Visual Studio debug profile
-/Microsoft.DotNet.ImageBuilder/src/Properties/launchSettings.json
+**/launchSettings.json
--- a/Microsoft.DotNet.ImageBuilder/src/GitHelper.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/GitHelper.cs
@@ -13,7 +13,7 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
             ProcessStartInfo startInfo = new ProcessStartInfo("git", $"log -1 --format=format:%h {filePath}");
             startInfo.RedirectStandardOutput = true;
             Process gitLogProcess = ExecuteHelper.Execute(
-                startInfo, false, $"Unable to retrieve the commit for {filePath}");
+                startInfo, false, $"Unable to retrieve the latest commit SHA for {filePath}");
             return gitLogProcess.StandardOutput.ReadToEnd().Trim();
         }
     }
--- a/Microsoft.DotNet.ImageBuilder/src/GitHelper.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/GitHelper.cs
@@ -1,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+
+namespace Microsoft.DotNet.ImageBuilder.ViewModel
+{
+    public static class GitHelper
+    {
+        public static string GetCommitSha(string filePath)
+        {
+            ProcessStartInfo startInfo = new ProcessStartInfo("git", $"log -1 --format=format:%h {filePath}");
+            startInfo.RedirectStandardOutput = true;
+            Process gitLogProcess = ExecuteHelper.Execute(
+                startInfo, false, $"Unable to retrieve the commit for {filePath}");
+            return gitLogProcess.StandardOutput.ReadToEnd().Trim();
+        }
+    }
+}
--- a/Microsoft.DotNet.ImageBuilder/src/Utilities.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Utilities.cs
@@ -4,8 +4,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.IO;
 using System.Text.RegularExpressions;
 
 namespace Microsoft.DotNet.ImageBuilder
@@ -13,45 +11,37 @@ namespace Microsoft.DotNet.ImageBuilder
     public static class Utilities
     {
         private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
-        private const string TagVariablePattern = "\\$\\((?<variable>[\\w]+)\\)";
+        private const string VariableGroupName = "variable";
+        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w]+)\\)";
 
         public static string SubstituteVariables(
-            IDictionary<string, string> variables,
+            IDictionary<string, string> userVariables,
             string expression,
-            Func<string, string> getVariableSubstitute = null)
+            Func<string, string> getSystemValue = null)
         {
             foreach (Match match in Regex.Matches(expression, TagVariablePattern))
             {
-                string variableName = match.Groups["variable"].Value;
+                string variableName = match.Groups[VariableGroupName].Value;
                 string variableValue = null;
-                if (variables == null || !variables.TryGetValue(variableName, out variableValue))
+                if (userVariables == null || !userVariables.TryGetValue(variableName, out variableValue))
                 {
-                    if (getVariableSubstitute != null)
+                    if (getSystemValue != null)
                     {
-                        variableValue = getVariableSubstitute(variableName);
+                        variableValue = getSystemValue(variableName);
                     }
-                    if (variableValue == null && match.Value == "$(TimeStamp)")
+                    if (variableValue == null && variableName == "TimeStamp")
                     {
                         variableValue = TimeStamp;
                     }
                     if (variableValue == null)
                     {
-                        throw new InvalidDataException($"A value was not found for the variable '{match.Value}'");
+                        throw new InvalidOperationException($"A value was not found for the variable '{match.Value}'");
                     }
                 }
                 expression = expression.Replace(match.Value, variableValue);
             }
 
             return expression;
-        }
-
-        public static string GetAbbreviatedCommitSha(string filePath)
-        {
-            ProcessStartInfo startInfo = new ProcessStartInfo("git", $"log -1 --format=format:%h {filePath}");
-            startInfo.RedirectStandardOutput = true;
-            Process gitLogProcess = ExecuteHelper.Execute(
-                startInfo, false, $"Unable to retrieve the commit for {filePath}");
-            return gitLogProcess.StandardOutput.ReadToEnd().Trim();
         }
 
         public static void WriteHeading(string heading)
--- a/Microsoft.DotNet.ImageBuilder/src/Utilities.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Utilities.cs
@@ -4,22 +4,67 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.DotNet.ImageBuilder
 {
     public static class Utilities
     {
-        public static string SubstituteVariables(IDictionary<string, string> variables, string expression)
+        private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
+        private const string TagVariablePattern = "\\$\\((?<variable>[\\w]+)\\)";
+
+        public static string SubstituteVariables(
+            IDictionary<string, string> variables,
+            string expression,
+            Func<string, string> getVariableValue = null)
         {
-            if (variables != null)
+            foreach (Match match in Regex.Matches(expression, TagVariablePattern))
             {
-                foreach (KeyValuePair<string, string> kvp in variables)
+                string variableName = match.Groups[1].Value;
+                string variableValue = null;
+                if (variables == null || !variables.TryGetValue(variableName, out variableValue))
                 {
-                    expression = expression.Replace($"$({kvp.Key})", kvp.Value);
+                    if (getVariableValue != null)
+                    {
+                        variableValue = getVariableValue(variableName);
+                    }
+                    if (variableValue == null && match.Value == "$(timeStamp)")
+                    {
+                        variableValue = TimeStamp;
+                    }
                 }
+                if (variableValue == null)
+                {
+                    throw new InvalidDataException(
+                            $"Unable to determine the value for the given variable. Variable: {match.Value}");
+                }
+                expression = expression.Replace(match.Value, variableValue);
             }
 
             return expression;
+        }
+
+        public static Func<string, string> GetSha(string filePath, bool isDryRun = false)
+        {
+            return delegate (string variableName)
+            {
+                if (!File.Exists(filePath))
+                {
+                    throw new FileNotFoundException($"Unable to find the file.", filePath);
+                }
+                if (!string.Equals("dockerfileSha", variableName))
+                {
+                    return null;
+                }
+
+                ProcessStartInfo startInfo = new ProcessStartInfo("git", $"log -1 --format=format:%h {filePath}");
+                startInfo.RedirectStandardOutput = true;
+                Process gitLogProcess = ExecuteHelper.Execute(
+                    startInfo, isDryRun, $"Unable to retrieve the commit timestamp for {filePath}");
+                return isDryRun ? "" : gitLogProcess.StandardOutput.ReadToEnd().Trim();
+            };
         }
 
         public static void WriteHeading(string heading)
--- a/Microsoft.DotNet.ImageBuilder/src/Utilities.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Utilities.cs
@@ -10,14 +10,14 @@ namespace Microsoft.DotNet.ImageBuilder
 {
     public static class Utilities
     {
+        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w]+)\\)";
         private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
         private const string VariableGroupName = "variable";
-        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w]+)\\)";
 
         public static string SubstituteVariables(
             IDictionary<string, string> userVariables,
             string expression,
-            Func<string, string> getSystemValue = null)
+            Func<string, string> getContextBasedSystemValue = null)
         {
             foreach (Match match in Regex.Matches(expression, TagVariablePattern))
             {
@@ -25,9 +25,9 @@ namespace Microsoft.DotNet.ImageBuilder
                 string variableValue = null;
                 if (userVariables == null || !userVariables.TryGetValue(variableName, out variableValue))
                 {
-                    if (getSystemValue != null)
+                    if (getContextBasedSystemValue != null)
                     {
-                        variableValue = getSystemValue(variableName);
+                        variableValue = getContextBasedSystemValue(variableName);
                     }
                     if (variableValue == null && variableName == "TimeStamp")
                     {
--- a/Microsoft.DotNet.ImageBuilder/src/Utilities.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Utilities.cs
@@ -18,27 +18,26 @@ namespace Microsoft.DotNet.ImageBuilder
         public static string SubstituteVariables(
             IDictionary<string, string> variables,
             string expression,
-            Func<string, string> getVariableValue = null)
+            Func<string, string> getVariableSubstitute = null)
         {
             foreach (Match match in Regex.Matches(expression, TagVariablePattern))
             {
-                string variableName = match.Groups[1].Value;
+                string variableName = match.Groups["variable"].Value;
                 string variableValue = null;
                 if (variables == null || !variables.TryGetValue(variableName, out variableValue))
                 {
-                    if (getVariableValue != null)
+                    if (getVariableSubstitute != null)
                     {
-                        variableValue = getVariableValue(variableName);
+                        variableValue = getVariableSubstitute(variableName);
                     }
-                    if (variableValue == null && match.Value == "$(timeStamp)")
+                    if (variableValue == null && match.Value == "$(TimeStamp)")
                     {
                         variableValue = TimeStamp;
                     }
-                }
-                if (variableValue == null)
-                {
-                    throw new InvalidDataException(
-                            $"Unable to determine the value for the given variable. Variable: {match.Value}");
+                    if (variableValue == null)
+                    {
+                        throw new InvalidDataException($"A value was not found for the variable '{match.Value}'");
+                    }
                 }
                 expression = expression.Replace(match.Value, variableValue);
             }
@@ -46,25 +45,13 @@ namespace Microsoft.DotNet.ImageBuilder
             return expression;
         }
 
-        public static Func<string, string> GetSha(string filePath, bool isDryRun = false)
+        public static string GetAbbreviatedCommitSha(string filePath)
         {
-            return delegate (string variableName)
-            {
-                if (!File.Exists(filePath))
-                {
-                    throw new FileNotFoundException($"Unable to find the file.", filePath);
-                }
-                if (!string.Equals("dockerfileSha", variableName))
-                {
-                    return null;
-                }
-
-                ProcessStartInfo startInfo = new ProcessStartInfo("git", $"log -1 --format=format:%h {filePath}");
-                startInfo.RedirectStandardOutput = true;
-                Process gitLogProcess = ExecuteHelper.Execute(
-                    startInfo, isDryRun, $"Unable to retrieve the commit timestamp for {filePath}");
-                return isDryRun ? "" : gitLogProcess.StandardOutput.ReadToEnd().Trim();
-            };
+            ProcessStartInfo startInfo = new ProcessStartInfo("git", $"log -1 --format=format:%h {filePath}");
+            startInfo.RedirectStandardOutput = true;
+            Process gitLogProcess = ExecuteHelper.Execute(
+                startInfo, false, $"Unable to retrieve the commit for {filePath}");
+            return gitLogProcess.StandardOutput.ReadToEnd().Trim();
         }
 
         public static void WriteHeading(string heading)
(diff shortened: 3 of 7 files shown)


Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,69 @@
...
+using System.IO;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.DotNet.ImageBuilder
 {
     public static class Utilities
     {
-        public static string SubstituteVariables(IDictionary<string, string> variables, string expression)
+        private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
+        private const string TagVariablePattern = "\\$\\((?<variable>[\\w]+)\\)";
```
MichaelSimons: Why did you define a capture group in the regex but you don't use it when substituting the variable?

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,69 @@
...
+
+        public static string SubstituteVariables(
+            IDictionary<string, string> variables,
+            string expression,
+            Func<string, string> getVariableValue = null)
```
MichaelSimons: From a readability standpoint, it is difficult to distinguish the difference between `variables` and `getVariableValue`.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,69 @@
...
+                    if (getVariableValue != null)
+                    {
+                        variableValue = getVariableValue(variableName);
+                    }
+                    if (variableValue == null && match.Value == "$(timeStamp)")
+                    {
+                        variableValue = TimeStamp;
+                    }
+                }
+                if (variableValue == null)
```
MichaelSimons: I think this should be moved into the previous else.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,69 @@
...
                 {
-                    expression = expression.Replace($"$({kvp.Key})", kvp.Value);
+                    throw new InvalidDataException(
+                            $"Unable to determine the value for the given variable. Variable: {match.Value}");
```
MichaelSimons: The usage of "given" is a little strange.  I think something like "A value was not found for the variable '{match.Value}'."

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,69 @@
...
                 }
+                expression = expression.Replace(match.Value, variableValue);
             }
 
             return expression;
         }
 
+        public static Func<string, string> GetSha(string filePath, bool isDryRun = false)
```
MichaelSimons: I don't understand why this is returning a `Func`.  Please explain.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,69 @@
...
+        {
+            return delegate (string variableName)
+            {
+                if (!File.Exists(filePath))
+                {
+                    throw new FileNotFoundException($"Unable to find the file.", filePath);
+                }
+                if (!string.Equals("dockerfileSha", variableName))
```
MichaelSimons: This function should be just about retrieving a commit sha.  It shouldn't be coupled to variables IMO.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
MichaelSimons: There are many types of Sha's.  The method name should indicate what sha is retrieved.

MichaelSimons: I don't see where isDryRun is ever specified.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
```diff
@@ -11,16 +12,17 @@ public class TagInfo
...
 
-        public static TagInfo Create(string name, Tag model, Manifest manifest, string repoName)
+        public static TagInfo Create(string name, Tag model, Manifest manifest, string repoName, string filePath = "")
         {
             TagInfo tagInfo = new TagInfo();
             tagInfo.Model = model;
-            tagInfo.Name = manifest.SubstituteTagVariables(name);
+            Func<string, string> GetSha = Utilities.GetSha(filePath);
+            tagInfo.Name = Utilities.SubstituteVariables(
+                variables: manifest.TagVariables, expression: name, getVariableValue: GetSha);
```
MichaelSimons: Curious why you are using named variables here?

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,69 @@
...
             }
 
             return expression;
         }
 
+        public static Func<string, string> GetSha(string filePath, bool isDryRun = false)
+        {
+            return delegate (string variableName)
+            {
+                if (!File.Exists(filePath))
```
MichaelSimons: I'm not sure this check is working correctly when there are shared tags in the manifest.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,69 @@
...
-                foreach (KeyValuePair<string, string> kvp in variables)
+                string variableName = match.Groups[1].Value;
+                string variableValue = null;
+                if (variables == null || !variables.TryGetValue(variableName, out variableValue))
+                {
+                    if (getVariableValue != null)
+                    {
+                        variableValue = getVariableValue(variableName);
+                    }
+                    if (variableValue == null && match.Value == "$(timeStamp)")
```
MichaelSimons: I think the built in variables should be CamelCased.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
```diff
@@ -11,16 +12,17 @@ public class TagInfo
         public string FullyQualifiedName { get; private set; }
         public Tag Model { get; private set; }
         public string Name { get; private set; }
-
         private TagInfo()
         {
         }
 
-        public static TagInfo Create(string name, Tag model, Manifest manifest, string repoName)
+        public static TagInfo Create(string name, Tag model, Manifest manifest, string repoName, string filePath = "")
```
MichaelSimons: `filePath` is not very descriptive.  What type of file are you intending this to be?

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
MichaelSimons: I'm not sure `dockerfileSha` is a good name.  This might get confused with the Docker image sha.  Althought it gets long I think `DockerfileCommitSha` or `DockerfileGitCommitSha` would be better.

.gitignore
```diff
@@ -6,3 +6,4 @@
 # Visual Studio Code cache/options directory
 .vscode/
 .vs
+/Microsoft.DotNet.ImageBuilder/src/Properties/launchSettings.json
```
MichaelSimons: I think it would be better to define this so that any launchSetting.json files are excluded

ravimeda: I'll add `# Visual Studio debug profile`

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Good catch. Updated usage as `match.Groups["variable"].Value`

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: I'll update to CamelCase.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: I'll update.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: As discussed offline this is partly due misunderstood requirement. I'll update the method as - `public static string GetAbbreviatedCommitSha(string filePath)`

ravimeda: Renamed as `GetAbbreviatedCommitSha`.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Yes, will remove the check.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Yes, not required. Misunderstood requirement.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Yes, will remove.

ravimeda: Let's use `DockerfileGitCommitSha`

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
ravimeda: Let's use `dockerfilePath`

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
ravimeda: I was following - use named variables when there are 3 or more arguments are being passed.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,56 @@
...
     public static class Utilities
     {
-        public static string SubstituteVariables(IDictionary<string, string> variables, string expression)
+        private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
+        private const string TagVariablePattern = "\\$\\((?<variable>[\\w]+)\\)";
+
+        public static string SubstituteVariables(
+            IDictionary<string, string> variables,
+            string expression,
+            Func<string, string> getVariableSubstitute = null)
```
MichaelSimons: I don't think `getVariableSubstitute` solved the naming issue I previously mentioned.  Maybe rename `variables` to `userValues` and `getVariableSubstitute` to `getSystemValue`

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,56 @@
...
+                string variableName = match.Groups["variable"].Value;
+                string variableValue = null;
+                if (variables == null || !variables.TryGetValue(variableName, out variableValue))
                 {
-                    expression = expression.Replace($"$({kvp.Key})", kvp.Value);
+                    if (getVariableSubstitute != null)
+                    {
+                        variableValue = getVariableSubstitute(variableName);
+                    }
+                    if (variableValue == null && match.Value == "$(TimeStamp)")
```
MichaelSimons: I was expecting `$(TimeStamp)` would be just `TimeStamp`.  We should not be spreading the variable substitution format throughout the code.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,56 @@
...
+                    {
+                        variableValue = TimeStamp;
+                    }
+                    if (variableValue == null)
+                    {
+                        throw new InvalidDataException($"A value was not found for the variable '{match.Value}'");
```
MichaelSimons: I'm not sure InvalidDataException is the correct Exception.  I think InvalidOperationException would be more appropriate.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,24 +4,56 @@
...
+                    }
                 }
+                expression = expression.Replace(match.Value, variableValue);
             }
 
             return expression;
         }
 
+        public static string GetAbbreviatedCommitSha(string filePath)
```
MichaelSimons: Looking forward, I think it would be a good idea to create a GitHelper class for this and rename it to `GitCommitSha`.  I personally don't think abbreviated in the name.  I could foresee an option being added  in the future that would allow the user to specify abbreviated or not.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
```diff
@@ -8,22 +8,33 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
 {
     public class TagInfo
     {
+        public string DockerfilePath { get; set; }
         public string FullyQualifiedName { get; private set; }
         public Tag Model { get; private set; }
         public string Name { get; private set; }
-
```
MichaelSimons: Please leave empty space between props and ctor

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
```diff
@@ -8,22 +8,33 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
 {
     public class TagInfo
     {
+        public string DockerfilePath { get; set; }
```
MichaelSimons: Seems like this should private.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
```diff
@@ -8,22 +8,33 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
...
             tagInfo.Model = model;
-            tagInfo.Name = manifest.SubstituteTagVariables(name);
+            tagInfo.DockerfilePath = dockerfilePath;
+            tagInfo.Name = Utilities.SubstituteVariables(manifest.TagVariables, name, tagInfo.GetDockerfileGitCommitSha);
             tagInfo.FullyQualifiedName = $"{repoName}:{tagInfo.Name}";
 
             return tagInfo;
         }
+
+        public string GetDockerfileGitCommitSha(string variableName)
```
MichaelSimons: I don't think this should be named specific to the one variable it supports today.  I think it would be better to keep it abstract so that in the future more variables can be added without needing to be refactored.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
```diff
@@ -8,22 +8,33 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
...
+        {
+            string commitSha = null;
+            if (variableName == "DockerfileGitCommitSha")
```
MichaelSimons: After thinking about this some more, I wonder if we shouldn't be using the Docker build context commit sha.  The point is that if anything is changed within the build context, we would want to identify this with the sha.

MichaelSimons: Shouldn't this check also see if a DockerfilePath exists?  We don't want to allow this variable to be used in shared tags.

ravimeda: Yes, consider a scenario where a commit modifies a script that is passed to `--build-arg`. Though the Dockerfile did not change, a new image should be built using the modified script.

To achieve this functionality, `BuildContextPath` will be requires, as opposed to `DockerfilePath` in `TagInfo`.

Addressing the comments about the method name, and checking if the file exists, this method would be updated as follows -

```C#
        public string GetSubstituteValue(string variableName)
        {
            string commitSha = null;
            if (variableName == "DockerfileGitCommitSha")
            {
                if (!File.Exists(Platform.DockerfilePath))
                {
                    throw new FileNotFoundException("Unable to locate the file.", Platform.DockerfilePath);
                }
                commitSha = GitHelper.GetCommitSha(Platform.BuildContextPath);
            }
            return commitSha;
        }
```

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
ravimeda: Will update.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
ravimeda: Yes. But this will be removed and replaced with `private PlatformInfo Platform { get; set; }`. 
See the comment below about other changes in `TagInfo.cs`

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
ravimeda: Please see the below comment about updates to `TagInfo.cs`

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Ok. Will update.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Is it like - `if (variableValue == null && variableName == "TimeStamp")` ?

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Good point. Will move this method to `GitHelper` class.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Alright. Will update.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: You mean, like this - 
```C#
        private const string VariableGroupName = "variable";
        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w]+)\\)";
```

MichaelSimons: yes...and then the const should be used when reading the matches.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
MichaelSimons: yes

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
MichaelSimons: I personally don't like passing PlatformInfo into TagInfo.  It creates a circular reference and limits the flexibility in future usage of the TagInfo.

ravimeda: Agree. Let's pass the required values i.e., `platformInfo.BuildContextPath` and `platformInfo.DockerfilePath`

.gitignore
```diff
@@ -6,3 +6,6 @@
 # Visual Studio Code cache/options directory
 .vscode/
 .vs
+
+# Visual Studio debug profile
+**/launchSettings.json
```
ravimeda: FYI `**` ignores all files named `launchSettings.json` in the repo. I'm not sure if a file with the same name will be needed for any other usecase.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,19 +4,41 @@
...
     {
-        public static string SubstituteVariables(IDictionary<string, string> variables, string expression)
+        private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
+        private const string VariableGroupName = "variable";
+        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w]+)\\)";
+
+        public static string SubstituteVariables(
+            IDictionary<string, string> userVariables,
+            string expression,
+            Func<string, string> getSystemValue = null)
```
ravimeda: `getSystemValue` - name does not seem to fit well. Though `TimeStamp` is a `System` property, `DockerGitCommitSha` is a repository characteristic.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
```diff
@@ -16,14 +19,36 @@ private TagInfo()
...
 
             return tagInfo;
         }
+
+        public string GetSubstituteValue(string variableName)
+        {
+            string commitSha = null;
+            if (variableName == "DockerfileGitCommitSha")
+            {
+                if (!File.Exists(this.DockerfilePath))
```
ravimeda: As discussed, this check is for shared tags, which do not have a corresponding Dockerfile but can have global variables.

Microsoft.DotNet.ImageBuilder/src/GitHelper.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Microsoft.DotNet.ImageBuilder.ViewModel
+{
+    public static class GitHelper
+    {
+        public static string GetCommitSha(string filePath)
+        {
+            ProcessStartInfo startInfo = new ProcessStartInfo("git", $"log -1 --format=format:%h {filePath}");
+            startInfo.RedirectStandardOutput = true;
+            Process gitLogProcess = ExecuteHelper.Execute(
+                startInfo, false, $"Unable to retrieve the commit for {filePath}");
```
MichaelSimons: Consider rewording to include "latest commit sha".  I think it would help make the message more meaningful.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
MichaelSimons: What is your suggestion?  'getContextBasedSystemValue' - kind of long but is descriptive.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
MichaelSimons: I don't think this check should be here at all and the TagInfo shouldn't need the Dockerfile Path.  IMO the code should be something like

`if (variableName == "DockerfileGitCommitSha" && this.BuildContextPath != null) `

This will only allow the variable to be used when there is a BuildContextPath - e.g. - it is a concrete tag.  If it is a shared tag and the variable is used, the user will get the standard variable value could be be retrieved exception.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
```diff
@@ -16,14 +19,36 @@ private TagInfo()
...
+            tagInfo.DockerfilePath = dockerfilePath;
+            tagInfo.Name = Utilities.SubstituteVariables(manifest.TagVariables, name, tagInfo.GetSubstituteValue);
             tagInfo.FullyQualifiedName = $"{repoName}:{tagInfo.Name}";
 
             return tagInfo;
         }
+
+        public string GetSubstituteValue(string variableName)
+        {
+            string commitSha = null;
```
MichaelSimons: This variable name assumes the `DockerfileGitCommitSha` variable.  The method is intended to be a place additional variables could be added.  I think `value` or `variableValue` would be a better name

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -4,19 +4,41 @@
...
 using System;
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.DotNet.ImageBuilder
 {
     public static class Utilities
     {
-        public static string SubstituteVariables(IDictionary<string, string> variables, string expression)
+        private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
```
MichaelSimons: Consider alphabetizing these fields/properties.

ravimeda: Ok. Will update.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Ok to settle with `getContextBasedSystemValue`.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
ravimeda: Good point. Will update as `variableValue`.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
ravimeda: Updated as - 
```C#
            string variableValue = null;
            if (variableName == "DockerfileGitCommitSha" && this.BuildContextPath != null)
            {
                variableValue = GitHelper.GetCommitSha(this.BuildContextPath);
            }
            return variableValue;
```

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Thinking about this more, order in a Class is -

- Constants
- Constructors
- Fields or Properties
- Methods
- <...>

Sort alphabetically in each group. If we go with that, then the in the current file the ordering will be -

```C#
        private const string VariableGroupName = "variable";
        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w]+)\\)";
        private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
```

If not, an alphabetical sort ignoring the const would be like -

```C#
        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w]+)\\)";
        private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
        private const string VariableGroupName = "variable";
```

My preference is alphabetical with no grouping.

MichaelSimons: I like that.

Microsoft.DotNet.ImageBuilder/src/ViewModel/TagInfo.cs
```diff
@@ -16,14 +18,32 @@ private TagInfo()
...
         }
 
-        public static TagInfo Create(string name, Tag model, Manifest manifest, string repoName)
+        public static TagInfo Create(
+            string name,
+            Tag model,
+            Manifest manifest,
+            string repoName,
+            string buildContextPath = null,
+            string dockerfilePath = null)
```
MichaelSimons: dockerfilePath is not used and should be removed.

ravimeda: Good one!  Will update.

Issue #65, opened 2017-12-21
Add support for build hook


ravimeda:
A hook is a script that can be run either before or after a Docker command.  Hook name is convention based, which means `pre-build` is the name of the script that runs before Docker build command, while `post-build` is the one that runs after build command.  Thus, this enables running a custom script before and/or after the build command.

PR #66, opened 2017-12-21
Enable build hooks in ImageBuilder

ravimeda:
Currently, _ImageBuilder_ does not offer any hooks.  A hook enables running a custom script before or after a Docker command. Refer to https://github.com/dotnet/docker-tools/issues/65

These changes add support for `pre-build` and `post-build`.  Contract for `InvokeBuildHook`, is -

1. Check if the named script is available in a folder named `hooks` in the build content. If yes, then executes the script as a Shell script.
2. If the script was not found in step-1, then look for a PowerShell script. This means named script with an extension `.ps1`.  If found , then execute that PowerShell script.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -92,13 +92,14 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 scriptPath = Path.ChangeExtension(scriptPath, ".ps1");
                 if (!File.Exists(scriptPath))
                 {
-                    Console.WriteLine($"Failed to locate build hook {scriptPath}'");
                     return;
                 }
+
                 startInfo = new ProcessStartInfo(
                     RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "PowerShell" : "pwsh", 
                     $"-NoProfile -File \"{scriptPath}\"");
             }
+
             ExecuteHelper.Execute(startInfo, Options.IsDryRun, $"Failed to execute build hook '{scriptPath}'");
         }
 
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -6,6 +6,7 @@ using Microsoft.DotNet.ImageBuilder.ViewModel;
 using System;
 using System.Collections.Generic;
 using System.CommandLine;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
@@ -52,10 +53,12 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                             .Select(tag => tag.FullyQualifiedName)
                             .Concat(platformTags);
                         string tagArgs = $"-t {string.Join(" -t ", allTags)}";
+                        InvokeBuildHook("pre-build", platform.BuildContextPath, Options.IsDryRun);
                         ExecuteHelper.Execute(
                             "docker",
                             $"build {tagArgs} -f {dockerfilePath} {platform.BuildContextPath}",
                             Options.IsDryRun);
+                        InvokeBuildHook("post-build", platform.BuildContextPath, Options.IsDryRun);
                         BuiltTags = BuiltTags.Concat(platformTags);
                     }
                     finally
@@ -66,6 +69,34 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                         }
                     }
                 }
+            }
+        }
+
+        private static void InvokeBuildHook(string hookName, string buildContextPath, bool isDryRun = false)
+        {
+            string hookFolder = Path.Combine(buildContextPath, "hooks");
+            if (!Directory.Exists(hookFolder))
+            {
+                return;
+            }
+
+            string hookPath = Path.Combine(hookFolder, hookName);
+            string PSScriptExtension = ".ps1";
+            hookPath = File.Exists(hookPath) ? hookPath
+                : File.Exists($"{hookPath}{PSScriptExtension}") ? $"{hookPath}{PSScriptExtension}" : "";
+            if (hookPath != "")
+            {
+                ProcessStartInfo startInfo = new ProcessStartInfo(hookPath);
+                startInfo.WorkingDirectory = buildContextPath;
+                if (Path.GetExtension(hookPath) == PSScriptExtension)
+                {
+                    startInfo.FileName = "PowerShell";
+                    startInfo.Arguments = $"-NoProfile -File hooks\\{hookName}{PSScriptExtension}";
+                }
+                ExecuteHelper.Execute(
+                    startInfo,
+                    isDryRun,
+                    $"Failure in {hookPath} hook");
             }
         }
 
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -9,6 +9,7 @@ using System.CommandLine;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
@@ -74,30 +75,31 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
         private void InvokeBuildHook(string hookName, string buildContextPath)
         {
-            string buildHookFolder = Path.Combine(buildContextPath, "hooks");
+            string buildHookFolder = Path.GetFullPath(Path.Combine(buildContextPath, "hooks"));
             if (!Directory.Exists(buildHookFolder))
             {
                 return;
             }
 
             string scriptPath = Path.Combine(buildHookFolder, hookName);
-            string psScript = $"{hookName}.ps1";
-            ProcessStartInfo startInfo = new ProcessStartInfo();
+            ProcessStartInfo startInfo;
             if (File.Exists(scriptPath))
             {
-                startInfo.FileName = scriptPath;
-            }
-            else if (File.Exists(Path.Combine(buildHookFolder, psScript)))
-            {
-                startInfo.FileName = "PowerShell";
-                startInfo.Arguments = $"-NoProfile -File {psScript}";
+                startInfo = new ProcessStartInfo(scriptPath);
             }
             else
             {
-                return;
+                scriptPath = Path.ChangeExtension(scriptPath, ".ps1");
+                if (!File.Exists(scriptPath))
+                {
+                    Console.WriteLine($"Failed to locate build hook {scriptPath}'");
+                    return;
+                }
+                startInfo = new ProcessStartInfo(
+                    RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "PowerShell" : "pwsh", 
+                    $"-NoProfile -File \"{scriptPath}\"");
             }
-            startInfo.WorkingDirectory = buildHookFolder;
-            ExecuteHelper.Execute(startInfo, Options.IsDryRun, $"Failed to execute '{hookName}' hook in '{buildHookFolder}'");
+            ExecuteHelper.Execute(startInfo, Options.IsDryRun, $"Failed to execute build hook '{scriptPath}'");
         }
 
         private void PullBaseImages()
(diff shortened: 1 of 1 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
 
+        private static void InvokeBuildHook(string hookName, string buildContextPath, bool isDryRun = false)
+        {
+            string hookFolder = Path.Combine(buildContextPath, "hooks");
+            if (!Directory.Exists(hookFolder))
+            {
+                return;
+            }
+
+            string hookPath = Path.Combine(hookFolder, hookName);
```
MichaelSimons: The variables `hookFolder` and `hookPath` are difficult to differentiate.  The naming should identify the purpose more clearly.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+            if (hookPath != "")
+            {
+                ProcessStartInfo startInfo = new ProcessStartInfo(hookPath);
+                startInfo.WorkingDirectory = buildContextPath;
+                if (Path.GetExtension(hookPath) == PSScriptExtension)
+                {
+                    startInfo.FileName = "PowerShell";
+                    startInfo.Arguments = $"-NoProfile -File hooks\\{hookName}{PSScriptExtension}";
+                }
+                ExecuteHelper.Execute(
```
MichaelSimons: Seems like the args could fit on one line?

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+                    startInfo,
+                    isDryRun,
+                    $"Failure in {hookPath} hook");
```
MichaelSimons: I think this error message could enhanced.  It is always a good idea to put some type of quote around user configurable data - '{hookPath}'.  `Failed to execute build hook '{hookpath}'.`

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+            hookPath = File.Exists(hookPath) ? hookPath
+                : File.Exists($"{hookPath}{PSScriptExtension}") ? $"{hookPath}{PSScriptExtension}" : "";
+            if (hookPath != "")
+            {
+                ProcessStartInfo startInfo = new ProcessStartInfo(hookPath);
+                startInfo.WorkingDirectory = buildContextPath;
+                if (Path.GetExtension(hookPath) == PSScriptExtension)
+                {
+                    startInfo.FileName = "PowerShell";
+                    startInfo.Arguments = $"-NoProfile -File hooks\\{hookName}{PSScriptExtension}";
```
MichaelSimons: Why construct the build hook path again?

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+            string PSScriptExtension = ".ps1";
+            hookPath = File.Exists(hookPath) ? hookPath
+                : File.Exists($"{hookPath}{PSScriptExtension}") ? $"{hookPath}{PSScriptExtension}" : "";
+            if (hookPath != "")
+            {
+                ProcessStartInfo startInfo = new ProcessStartInfo(hookPath);
+                startInfo.WorkingDirectory = buildContextPath;
+                if (Path.GetExtension(hookPath) == PSScriptExtension)
+                {
+                    startInfo.FileName = "PowerShell";
```
MichaelSimons: Have you verified that this works with PowerShell on linux?

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
             }
         }
 
+        private static void InvokeBuildHook(string hookName, string buildContextPath, bool isDryRun = false)
```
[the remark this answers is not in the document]
ravimeda: Ok. Will update.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
ravimeda: I'll rename/refactor.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+            string hookFolder = Path.Combine(buildContextPath, "hooks");
+            if (!Directory.Exists(hookFolder))
+            {
+                return;
+            }
+
+            string hookPath = Path.Combine(hookFolder, hookName);
+            string PSScriptExtension = ".ps1";
+            hookPath = File.Exists(hookPath) ? hookPath
+                : File.Exists($"{hookPath}{PSScriptExtension}") ? $"{hookPath}{PSScriptExtension}" : "";
```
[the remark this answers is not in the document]
ravimeda: Ok. I'll refactor.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+            if (hookPath != "")
+            {
+                ProcessStartInfo startInfo = new ProcessStartInfo(hookPath);
+                startInfo.WorkingDirectory = buildContextPath;
+                if (Path.GetExtension(hookPath) == PSScriptExtension)
```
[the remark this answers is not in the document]
ravimeda: Ok. I'll refactor.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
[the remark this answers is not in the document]
ravimeda: Ok. I'll refactor.

ravimeda: So far verified on macOS. I will do a sanity check on Ubuntu.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
ravimeda: Yes. WIll update.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
ravimeda: Good one. I'll update.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+            string scriptPath = Path.Combine(buildHookFolder, hookName);
+            string psScript = $"{hookName}.ps1";
+            ProcessStartInfo startInfo = new ProcessStartInfo();
+            if (File.Exists(scriptPath))
+            {
+                startInfo.FileName = scriptPath;
+            }
+            else if (File.Exists(Path.Combine(buildHookFolder, psScript)))
+            {
+                startInfo.FileName = "PowerShell";
```
ravimeda: If PowerShell is not available on the machine, which means either of the following cases -

1. Neither `PowerShell` nor `pwsh` is available
2. `pwsh` is available but not `PowerShell`

In either of those cases, the error would be -

```
-- EXECUTING: PowerShell -NoProfile -File pre-build.ps1
System.ComponentModel.Win32Exception (0x80004005): No such file or directory
```

This error is a bit misleading since my first guess was `pre-build.ps1` is not available/accessible. Only later I realized PowerShell was not installed on the machine.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+        private void InvokeBuildHook(string hookName, string buildContextPath)
+        {
+            string buildHookFolder = Path.Combine(buildContextPath, "hooks");
+            if (!Directory.Exists(buildHookFolder))
+            {
+                return;
+            }
+
+            string scriptPath = Path.Combine(buildHookFolder, hookName);
+            string psScript = $"{hookName}.ps1";
```
MichaelSimons: It would be easy to defer constructing this in the case it is not needed.  Also I would suggest using Path.ChangeExtension.  It would be more consistent with Path.Combine and is more robust.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+            else if (File.Exists(Path.Combine(buildHookFolder, psScript)))
+            {
+                startInfo.FileName = "PowerShell";
+                startInfo.Arguments = $"-NoProfile -File {psScript}";
+            }
+            else
+            {
+                return;
+            }
+            startInfo.WorkingDirectory = buildHookFolder;
```
MichaelSimons: Consider a blank line after the else block.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +72,34 @@ private void BuildImages()
...
+            ExecuteHelper.Execute(startInfo, Options.IsDryRun, $"Failed to execute '{hookName}' hook in '{buildHookFolder}'");
```
MichaelSimons: Consider a pattern where the fully qualified hook is used - e.g. '{scriptPath}'.  I think it is easier to decipher.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
MichaelSimons: Doesn't this need to support `pwsh` in order to work with the latest *nix PS versions?

ravimeda: Updated as - if Windows use `PowerShell` else `pwsh`

---
In reply to: [158570976](https://github.com/dotnet/docker-tools/pull/66#discussion_r158570976) [](ancestors = 158570976)

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +73,35 @@ private void BuildImages()
...
+            if (File.Exists(scriptPath))
+            {
+                startInfo = new ProcessStartInfo(scriptPath);
+            }
+            else
+            {
+                scriptPath = Path.ChangeExtension(scriptPath, ".ps1");
+                if (!File.Exists(scriptPath))
+                {
+                    Console.WriteLine($"Failed to locate build hook {scriptPath}'");
```
MichaelSimons: I don't think this should be emitting any type of message.  It is no different than the return on line 81 - this should not be considered a `failure`.  For example, it will be distracting to see this message if there is a pre-build hook but no post-build hook.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +73,35 @@ private void BuildImages()
...
+                    return;
+                }
+                startInfo = new ProcessStartInfo(
+                    RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "PowerShell" : "pwsh", 
+                    $"-NoProfile -File \"{scriptPath}\"");
+            }
+            ExecuteHelper.Execute(startInfo, Options.IsDryRun, $"Failed to execute build hook '{scriptPath}'");
```
MichaelSimons: IMO a blank line is justified before this.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -69,6 +73,35 @@ private void BuildImages()
...
+            }
+            else
+            {
+                scriptPath = Path.ChangeExtension(scriptPath, ".ps1");
+                if (!File.Exists(scriptPath))
+                {
+                    Console.WriteLine($"Failed to locate build hook {scriptPath}'");
+                    return;
+                }
+                startInfo = new ProcessStartInfo(
```
MichaelSimons: IMO a blank line is justified before this.

PR #68, opened 2018-01-05
Add support to login to registry server other than Docker Hub

MichaelSimons:
This is needed in order to use ImageBuilder to push images to an Azure Container Registry.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -117,11 +117,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             {
                 Utilities.WriteHeading("PUSHING IMAGES");
 
-                if (Options.Username != null)
-                {
-                    DockerHelper.Login(Options.Username, Options.Password, Options.Server, Options.IsDryRun);
-                }
-
+                DockerHelper.Login(Options.Username, Options.Password, Options.Server, Options.IsDryRun);
                 try
                 {
                     foreach (string tag in BuiltTags)
@@ -131,10 +127,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 }
                 finally
                 {
-                    if (Options.Username != null)
-                    {
-                        DockerHelper.Logout(Options.Server, Options.IsDryRun);
-                    }
+                    DockerHelper.Logout(Options.Server, Options.IsDryRun);
                 }
             }
         }
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -119,17 +119,22 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
                 if (Options.Username != null)
                 {
-                    DockerHelper.Login(Options.Username, Options.Password, Options.IsDryRun);
+                    DockerHelper.Login(Options.Username, Options.Password, Options.Server, Options.IsDryRun);
                 }
 
-                foreach (string tag in BuiltTags)
+                try
                 {
-                    ExecuteHelper.ExecuteWithRetry("docker", $"push {tag}", Options.IsDryRun);
+                    foreach (string tag in BuiltTags)
+                    {
+                        ExecuteHelper.ExecuteWithRetry("docker", $"push {tag}", Options.IsDryRun);
+                    }
                 }
-
-                if (Options.Username != null)
+                finally
                 {
-                    ExecuteHelper.Execute("docker", $"logout", Options.IsDryRun);
+                    if (Options.Username != null)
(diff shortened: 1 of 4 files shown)


Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
```diff
@@ -96,15 +96,19 @@ public static void Login(string username, string password, bool isDryRun)
             }
             else
             {
-                string loginArgsWithoutPassword = $"login -u {username} -p";
                 ExecuteHelper.Execute(
                     "docker",
-                    $"{loginArgsWithoutPassword} {password}",
+                    $"login -u {username} -p {password} {server}",
```
ravimeda: Clarifying questions - Should we check if the server is available before attempting a login? Is there a default timeout or number of retries for the login command?
It wasn't clear from the documentation (https://docs.docker.com/engine/reference/commandline/login/#options) hence asking.

Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
```diff
@@ -21,48 +21,61 @@ public PublishManifestCommand() : base()
...
-                    manifestYml.AppendLine($"      architecture: {platform.Model.Architecture.ToString().ToLowerInvariant()}");
-                    manifestYml.AppendLine($"      os: {platform.Model.OS.ToString().ToLowerInvariant()}");
-                    if (platform.Model.Variant != null)
-                    {
-                        manifestYml.AppendLine($"      variant: {platform.Model.Variant}");
-                    }
+                    Console.WriteLine($"-- PUBLISHING MANIFEST:{Environment.NewLine}{manifest}");
+                    File.WriteAllText("manifest.yml", manifest);
+
+                    // ExecuteWithRetry because the manifest-tool fails periodically with communicating
```
ravimeda: Nit - `periodically while communicating...`

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -119,17 +119,22 @@ private void PushImages()
...
+                    {
+                        ExecuteHelper.ExecuteWithRetry("docker", $"push {tag}", Options.IsDryRun);
+                    }
                 }
-
-                if (Options.Username != null)
+                finally
                 {
-                    ExecuteHelper.Execute("docker", $"logout", Options.IsDryRun);
+                    if (Options.Username != null)
```
ravimeda: Nit - Checking against `null` seems to be the pattern. But I suggest `IsStringNullOrEmpty(Options.Username)`

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -119,17 +119,22 @@ private void PushImages()
...
+                    DockerHelper.Login(Options.Username, Options.Password, Options.Server, Options.IsDryRun);
                 }
 
-                foreach (string tag in BuiltTags)
+                try
                 {
-                    ExecuteHelper.ExecuteWithRetry("docker", $"push {tag}", Options.IsDryRun);
+                    foreach (string tag in BuiltTags)
+                    {
+                        ExecuteHelper.ExecuteWithRetry("docker", $"push {tag}", Options.IsDryRun);
```
ravimeda: Clarifying question - `ExecuteWithRetry` does this throw an exception only after exhausting all retries? If it isn't the case, which means `ExecuteWithRetry` can throw on first failed attempt, then I think we should pull try into the foreach block. Something like this -

```
foreach ()
{
try { }
}
Logout
```

Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
MichaelSimons: Fixed

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
MichaelSimons: I removed this check completely because the username/password are required when performing a push.  I logged https://github.com/dotnet/docker-tools/issues/69 to add up front validation for this scenario.

Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
MichaelSimons: Login will validate the server.  I have never seen login fail where it would have passed with a retry.  leaving as is for now.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
MichaelSimons: ExecuteWithRetry will catch any exceptions and retry until the max number of retries is hit at which the exception will bubble up to this code.  I think the current pattern is correct.

PR #70, opened 2018-01-08
Add support for modeled docker build args

MichaelSimons:
This also includes variable support for referencing tags - e.g. $(tagRef:debian-8.2)

Sample Manifest Usage:

```
{
  "platforms": [
    {
      "dockerfile": "debian/8.2",
      "os": "linux",
      "tags": {
        "debian-8.2-$(DockerfileGitCommitSha)-$(TimeStamp)": {
          "id": "debian-8.2"
        }
      }
    }
  ]
},
{
  "platforms": [
    {
      "buildArgs": {
        "BASE_IMAGE":"$(tagRef:debian-8.2)"
      },
      "dockerfile": "debian/8.2/debpkg",
      "os": "linux",
      "tags": {
        "debian-8.2-debpkg-$(system:DockerfileGitCommitSha)-$(system:TimeStamp)": {}
      }
    }
  ]
},
```

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -54,10 +54,12 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                             .Select(tag => tag.FullyQualifiedName)
                             .Concat(platformTags);
                         string tagArgs = $"-t {string.Join(" -t ", allTags)}";
+                        string buildArgs = GetDockerBuildArgs(platform);
+
                         InvokeBuildHook("pre-build", platform.BuildContextPath);
                         ExecuteHelper.Execute(
                             "docker",
-                            $"build {tagArgs} -f {dockerfilePath} {platform.BuildContextPath}",
+                            $"build {tagArgs} -f {dockerfilePath}{buildArgs} {platform.BuildContextPath}",
                             Options.IsDryRun);
                         InvokeBuildHook("post-build", platform.BuildContextPath);
                         BuiltTags = BuiltTags.Concat(platformTags);
@@ -71,6 +73,17 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                     }
                 }
             }
+        }
+
+        private string GetDockerBuildArgs(PlatformInfo platform)
+        {
+            IEnumerable<string> buildArgs = platform.BuildArgs
+                .Select(buildArg => 
+                    {
+                        string buildArgValue = Utilities.SubstituteVariables(null, buildArg.Value, Manifest.GetReferenceVariableValue);
+                        return $" --build-arg {buildArg.Key}={buildArgValue}";
+                    });
+            return string.Join(string.Empty, buildArgs);
         }
 
         private void InvokeBuildHook(string hookName, string buildContextPath)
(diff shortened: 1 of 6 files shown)


Microsoft.DotNet.ImageBuilder/src/Utilities.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.DotNet.ImageBuilder
 {
     public static class Utilities
     {
-        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w]+)\\)";
+        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w:\\-.]+)\\)";
```
ravimeda: Clarifying question - A word followed by a colon would meet the need.  So, I'm unclear on what `\\-` is for?

Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -15,6 +16,7 @@ public class PlatformInfo
     {
         private static Regex FromRegex { get; } = new Regex(@"FROM\s+(?<fromImage>\S+)");
 
+        public IDictionary<string, string> BuildArgs { get; private set; }
```
ravimeda: `BuildArgs` having the same name in `Platform` class makes it a bit hard to read. I don't have a suggestion at this point..

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
MichaelSimons: We need to allow `-` and `.` in order to use tags as variable names - e.g. `$(tagRef:debian-8.2)`

Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
MichaelSimons: This is the pattern used within the entire ViewModel.  The property names align with the model for the pieces that are logically identical.

Microsoft.DotNet.ImageBuilder/src/Utilities.cs
ravimeda: Ok. Thanks.

Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
```diff
@@ -44,6 +59,21 @@ public IEnumerable<string> GetExternalFromImages()
                 .Distinct();
         }
 
+        public string GetReferenceVariableValue(string variableName)
+        {
+            const string tagRef = "tagRef:";
```
ravimeda: Pascal case - `TagRef`

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -73,6 +75,17 @@ private void BuildImages()
             }
         }
 
+        private string GetDockerBuildArgs(PlatformInfo platform)
+        {
+            IEnumerable<string> buildArgs = platform.BuildArgs
+                .Select(buildArg => 
+                    {
+                        string buildArgValue = Utilities.SubstituteVariables(null, buildArg.Value, Manifest.GetReferenceVariableValue);
```
ravimeda: I think the call to `SubstituteVariables` should down the stack where TagInfo is created. Since by the time Build command is invoked, the tags are fully qualified, meaning variables substituted.

Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
MichaelSimons: Good one - Changed

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
MichaelSimons: Per our discussion - I will address this comment with the variable refactoring I am currently working on.

PR #71, opened 2018-01-09
Variable refactoring

MichaelSimons:
1. Made the `TagVariables` property in the model generally applicable to anywhere variables are used in the model.  As part of this the `TagVariables` property was renamed to `Variables`.
2. Refactored the `build` command's `test-var` option to be generally applicable to any command.  The option was renamed to `var`.
3. Added a prefix to all built in variables to make the manifest easier to read in that it is clear where the variable values come from.
4. Refactored the variable substitution logic so that it is more centrally located within one class.  This refactoring should make it easy to add additional built in variables in the future in addition to adding more places within the manifest that support referencing variables.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
@@ -1,0 +1,100 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.ImageBuilder.Model;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.ImageBuilder.ViewModel
+{
+    public class VariableHelper
+    {
+        private const char BuiltInDelimiter = ':';
+        private const string SystemVariableTypeId = "System";
+        private const string TagVariableTypeId = "TagRef";
+        private const string TimeStampVariableName = "TimeStamp";
+        private const string VariableGroupName = "variable";
+
+        private static string TagVariablePattern = $"\\$\\((?<{VariableGroupName}>[\\w:\\-.]+)\\)";
+        private static string TimeStamp { get; } = DateTime.UtcNow.ToString("yyyymmddhhmmss");
+
+        private Func<string, TagInfo> GetTagById { get; set; }
+        private Manifest Manifest { get; set; }
+        private IDictionary<string, string> OptionVariables { get; set; }
+
+        public VariableHelper(Manifest manifest, IDictionary<string, string> optionVariables, Func<string, TagInfo> getTagById)
+        {
+            GetTagById = getTagById;
+            Manifest = manifest;
+            OptionVariables = optionVariables;
+        }
+
+        public string SubstituteValues(string expression, Func<string, string> getContextBasedSystemValue = null)
+        {
+            foreach (Match match in Regex.Matches(expression, TagVariablePattern))
+            {
+                string variableValue;
+                string variableName = match.Groups[VariableGroupName].Value;
+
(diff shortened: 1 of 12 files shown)


Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            return expression;
+        }
+
+        private string GetBuiltInValue(string variableName, Func<string, string> getContextBasedSystemValue)
+        {
+            string variableValue = null;
+
+            string[] variableNameParts = variableName.Split(BuiltInDelimiter, 2);
```
ravimeda: Consider passing `StringSplitOptions.RemoveEmptyEntries`

Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            return variableValue;
+        }
+
+        private string GetUserValue(string variableName)
+        {
+            if (!OptionVariables.TryGetValue(variableName, out string variableValue)
+                && Manifest.Variables != null)
+            {
+                Manifest.Variables.TryGetValue(variableName, out variableValue);
```
ravimeda: Consider simplifying with null-conditional operator (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-conditional-operators) -

```
if (!OptionVariables.TryGetValue(variableName, out string variableValue))
            {
                Manifest.Variables?.TryGetValue(variableName, out variableValue);
            }
```

Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.ImageBuilder.Model;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.ImageBuilder.ViewModel
```
ravimeda: Clarifying question - can a collection of helpers be in `ViewModel`?

Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
MichaelSimons: I'm not sure that would provide the desired behavior.  With RemoveEmptyEntries, variables like `:System:Timestamp` would be considered valid.  I don't think this is the desired behavior.

Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
MichaelSimons: I don't understand the question.  Are you asking if it is appropriate to have the helper within the ViewModel namespace?  The reason I put it here is because the helper is only used within the ViewModel classes.  It is not referenced outside of the namespace, therefore it belongs in the namespace IMO.

Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
MichaelSimons: Changed

Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
ravimeda: Yes, `StringSplitOptions.RemoveEmptyEntries` would reduce the strictness. Leave as-is.

Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
ravimeda: Yes, and I agree with the reasoning.

PR #74, opened 2018-01-11
Ignore pulling base images if name contains a build-arg

ravimeda:
A base image can contain a [build-arg](https://docs.docker.com/engine/reference/builder/#arg). For example -

```
ARG  CODE_VERSION=latest
FROM base:${CODE_VERSION}
```
Pulling `base:${CODE_VERSION}` would fail if the build-arg is not replaced with the corresponding value. Hence these changes enable _ImageBuilder_ to skip pulling base images that contains a build-arg.

A related change is validating the uniqueness of tag IDs.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
@@ -37,21 +37,33 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
         {
             IEnumerable<string> sharedTags = repo.Images
                 .SelectMany(images => images.SharedTags?.Keys ?? Enumerable.Empty<string>());
-            IEnumerable<string> platformTags = repo.Images
+            IDictionary<string, Tag> platformTags = repo.Images
                 .SelectMany(image => image.Platforms)
-                .SelectMany(platform => platform.Tags.Keys);
+                .SelectMany(platform => platform.Tags)
+                .ToDictionary(t => t.Key, t => t.Value);
             IEnumerable<string> allTags = sharedTags
-                .Concat(platformTags)
+                .Concat(platformTags.Keys)
                 .ToArray();
+            CheckDuplicates(allTags, "tags");
 
-            if (allTags.Count() != allTags.Distinct().Count())
+            IEnumerable<string> tagIds = platformTags
+                .Where(t => t.Value.Id != null)
+                .Select(t => t.Value.Id)
+                .ToArray();
+            CheckDuplicates(tagIds, "tagIDs");
+        }
+
+        private static void CheckDuplicates(IEnumerable<string> source, string type)
+        {
+            if (source.Count() != source.Distinct().Count())
             {
-                IEnumerable<string> duplicateTags = allTags
+                IEnumerable<string> duplicates = source
                     .GroupBy(x => x)
                     .Where(x => x.Count() > 1)
                     .Select(x => x.Key);
+
                 throw new ValidationException(
-                    $"Duplicate tags found: {Environment.NewLine}{string.Join(Environment.NewLine, duplicateTags)}");
+                    $"Duplicate '{type}' found: {Environment.NewLine}{string.Join(Environment.NewLine, duplicates)}");
             }
         }
     }
--- a/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
@@ -35,25 +35,28 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
 
         private static void ValidateUniqueTags(Repo repo)
         {
-            IEnumerable<string> sharedTags = repo.Images
-                .SelectMany(images => images.SharedTags?.Keys ?? Enumerable.Empty<string>());
-            IDictionary<string, Tag> platformTags = repo.Images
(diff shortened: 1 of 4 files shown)


Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -74,13 +74,16 @@ public static PlatformInfo Create(Platform model, string repoName, VariableHelpe
...
+
             if (!fromMatches.Any())
             {
                 throw new InvalidOperationException($"Unable to find a FROM image in {this.DockerfilePath}.");
             }
 
-            FromImages = fromMatches.Select(match => match.Groups["fromImage"].Value).ToArray();
+            FromImages = fromMatches
+                .Select(match => match.Groups["fromImage"].Value)
+                .Where(match => !match.Contains("$")).ToArray();
```
MichaelSimons: Formatting - ToArray should be on it's own line to be consistent with the formatting.  It can easily get overlooked where it is.

MichaelSimons: `Where(match => !match.Contains("$")` - match isn't a good name given the previous select.  Something like `from` would be a better choice.

Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -37,21 +37,33 @@ private static void ValidateUniqueTags(Repo repo)
...
             {
-                IEnumerable<string> duplicateTags = allTags
+                IEnumerable<string> duplicates = source
                     .GroupBy(x => x)
                     .Where(x => x.Count() > 1)
                     .Select(x => x.Key);
+
                 throw new ValidationException(
-                    $"Duplicate tags found: {Environment.NewLine}{string.Join(Environment.NewLine, duplicateTags)}");
+                    $"Duplicate '{type}' found: {Environment.NewLine}{string.Join(Environment.NewLine, duplicates)}");
```
MichaelSimons: I don't think `type` should be quoted.  This isn't a user supplied value.

Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -37,21 +37,33 @@ private static void ValidateUniqueTags(Repo repo)
...
+                .Concat(platformTags.Keys)
                 .ToArray();
+            CheckDuplicates(allTags, "tags");
 
-            if (allTags.Count() != allTags.Distinct().Count())
+            IEnumerable<string> tagIds = platformTags
+                .Where(t => t.Value.Id != null)
+                .Select(t => t.Value.Id)
+                .ToArray();
+            CheckDuplicates(tagIds, "tagIDs");
```
MichaelSimons: `tagIDs` should be two words since it isn't an element name in the json - `tag IDs`

Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -37,21 +37,33 @@ private static void ValidateUniqueTags(Repo repo)
...
+                .SelectMany(platform => platform.Tags)
+                .ToDictionary(t => t.Key, t => t.Value);
             IEnumerable<string> allTags = sharedTags
-                .Concat(platformTags)
+                .Concat(platformTags.Keys)
                 .ToArray();
+            CheckDuplicates(allTags, "tags");
 
-            if (allTags.Count() != allTags.Distinct().Count())
+            IEnumerable<string> tagIds = platformTags
```
MichaelSimons: This should be operating on all tags not just the platform tags.

Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -37,21 +37,33 @@ private static void ValidateUniqueTags(Repo repo)
...
+                .Where(t => t.Value.Id != null)
+                .Select(t => t.Value.Id)
+                .ToArray();
+            CheckDuplicates(tagIds, "tagIDs");
+        }
+
+        private static void CheckDuplicates(IEnumerable<string> source, string type)
```
MichaelSimons: I think the verb used within the methods of this class should be consistent.  Current `Validate` is used.  In that theme `ValidateNoDuplicates` seems more consistent.  BTW I am not as concerned what the verb is, I am more concerned with consistency in the pattern.

Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -35,35 +35,37 @@ private static void ValidateCommonRepoOwner(Manifest manifest)
...
-                .Select(t => t.Value.Id)
+                .Select(kvp => kvp.Value.Id)
+                .Where(id => id != null)
                 .ToArray();
-            CheckDuplicates(tagIds, "tagIDs");
+            ValidateUniqueElements(tagIds, "tag IDs");
         }
 
-        private static void CheckDuplicates(IEnumerable<string> source, string type)
+        private static void ValidateUniqueElements(IEnumerable<string> source, string element)
```
MichaelSimons: Minor, I think `element` is an ambiguous name.  The name makes me think it is somehow one of the elements that is checked for uniqueness.  How about something like `elementsDescription`?

PR #80, opened 2018-01-26
General logging cleanup

### LINKED ISSUE #78: Skip displaying a header if the section has no content (or) display an appropriate message
## Steps to reproduce the issue

1. git clone https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_git/DotNetCore
2. cd dockerfiles
3. ./build.ps1 -ImageBuilderCustomArgs --dry-run

## Expected behavior

Do not show a heading if there is no content in that section. User might think that the testing was skipped because the correct argument was not passed to the script, until the user opens `manifest.json` and finds that there are no test commands. 

Another option is show the heading but include an appropriate message. For example, "No test command was found in manifest".  Same applies to other headers.  For example, if no image was pushed to a registry, then `PUSH IMAGES` can display a message indicating the same. This saves investigation time and avoids confusion for users.

## Actual behavior

`TESTING IMAGES` section is blank. See the console output below 

```
TESTING IMAGES
--------------

IMAGES BUILT
------------
```

## Additional information (e.g. issue happens only occasionally)

Issue happens consistently. Can be reproduced with any other repositories where `manifest.json` does not have a test command.

## Output of `docker version`

```
Client:
 Version:      17.09.0-ce
 API version:  1.32
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:40:09 2017
 OS/Arch:      windows/amd64

Server:
 Version:      17.09.0-ce
 API version:  1.32 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:45:38 2017
 OS/Arch:      linux/amd64
 Experimental: true
```

## Output of `docker info`

```
Containers: 0
 Running: 0
 Paused: 0
 Stopped: 0
Images: 1
Server Version: 17.09.0-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host ipvlan macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 06b9cb35161009dcb7123345749fef02f7cea8e0
runc version: 3f2f8b84a77f73d38244dd690525642a72156c64
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.49-moby
Operating System: Alpine Linux v3.5
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 3.837GiB
Name: moby
ID: 2U3X:7GXP:Y6JF:3NV3:3PIT:K5MQ:RWUT:KANE:YNCT:YWXR:NXKE:A4FK
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): true
 File Descriptors: 18
 Goroutines: 26
 System Time: 2018-01-19T03:19:54.8106644Z
 EventsListeners: 0
Registry: https://index.docker.io/v1/
Experimental: true
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false

```

MichaelSimons:
- Created a Logger class as a place to encapsulate how logging is done
- Refactored various logging output to make it more clear when nothing was done (e.g. no images built)
- Added a summary for the publishManifest command

fixes #78


Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -196,14 +201,14 @@ private bool UpdateDockerfileFromCommands(PlatformInfo platform, out string dock
                 {
                     Regex fromRegex = new Regex($@"FROM\s+{Regex.Escape(fromImage)}[^\S\r\n]*");
                     string newFromImage = DockerHelper.ReplaceImageOwner(fromImage, Options.RepoOwner);
-                    Console.WriteLine($"Replacing FROM `{fromImage}` with `{newFromImage}`");
+                    Logger.WriteMessage($"Replacing FROM `{fromImage}` with `{newFromImage}`");
```
ravimeda: `using System;` can be removed now.

Microsoft.DotNet.ImageBuilder/src/Logger.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        public static void WriteHeading(string heading)
+        {
+            Console.WriteLine();
+            Console.WriteLine(heading);
+            Console.WriteLine(new string('-', heading.Length));
+        }
+
+        public static void WriteMessage()
```
ravimeda: Not sure why we need this overload that does only `Console.WriteLine();`

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
MichaelSimons: removed

Microsoft.DotNet.ImageBuilder/src/Logger.cs
MichaelSimons: The whole point of the Logger class is to encapsulate the logging logic.  It abstracts away the underlying mechanism that is used for logging so that it can change over time.  For example it may be useful to hookup to the .NET tracing infrastructure.  To accomplish this, all logging logic must be encapsulated here including the ability to write a blank line.

Issue #11, opened 2017-06-09
Enable CI on repo


MichaelSimons:
Once #8 is implemented, CI should be enabled on the repo to prevent regressions

PR #82, opened 2018-01-26
Introduce ImageBuilder build script that can be shared with official builds and ci

MichaelSimons:
related to #11


Microsoft.DotNet.ImageBuilder/build.ps1
```diff
@@ -0,0 +1,60 @@
...
+try {
+    if ($activeOS -eq "windows") {
+        $osFlavor = "nanoserver"
+    }
+    else {
+        $osFlavor = "debian"
+    }
+
+    $stableTag = "$($DockerRepo):image-builder-$osFlavor-$((Get-Date -Format yyyyMMddHHmmss).ToLower())"
+    $floatingTag = "image-builder"
```
ravimeda: All these statements are unlikely to throw.  So, can be removed from try block.

Microsoft.DotNet.ImageBuilder/Dockerfile.linux.runner
```diff
@@ -0,0 +1,7 @@
+# Use this Dockerfile to create a Linux PowerShell runner image
+# Usage: docker run --rm -v /var/run/docker.sock:/var/run/docker.sock runner pwsh -File build.ps1 <params....>
+
+FROM microsoft/dotnet-buildtools-prereqs:debian-stretch-docker-testrunner-63f2145-20184325094343
```
ravimeda: `20184325094343` - consider building and picking up a newer image.

MichaelSimons: What is the value in forcing a build when the Dockerfile didn't change?  To get the timestamp fixed?  I would rather not trigger a build when the Dockerfile hasn't changed.

PR #85, opened 2018-01-30
Add --pull to docker build command

MichaelSimons:


MichaelSimons:
@dotnet-bot test ci please

netci.groovy
```diff
@@ -14,7 +14,7 @@ platformList.each { platform ->
                 batchFile("powershell -NoProfile -File .\\Microsoft.DotNet.ImageBuilder\\build.ps1 -CleanupDocker")
             }
             else {
-                shell("docker build -t runner -f ./Microsoft.DotNet.ImageBuilder/Dockerfile.linux.runner .")
+                shell("docker build -t runner -f ./Microsoft.DotNet.ImageBuilder/Dockerfile.linux.runner --pull .")
```
ravimeda: Consider putting options before the path -
`docker build -t runner --pull -f ./Microsoft.DotNet.ImageBuilder/Dockerfile.linux.runner .`

MichaelSimons: What is the value in doing this?  `-t` and `-f` and the more common args.

ravimeda: Prefer going by the help text - `Usage:  docker build [OPTIONS] PATH | URL | -`

MichaelSimons: But that isn't the order you proposed is it?  Your suggestion was the reverse order.  I personally think the tag should be listed first as it is the one identifier you need to know in order to use the resulting image.  Leaving as is.

PR #87, opened 2018-02-20
Refactor username/password option validation

### LINKED ISSUE #69: Add command line validation to ensure credential args are specified
Currently there is no validation logic that runs to ensure username and password are specified when performing operations on a Docker Registry (e.g. pushing images/manifests).  This validation logic should run up front in order to fail early.

MichaelSimons:
Fixes #69

- Add validation to ensure both username and password are specified if one is specified.
- If a username and password are not specified, skip docker registry login.  This is useful when running locally.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryOptions.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryOptions.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.CommandLine;
 
 namespace Microsoft.DotNet.ImageBuilder.Commands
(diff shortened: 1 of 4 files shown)


MichaelSimons:
@dotnet-bot test Debian please

Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryOptions.cs
```diff
@@ -36,11 +36,17 @@ public override void ParseCommandLine(ArgumentSyntax syntax)
...
 
             string username = null;
-            syntax.DefineOption(
+            Argument<string> usernameArg = syntax.DefineOption(
                 "username",
                 ref username,
                 "Username for the Docker Registry the images are pushed to");
             Username = username;
+
+            if (password != null ^ username != null)
```
ravimeda: Is there a case where empty string is allowed? If not replace null-check with `IsNullOrWhiteSpace`

MichaelSimons: After thinking about this, I changed this to `passwordArg.IsSpecified ^ usernameArg.IsSpecified` as this better expresses the intent.  I don't want any logic here on what a valid username/password is.  That logic belongs to Docker.  This is more about making sure both options are used together.

PR #91, opened 2018-02-27
Add logic to handle stage references in FROM instructions

### LINKED ISSUE #88: Logic that pulls external images does not handle multi-stage Dockerfiles correctly
Image-Builder has logic to pull all external image dependencies before the build is started.  The logic that determines if a dependency is external does not handle multi-stage Dockerfiles where a stage is based on a previous stage.  When this happens, the logic tries to pull the stage images.

```
FROM microsoft/dotnet:2.0-sdk AS build
WORKDIR /app

# copy csproj and restore as distinct layers
COPY *.sln .
COPY dotnetapp/*.csproj ./dotnetapp/
COPY utils/*.csproj ./utils/
COPY tests/*.csproj ./tests/
RUN dotnet restore

# copy everything else and build app
COPY . .
WORKDIR /app/dotnetapp
RUN dotnet build


FROM build AS test
WORKDIR /app/tests
RUN dotnet test
```

MichaelSimons:
Fixes #88

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
@@ -98,9 +98,8 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
         {
             bool isStageReference = false;
 
-            for (int i = 0; i < fromMatches.Count; i++)
+            foreach (Match fromMatch in fromMatches)
             {
-                Match fromMatch = fromMatches[i];
                 if (string.Equals(fromImage, fromMatch.Groups[FromImageMatchName].Value, StringComparison.Ordinal))
                 {
                     // Stage references can only be to previous stages so once the fromImage is reached, stop searching.
--- a/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
@@ -14,11 +14,18 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
 {
     public class PlatformInfo
     {
-        private static Regex FromRegex { get; } = new Regex(@"FROM\s+(?<fromImage>\S+)");
+        private const string FromImageMatchName = "fromImage";
+        private const string StageIdMatchName = "stageId";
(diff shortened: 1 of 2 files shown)


Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -73,18 +80,42 @@ public static PlatformInfo Create(Platform model, string repoName, VariableHelpe
...
+                .Select(match => match.Groups[FromImageMatchName].Value)
+                .Where(from => !IsStageReference(from, fromMatches) && !from.Contains("$"))
                 .ToArray();
         }
+
+        private static bool IsStageReference(string fromImage, IList<Match> fromMatches)
+        {
+            bool isStageReference = false;
+
+            for (int i = 0; i < fromMatches.Count; i++)
```
dagood: Why `for`? I don't see the `i` being used except to get the element.

Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -73,18 +80,42 @@ public static PlatformInfo Create(Platform model, string repoName, VariableHelpe
...
             {
-                throw new InvalidOperationException($"Unable to find a FROM image in {this.DockerfilePath}.");
+                throw new InvalidOperationException($"Unable to find a FROM image in {DockerfilePath}.");
             }
 
             FromImages = fromMatches
-                .Select(match => match.Groups["fromImage"].Value)
-                .Where(from => !from.Contains("$"))
+                .Select(match => match.Groups[FromImageMatchName].Value)
+                .Where(from => !IsStageReference(from, fromMatches) && !from.Contains("$"))
```
dagood: Seems like perf could be better for large number of stages, but this is a nicely minimal and clear change. (And how many stages can there even be realistically?)

MichaelSimons: Yes - this could be written in a way which is faster but I settled on this approach for improved readability.  The most number of stages in our repos is 5, the majority is 1 or 2.

Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
MichaelSimons: A remnant of refactoring - will change to foreach

PR #92, opened 2018-03-02
Refactor GenerateTagsReadmeCommand to support user defined templates

### LINKED ISSUE #89: Add flexible support for sorting/grouping the readme tags
Currently the readme tag order is determined by the readmeOrder property if specified else the order of the images in the manifest.  Using the readmeOrder property can get clunky to maintain.  For example adding a new image can require all readmeOrder properties to need to be incremented by one.  

The current functionality doesn't have flexibility in grouping.  With the introduction of the .NET Core 2.1 preview images, there is discussion of wanting a preview section.  Creating built in sort/grouping logic to handle all these changing scenarios is not feasible/maintainable nor is this something the tool should be responsible for.

A more general solution should be implemented in which the individual repos have complete control and still benefit from having a tool generate the tag information.  Without this, the tag information is prone to errors such as incorrect/missing images and tags as well as broken hyperlinks, etc.

MichaelSimons:
Fixes #89 

I decided to try a readme template approach.  I'm not really in love with this but it does address the issues discussed in #89.  You can see an example [here](https://github.com/MichaelSimons/dotnet-docker/blob/issue-337/manifest.json#L520).  I had thought about using something like a T4 template but it is far more complicated and expensive to implement/maintain than I felt was needed.   I would like to keep this as simple as possible, yet allow PMs/repo owners the flexibility in changing the readme format yet still get the benefit of having a tool generate the individual tag lines.

@natemcmaster - I would like to get your feedback on this.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateTagsReadmeCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateTagsReadmeCommand.cs
@@ -32,8 +32,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                     .Where(info => info.DocumentedTags.Any())
                     .ToList();
 
-                string tagsDoc = repo.Model.ReadmeTemplate == null ?
-                    GetDefaultDocumentation() : GetTemplateBasedDocumentation(repo.Model.ReadmeTemplate);
+                string tagsDoc = Options.Template == null ?
+                    GetManifestBasedDocumentation() : GetTemplateBasedDocumentation(Options.Template);
 
                 Logger.WriteSubheading($"{repo.Name} Tags Documentation:");
                 Logger.WriteMessage();
@@ -65,7 +65,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             return displayName;
         }
 
-        private string GetDefaultDocumentation()
+        private string GetManifestBasedDocumentation()
         {
             StringBuilder tagsDoc = new StringBuilder();
 
@@ -83,7 +83,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 tagsDoc.AppendLine();
 
                 IEnumerable<string> tagLines = platformGroup
-                    .Select(info => GetImageDocumentation(info))
+                    .Select(info => GetTagDocumentation(info))
                     .Where(doc => doc != null);
                 tagsDoc.AppendLine(string.Join(Environment.NewLine, tagLines));
                 tagsDoc.AppendLine();
@@ -116,7 +116,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             return displayName;
         }
 
-        private string GetImageDocumentation(ImageDocumentationInfo info)
+        private string GetTagDocumentation(ImageDocumentationInfo info)
         {
             string tags = info.DocumentedTags
                 .Select(tag => $"`{tag}`")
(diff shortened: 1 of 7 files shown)


natemcmaster:
I'm perfectly happy with the current way we generate tags in the readme. It seems like this may actually make it harder to work with. As an alternative, you might consider using an existing templating language/tool such as [mustache](https://mustache.github.io/mustache.5.html)

MichaelSimons:
The current approach has worked well for repos like microsoft/aspnetcore but it hasn't been great for microsoft/dotnet.  It not longer meets our needs as soon as we want to have separate groupings for preview images and sort the tags in a non-deterministic way such as by usage (given the current metadata).   The current path is just not the answer IMO.  I've looked at mustache before but it didn't have a .net core or cross platform solution which was a turnoff.

dagood:
What about Razor? I have no idea how difficult it is to use outside the context of ASP.NET, but it certainly seems like it would be a good fit.

natemcmaster:
Razor is unfortunately difficult to use outside of aspnet.

MichaelSimons:
I factored out the template into a separate file.  I think this will make the maintenance a little better.  

Note the template is completely optional and the existing functionality remains the same if it suites your needs.  The template technology can be swapped out at any time.  I just don't see a good xplat .net core solution at this point in time therefore I would like to proceed with the basic functionality proposed which makes use of existing Image-Builder functionality.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateTagsReadmeCommand.cs
```diff
@@ -57,20 +65,12 @@ private static string GetArchitectureDisplayName(Architecture architecture)
...
-                .Select(tag => tag.Name);
-        }
-
-        private string GetTagsDocumentation(RepoInfo repo)
+        private string GetManifestBasedDocumentation()
         {
             StringBuilder tagsDoc = new StringBuilder();
 
-            var platformGroups = repo.Images
-                .OrderBy(image => image.Model.ReadmeOrder)
```
dagood: > Note the template is completely optional and the existing functionality remains the same if it suites your needs.

Maybe I'm taking this too literally, but how will `ReadmeOrder` be used after this change? I don't see other usages in this file.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateTagsReadmeCommand.cs
```diff
@@ -128,6 +116,48 @@ private static string GetOsDisplayName(OS os, string osVersion)
...
+            }
+
+            return tagsDoc;
+        }
+
+        private string GetVariableValue(string variableType, string variableName)
+        {
+            string variableValue = null;
+
+            if (string.Equals(variableType, "TagDoc", StringComparison.Ordinal))
```
dagood: It's only used here for now, but I think `"TagDoc"` should be a const string for futureproofing because the other types are already const strings.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateTagsReadmeCommand.cs
MichaelSimons: Good catch - I meant to remove the property from the model since microsoft/dotnet was the only repo that was using this functionality. Fixed

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateTagsReadmeCommand.cs
MichaelSimons: Fixed.

PR #98, opened 2018-05-22
Update generateTagsReadme command to handle new windows versions

### LINKED ISSUE #97: GenerateTagsReadme command does not support Windows Server version 1803
The current header that gets generated for Windows Server, version 1803 tags is 

# Windows Server 2016 amd64 tags

It should be

# Windows Server, version 1803 amd64 tags

MichaelSimons:
Fixes #97 

This is a change that will hopefully cover the next versions of Windows Server assuming they stay with the `Windows Server, version N` naming scheme. 

Note: we are now only setting the version number in the manifests to the shortened stable release version - e.g. 1803, 1709, etc.)

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateTagsReadmeCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateTagsReadmeCommand.cs
@@ -99,13 +99,13 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             switch (os)
             {
                 case OS.Windows:
-                    if (osVersion != null && (osVersion.Contains("1709") || osVersion.Contains("16299")))
+                    if (osVersion == null || osVersion.Contains("2016"))
                     {
-                        displayName = "Windows Server, version 1709";
+                        displayName = "Windows Server 2016";
                     }
                     else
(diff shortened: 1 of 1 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/GenerateTagsReadmeCommand.cs
```diff
@@ -99,13 +99,13 @@ private static string GetOsDisplayName(OS os, string osVersion)
             switch (os)
             {
                 case OS.Windows:
-                    if (osVersion != null && (osVersion.Contains("1709") || osVersion.Contains("16299")))
+                    if (osVersion == null || osVersion.Contains("2016"))
```
dagood: Any concern `osVersion` will make it to `2016` (and hit `2016` rather than pass it) before this code changes again? :smile:

MichaelSimons: No - After `Windows Server 2016`, the version number syntax is YYMM.  `Windows Server, version 1803` was released March of 2018.  Unless they add four more months to the calendar, `2016` is not possible 😃

PR #99, opened 2018-05-31
Add support for referencing the command options as variables in the manifest

MichaelSimons:


Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
```diff
@@ -84,12 +81,16 @@ private string GetBuiltInValue(string variableName, Func<string, string, string>
                 }
                 else if (string.Equals(variableName, RepoOwnerVariableName, StringComparison.Ordinal))
                 {
-                    variableValue = variableValue = RepoOwnerOverride ?? DockerHelper.GetImageOwner(Manifest.Repos.First().Name);
+                    variableValue = variableValue = Options.RepoOwner ?? DockerHelper.GetImageOwner(Manifest.Repos.First().Name);
```
dseefeld: Why is variableValue specified twice here?

MichaelSimons: Silly mistake - fixed.

PR #104, opened 2018-06-26
Add copyImages command

MichaelSimons:
This command is intended to be used to copy images from a staging repo to the public repo (e.g. Docker Hub).

The primary changes are in CopyImagesCommand.cs and CopyImagesOptions.cs.  The rest of the changes are misc refactorings to share common code with existing commands.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/CopyImagesOptions.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/CopyImagesOptions.cs
@@ -1,0 +1,97 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.CommandLine;
+using Microsoft.DotNet.ImageBuilder.Model;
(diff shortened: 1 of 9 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/CopyImagesOptions.cs
```diff
@@ -0,0 +1,97 @@
...
+                "source-username",
+                ref sourceUsername,
+                "Username for the Docker Registry the images are pulled from");
+            SourceUsername = sourceUsername;
+
+            string sourceRepo = null;
+            syntax.DefineParameter("source-repo", ref sourceRepo, "Docker repository to pull images from");
+            SourceRepo = sourceRepo;
+
+            if (destinationPasswordArg.IsSpecified ^ destinationUsernameArg.IsSpecified)
```
dseefeld: What happens if both are not specified?

MichaelSimons: Then it assumes anonymous access and no explicit login is executed.  This is the code path for a public repo on Docker Hub.

PR #112, opened 2018-08-24
Add generateBuildQueue command

MichaelSimons:
Added a command that reads the manifest and generates the VSTS build queue variables.  The generated legs are the smallest possible dependency graphs as defined by the intra-repo Dockerfile depencencies.

Currently the output from this command can be copied and pasted into the VSTS build definitions.  Eventually I would like to invoke this command directly from the build to make everything dynamic.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildQueueCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildQueueCommand.cs
@@ -59,6 +59,10 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             return Task.CompletedTask;
         }
 
+        /// <summary>
+        /// Formats a build leg name from the specified Dockerfile path. Any parts of the Dockerfile path that are in common with the
+        /// containing queue name are trimmed. The resulting leg name uses '-' characters as word separators.
+        /// </summary>
         private static string FormatLegName(string[] dockerfilePath, string[] queueNameParts)
         {
             string legName = dockerfilePath.First().Split('/')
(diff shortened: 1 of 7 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildQueueCommand.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            if (dockerfilePath.Length > 1)
+            {
+                legName += "-graph";
+            }
+
+            return legName;
+        }
+
+        private static string FormatQueueName(string[] parts)
```
dseefeld: Could you join all parts with '-'  first and then use linq to simplify this?  Plus, a comment stating the format you want would be helpful.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildQueueCommand.cs
```diff
@@ -0,0 +1,104 @@
...
+                        .Select(path => $"--path {path}")
+                        .Aggregate((working, next) => $"{working} {next}");
+                    Logger.WriteMessage($"      imageBuilderPaths: {pathArgs}");
+                }
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private static string FormatLegName(string[] dockerfilePath, string[] queueNameParts)
```
dseefeld: A comment here with the desired format of the leg name would be helpful.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildQueueCommand.cs
MichaelSimons: How's this?

```
        /// <summary>
        /// Formats a queue name by joining the specified parts. The resulting queue name is camelCased.
        /// Any '-' occurrences within the specified parts will be treated as word boundaries.
        /// </summary>
        private static string FormatQueueName(string[] parts)
        {
            string[] allParts = parts.SelectMany(part => part.Split('-')).ToArray();
            for (int i = 1; i< allParts.Length; i++)
            {
                string part = allParts[i];
                allParts[i] = char.ToUpper(part[0]) + part.Substring(1);
            }

            return string.Join(string.Empty, allParts);
        }
```

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildQueueCommand.cs
MichaelSimons: Added

Microsoft.DotNet.ImageBuilder/src/GraphExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            foreach (T item in source)
+            {
+                if (!nodes.TryGetValue(item, out Node<T> itemNode))
+                {
+                    itemNode = new Node<T>() { Item = item };
+                    nodes.Add(item, itemNode);
+                }
+
+                foreach (T parent in getDependencies(item))
```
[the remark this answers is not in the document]
MichaelSimons: I originally liked `getDependencies` in the public API but the more I think about it, it is a little abstract so perhaps get parents is better.  I don't have a strong opinion - changed.

PR #121, opened 2018-10-29
Remove test command support

MichaelSimons:
This functionality is no longer being used and should be cleaned up.


MichaelSimons:
@dseefeld could you review this please?

PR #133, opened 2019-01-14
Add CopyAcrImages command to support fast coping within Azure

MichaelSimons:
This is being added to support moving to our Docker repos to MCR.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
@@ -1,0 +1,42 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.DotNet.ImageBuilder.ViewModel;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class CopyAcrImagesCommand : Command<CopyAcrImagesOptions>
+    {
+        public CopyAcrImagesCommand() : base()
+        {
+        }
+
+        public override Task ExecuteAsync()
+        {
+            Logger.WriteHeading("COPING IMAGES");
+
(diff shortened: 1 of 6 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
```diff
@@ -0,0 +1,42 @@
...
+                    .SelectMany(image => image.ActivePlatforms)
+                    .SelectMany(platform => platform.Tags);
+                string fullRegistryName = $"{Options.Registry}.azurecr.io/";
+
+                foreach (TagInfo platformTag in platformTags)
+                {
+                    string sourceImage = $"{Options.SourceRepository}:{platformTag.Name}";
+                    string destImage = platformTag.FullyQualifiedName.TrimStart(fullRegistryName);
+                    helper.ExecuteAzCommand(
+                        $"acr import -n {Options.Registry} --source {sourceImage} -t {destImage} --force",
```
dseefeld: Are you always going to want this to be --force? Or should that be an option?

Microsoft.DotNet.ImageBuilder/src/StringExtensions.cs
```diff
@@ -6,9 +6,24 @@ namespace Microsoft.DotNet.ImageBuilder
 {
     public static class StringExtensions
     {
-        public static string TrimEnd(this string source, string value)
+        public static string TrimEnd(this string source, string trimString)
         {
-            return source.EndsWith(value) ? source.Remove(source.LastIndexOf(value)) : source;
+            while (source.EndsWith(trimString))
```
dseefeld: I don't understand the usage of "while" here.  If there are multiple occurrences of trimString, they will all be removed?  When do you need that?

MichaelSimons: I thought it was beneficial to mirror the existing `TrimStart(char)` and `TrimEnd(char)` methods from the Fx which trim multiple occurrences.  I thought this was valuable given they are named the same and are extension methods.  The current usage does't require this functionality.  I could be convinced to modify it if you feel strongly.

MichaelSimons:
@dseefeld - Thanks for reviewing!

PR #141, opened 2019-02-07
Add validateImageSizes command

MichaelSimons:
The command adds support for creating an image size baseline and then validating image sizes against the baseline.

Sample Usage:

Create baseline - `validateImageSize --update --pull ImageSize.linux.json`
Validate local images against baseline `validateImageSize ImageSize.linux.json`


Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
```diff
@@ -175,5 +188,11 @@ public static string ReplaceRepo(string image, string newRepo)
         {
             return newRepo + image.Substring(image.IndexOf(':'));
         }
+
+        private static bool ResourceExists(string type, string filterArg, bool isDryRun)
+        {
+            string output = ExecuteCommand($"{type} ls -a -q {filterArg}", "Failed to find resource", isDryRun: isDryRun);
```
dseefeld: The "ls" options seem specific to containers and images.  Is this meant to be used for other resources as well?  Should type be an enum instead to limit it to just the supported resources?

MichaelSimons: 'ls' applies to a number of the "management types" -e.g. volume, network, config, node, etc.  This ResourceExists method could be used for any of them.  The method is private now and only used in one place but could get used in more scenarios over time.  I added an enum as suggested but only added `container` and `image`.

PR #144, opened 2019-02-15
Refactoring validateImageSize output

MichaelSimons:
Misc refactoring to make the output more friendly to scan.  The results are no longer intermixed with the docker pull output.  I also added more classification to the results - exact match, in range variance, out of range variance, no baseline data.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
@@ -168,7 +168,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             public double? MaxVariance => BaselineSize + AllowedVariance;
             public double? MinVariance => BaselineSize - AllowedVariance;
             public long? SizeDifference => CurrentSize - BaselineSize;
-            public bool WithinAllowedVariance => AllowedVariance > Math.Abs(SizeDifference.Value);
+            public bool WithinAllowedVariance => BaselineSize.HasValue && AllowedVariance > Math.Abs(SizeDifference.Value);
         }
     }
 }
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
@@ -42,12 +42,11 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 Logger.WriteMessage,
                 "Images with no size change:");
             LogResults(
-                baselinedImageData.Where(info =>
-                    info.SizeDifference != 0 && info.AllowedVariance > Math.Abs(info.SizeDifference.Value)),
+                baselinedImageData.Where(info => info.SizeDifference != 0 && info.WithinAllowedVariance),
                 Logger.WriteMessage,
                 "Images with allowed size change:");
             LogResults(
-                baselinedImageData.Where(info => info.AllowedVariance < Math.Abs(info.SizeDifference.Value)),
+                baselinedImageData.Where(info => !info.WithinAllowedVariance),
                 Logger.WriteError,
                 "Images exceeding size variance:");
             LogResults(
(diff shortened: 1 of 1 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
```diff
@@ -32,12 +32,54 @@ public override Task ExecuteAsync()
...
+            IEnumerable<ImageInfo> baselinedImageData = imageData.Where(info => info.BaselineSize.HasValue).ToArray();
+
+            Logger.WriteHeading("VALIDATION RESULTS");
+            LogResults(
+                baselinedImageData.Where(info => info.SizeDifference == 0),
+                Logger.WriteMessage,
+                "Images with no size change:");
+            LogResults(
+                baselinedImageData.Where(info =>
+                    info.SizeDifference != 0 && info.AllowedVariance > Math.Abs(info.SizeDifference.Value)),
```
dseefeld: Can this calculation be moved to a boolean property on info?  i.e. info.InVariance?

Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
```diff
@@ -32,12 +32,54 @@ public override Task ExecuteAsync()
...
                 Logger.WriteSubheading(header);
-                messages.ForEach(msg => Logger.WriteError(msg));
+
+                foreach (ImageInfo info in imageData)
+                {
+                    string msg = $"{info.Id}{Environment.NewLine}"
+                        + $"    Actual:     {info.CurrentSize,15:N0}{Environment.NewLine}";
+                    if (info.BaselineSize.HasValue)
+                    {
+                        double? minVariance = info.BaselineSize - info.AllowedVariance;
```
dseefeld: Can these be added as properties on info?

Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
```diff
@@ -94,50 +136,41 @@ private void ValidateImages()
...
-                    if (imageSize < allowedMin || imageSize > allowedMax)
-                    {
-                        invalidImages.Add(msg);
-                    }
-                }
+                    Id = imageId,
+                    CurrentSize = imageSize,
+                    BaselineSize = baseline,
+                    AllowedVariance = baseline * (double)Options.AllowedVariance / 100,
+                    SizeDifference = imageSize - baseline
```
dseefeld: Can this just be the getter for SizeDifference in ImageInfo?

Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
MichaelSimons: Yes - added.

Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
MichaelSimons: yes - changed.

Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
MichaelSimons: yes - I added a WithinAllowedVariance property

MichaelSimons:
Thanks for the great feedback @dseefeld.

Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
```diff
@@ -94,50 +132,43 @@ private void ValidateImages()
...
+        private class ImageInfo
+        {
+            public double? AllowedVariance { get; set; }
+            public long? BaselineSize { get; set; }
+            public long CurrentSize { get; set; }
+            public string Id { get; set; }
+            public double? MaxVariance => BaselineSize + AllowedVariance;
+            public double? MinVariance => BaselineSize - AllowedVariance;
+            public long? SizeDifference => CurrentSize - BaselineSize;
+            public bool WithinAllowedVariance => AllowedVariance > Math.Abs(SizeDifference.Value);
```
dseefeld: Will there be a time when these nullable types don't have a value?  These will throw if they don't.

MichaelSimons: Only the WithinAllowedVariance would throw because the arithmetic operators overloaded for nullable types.  

It is unnecessary per current usage but in general a good idea - I fixed WithinAllowedVariance.

PR #152, opened 2019-02-28
Update build command to update architecture metadata if necessary

mthalman:
Ensures the build command on ImageBuilder will produce an image whose architecture matches the expected value in case Docker decides to misrepresent it.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.CommandLine;
 using System.Diagnostics;
@@ -11,6 +12,8 @@ using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.DotNet.ImageBuilder.Commands
 {
@@ -68,6 +71,11 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                             ExecuteHelper.Execute("docker", dockerArgs, Options.IsDryRun);
                         }
 
+                        if (!Options.IsDryRun)
+                        {
+                            EnsureArchitectureMatches(platform);
+                        }
+
                         InvokeBuildHook("post-build", platform.BuildContextPath);
                         BuiltTags = BuiltTags.Concat(platform.Tags);
                     }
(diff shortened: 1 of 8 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -84,6 +92,75 @@ private void BuildImages()
             BuiltTags = BuiltTags.ToArray();
         }
 
+        private void EnsureArchitectureMatches(PlatformInfo platform)
+        {
+            foreach (TagInfo tag in platform.Tags)
```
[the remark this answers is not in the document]
mthalman: Yes, I had the same question when I was implementing it and so I had verified that it needs to be done on a per-tag basis.  Your suggestion sounds good.  I'll work on that.

Microsoft.DotNet.ImageBuilder/src/ExecuteHelper.cs
```diff
@@ -4,6 +4,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Text;
```
MichaelSimons: Why is this needed when no source changed in this file?

Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
```diff
@@ -20,20 +20,6 @@ public static class DockerHelper
         public static Architecture Architecture => s_architecture.Value;
         public static OS OS => s_os.Value;
```
MichaelSimons: Shuffling the methods intermixed with code changes makes it rather difficult to review what the concrete changes were made 😸

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -52,10 +55,11 @@ private void BuildImages()
 
                         // Tag the built images with the shared tags as well as the platform tags.
                         // Some tests and image FROM instructions depend on these tags.
-                        IEnumerable<string> platformTags = platform.Tags
-                            .Select(tag => tag.FullyQualifiedName)
-                            .ToArray();
-                        string tagArgs = GetDockerTagArgs(image, platformTags);
+                        IEnumerable<string> allTags = platform.Tags
```
MichaelSimons: Since you are touching this, now would be a great time to cleanup a historical artifact.  I am pretty certain we don't do anything with the shared tags here anymore.  We used to test off of the shared tags long time ago and even had Dockerfile references to them.  We no longer do this therefore tagging the images with the shared tags is pure overhead.  We don't push them today.  IMO it would be a good time to clean this up.

Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
```diff
@@ -198,6 +178,47 @@ private static bool ResourceExists(ManagementType type, string filterArg, bool i
...
+            }
+
+            return architecture;
+        }
+
+        private static string ExecuteCommand(
+            string command, string errorMessage, string additionalArgs = null, bool isDryRun = false)
+        {
+            ProcessStartInfo startInfo = new ProcessStartInfo(
+                "docker", $"{command} {additionalArgs}");
```
MichaelSimons: NIT - consider making the method call one line.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
MichaelSimons: I am retracting this comment.  This suggestion makes sense when running in the context of the official build but not from the build-and-test script perspective.

Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
mthalman: Sorry bout that.  I have an allergic reaction to having private and public methods mixed.

Microsoft.DotNet.ImageBuilder/src/ExecuteHelper.cs
mthalman: Not needed.  Leftovers from my previous change to the file that was reverted.

PR #164, opened 2019-03-18
Migrate pipelines for use in dnceng

mthalman:
* Updates the existing pipeline to work with Hosted Windows Container (at least temporarily until we have another agent pool option for Windows).
* Adds another pipeline intended for PR builds.  This pipeline doesn't end up pushing the images that are built.  Consequently, it doesn't need the referenced variable groups or docker login. 
* Updated the pipelines to not clean up Docker by default since it's not needed for hosted agents.
* Parameterized the shared job so it can be used by both the official and PR build pipelines.
* Deleted the netci.groovy script since we're migrating to the AzDO pipeline for PR validation.


.vsts-pipelines/dotnet-buildtools-image-builder.yml
```diff
@@ -3,6 +3,8 @@ trigger:
   branches:
```
MichaelSimons: Thoughts on renaming this to end in `-official` to help differentiate it from `-pr`?

.vsts-pipelines/jobs/build-image-builder.yml
```diff
@@ -1,22 +1,27 @@
 parameters:
   name: null
   pool: {}
+  pushImages: false
```
MichaelSimons: Do you think `publishImages` would be a better name to separate the implementation away from the real intent?

.vsts-pipelines/jobs/build-image-builder.yml
```diff
@@ -1,22 +1,27 @@
...
-      displayName: Docker Login
+    - ${{ if eq(parameters.pushImages, 'true') }}:
+      - script: docker login -u $(acr.userName) -p $(BotAccount-dotnet-docker-acr-bot-password) $(acr.server)
+        displayName: Docker Login
     - powershell: >
-        Microsoft.DotNet.ImageBuilder/build.ps1
-        -DockerRepo $(acr.server)/public/dotnet-buildtools/image-builder
-        -PushImages
-        -CleanupDocker
+        Microsoft.DotNet.ImageBuilder/build.ps1 $(buildParameters)
```
MichaelSimons: Given the simplification, consider making this a single line.

.vsts-pipelines/dotnet-buildtools-image-builder.yml
mthalman: Sounds good to me.

.vsts-pipelines/jobs/build-image-builder.yml
mthalman: I'm fine with that.

PR #167, opened 2019-03-26
Update build matrix to support PR builds

mthalman:
For the PR builds, we need to be able to build and test within a single job.  But the build matrix that is currently generated is not suited for that scenario because it is grouped on OS/arch.  This splits out the building of images like runtime and sdk that have a dependency on each other.  This won't work for the PR builds because the tests will require the set of images (runtime, sdk, etc) to be available which means that the build phase must have built all of those images in that job.

To accomplish this, we need to just use the same matrix as is generated for tests but use it as a build matrix instead.  A new option is added for the generateBuildMatrix command to indicate that it is for a PR build.  If that option is set, it will return a build matrix that is shaped just like the test matrix when the build matrix is requested.  In addition to that, the test matrix also needs to include some extra metadata that was available in the build matrix: imageBuilderPaths.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
@@ -41,21 +41,32 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             IEnumerable<IEnumerable<PlatformInfo>> subgraphs = platformGrouping.GetCompleteSubgraphs(GetPlatformDependencies);
             foreach (IEnumerable<PlatformInfo> subgraph in subgraphs)
             {
-                string[] dockerfilePaths = subgraph
-                    .Select(platform => platform.Model.Dockerfile)
-                    .ToArray();
+                string[] dockerfilePaths = GetDockerfilePaths(subgraph);
+
                 LegInfo leg = new LegInfo()
-                { 
+                {
                     Name = GetDockerfilePathLegName(dockerfilePaths, platformNameParts, matrixNameParts)
                 };
                 matrix.Legs.Add(leg);
 
-                string pathArgs = dockerfilePaths
-                    .Select(path => $"--path {path}")
-                    .Aggregate((working, next) => $"{working} {next}");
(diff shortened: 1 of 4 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -179,14 +193,25 @@ private IEnumerable<MatrixInfo> GenerateMatrixInfo()
...
-                            break;
+                        }
+                        else
+                        {
+                            AddDockerfilePathLegs(matrix, matrixNameParts, platformGrouping);
+                        }
+                    }
+                    else if (Options.MatrixType == MatrixType.Test)
+                    {
+                        AddVersionedOsLegs(matrix, platformGrouping);
```
MichaelSimons: I'm not crazy about coupling the ImageBuilder more to how it is being used - e.g. a PR build.  It feels like this should be abstracted away.  It's also strange that two different option configurations yield the same output (Test and Build/IsPull).  

Would it be better to decouple the MatrixType from how it is being used in the build all together?  Would it make sense for the names to be in terms of what they produce instead of how they are intended to be used?  Build => DependencyGraphByPlatform, Test => VersionedOSByPlatform, publish => DependencyGraph.  I'm not sure I'm crazy about those names however.  Other options (PlatformDependencyGraph, PlatformVersionedOS, DependencyGraph)

Another option would be to simply use the Test matrix during the build of PR validation and not add the isPullRequest option.  

Another option would be to add a BuildAndTest MatrixType instead of the isPullRequest option.  This would keep the ImageBuilder decoupled from the knowledge of PR builds.  It would still have two options that yield the same result.

mthalman: I see your point.  My main motivation was to simplify the pipeline code to avoid the conditional logic of determining which matrix type it needed.  I do like your suggestion of changing the matrix type names.  That would make it less confusing because otherwise there would be pipeline logic that uses the test matrix for build purposes.  I think renaming things would help with that.

PR #168, opened 2019-04-03
Refactor build to use the same tag timestamp across images

MichaelSimons:
Currently the timestamp portion of the linux and windows images are different.  It would be nice if they aligned as it helps to consumption story.  This change defines a shared timestamp variable that the linux and windows build legs consume.  

I'm open to suggestions to simplify the pattern used here.

### DIFF
--- a/.vsts-pipelines/jobs/build-image-builder-image.yml
+++ b/.vsts-pipelines/jobs/build-image-builder-image.yml
@@ -1,0 +1,29 @@
+parameters:
+  name: null
+  pool: {}
+  publishImages: false
+
+jobs:
+- job: ${{ parameters.name }}
+  pool: ${{ parameters.pool }}
+  dependsOn: Initialize
+  variables:
+    ${{ if eq(parameters.publishImages, 'true') }}:
+      buildParameters: $(commonBuildParameters) -DockerRepo $(acr.server)/public/dotnet-buildtools/image-builder -PushImages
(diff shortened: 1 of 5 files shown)


.vsts-pipelines/jobs/build-image-builder-image.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+- job: ${{ parameters.name }}
+  pool: ${{ parameters.pool }}
+  dependsOn: Initialize
+  variables:
+    ${{ if eq(parameters.publishImages, 'true') }}:
+      buildParameters: $(commonBuildParameters) -DockerRepo $(acr.server)/public/dotnet-buildtools/image-builder -PushImages
+    ${{ if eq(parameters.publishImages, 'false') }}:
+      buildParameters: $(commonBuildParameters)
```
mthalman: This actually works to reference commonBuildParameters here?

MichaelSimons: It works in the test builds.  What are you concerned about?  Is it that commonBuildParameters is declared after this variable?

mthalman: Right, exactly.  I'm just surprised that actually works.  But if it works, it works.

MichaelSimons: I've used this pattern in the past as I always try to alphabetize the variables so they are easier to find.

PR #169, opened 2019-04-08
Add TestDependencyGraph matrix type

mthalman:
In order to build and test the images in a given .NET version/OS leg in a single job, we need to ensure that all of images that are needed in order to tests are included in that leg.  In the case of .NET Core 1.0, the 1.1 SDK image is required in order to run the test.  But the platformVersionedOs matrix that is output by ImageBuilder doesn't include the 1.1 SDK Dockerfile path when generating the set of Dockerfile paths for the 1.0 legs.  The manifest doesn't currently have a mechanism for describing such a dependency that is required solely for testing.

These changes update the manifest schema such that a platform can have a set of test dependencies (i.e. image tags required by the platform in order to test it).  Here's a snippet of the manifest for the 1.0 section illustrating how this new attribute is used:
```
"platforms": [
  {
    "buildArgs": {
      "REPO": "$(Repo:runtime-deps)"
    },
    "dockerfile": "1.0/runtime/jessie/amd64",
    "os": "linux",
    "tags": {
      "$(1.0-RuntimeVersion)-jessie": {},
      "1.0-jessie": {}
    },
    "testDependencies": [
      "dotnet/core-nightly/sdk:1.1-jessie"
    ]
  },
  {
    "dockerfile": "1.0/runtime/nanoserver-1809/amd64",
    "os": "windows",
    "osVersion": "nanoserver-1809",
    "tags": {
      "$(1.0-RuntimeVersion)-nanoserver-1809": {},
      "1.0-nanoserver-1809": {}
    },
    "testDependencies": [
      "dotnet/core-nightly/sdk:1.1-nanoserver-1809"
    ]
  }
]
```
This new attribute will be used when generating a newly defined matrix type of testDependencyGraph.  For example, the jessie-1.0 diagnostic output would look like this:
```
1.0-jessie:
  osType: linux
  architecture: amd64
  osVersion: *
  dotnetVersion: 1.0
  osVariant: jessie
  imageBuilderPaths: --path 1.0/runtime-deps/jessie/amd64 --path 1.0/runtime/jessie/amd64 --path 1.1/sdk/jessie/amd64
```

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
@@ -65,8 +65,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
         private string[] GetCustomLegGroupingDockerfilePaths(IEnumerable<PlatformInfo> platforms)
         {
-            return GetDockerfilePaths(platforms
-                .GetCompleteSubgraphs(platform =>
+            IEnumerable<PlatformInfo> subgraphs = platforms.GetCompleteSubgraphs(platform =>
                 {
                     if (Options.CustomBuildLegGrouping != null &&
                         platform.CustomLegGroupings.TryGetValue(
@@ -76,12 +75,11 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                         return customBuildLegGroupingInfo.DependencyImages
                             .Select(image => Manifest.GetPlatformByTag(image));
                     }
-                    else
-                    {
-                        return Enumerable.Empty<PlatformInfo>();
-                    }
+
+                    return Enumerable.Empty<PlatformInfo>();
                 })
-                .SelectMany(image => image));
+                .SelectMany(image => image);
+            return GetDockerfilePaths(subgraphs);
         }
 
         private static void AddImageBuilderPathsVariable(string[] dockerfilePaths, LegInfo leg)
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
@@ -61,6 +61,13 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 .ToArray();
         }
 
+        private string[] GetTestDependencyDockerfilePaths(IEnumerable<PlatformInfo> platforms)
+        {
+            return GetDockerfilePaths(platforms
+                .GetCompleteSubgraphs(platform => platform.TestDependencyImages.Select(image => Manifest.GetPlatformByTag(image)))
+                .SelectMany(image => image));
+        }
+
         private static void AddImageBuilderPathsVariable(string[] dockerfilePaths, LegInfo leg)
         {
             string pathArgs = dockerfilePaths
@@ -76,7 +83,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             leg.Variables.Add(("osVersion", platformGrouping.Key.OsVersion ?? "*"));
         }
 
-        private static void AddVersionedOsLegs(MatrixInfo matrix, IGrouping<PlatformId, PlatformInfo> platformGrouping)
+        private void AddVersionedOsLegs(MatrixInfo matrix, IGrouping<PlatformId, PlatformInfo> platformGrouping)
         {
             var versionGroups = platformGrouping
                 .GroupBy(platform => new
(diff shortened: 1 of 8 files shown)


Microsoft.DotNet.ImageBuilder/src/Model/Platform.cs
```diff
@@ -30,6 +31,8 @@ public class Platform
         [JsonProperty(Required = Required.Always)]
         public IDictionary<string, Tag> Tags { get; set; }
 
+        public string[] TestDependencies { get; set; } = Array.Empty<string>();
```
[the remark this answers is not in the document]
mthalman: I agree with your point about the naming being a cause for misunderstanding how it's used.  So I think it's probably important to use a name that indicates this is a customization of the graph.  How about a `customBuildLegGrouping` array that lets you define any number of named customizations to the build legs:

```
"platforms": [
  {
    "buildArgs": {
      "REPO": "$(Repo:runtime-deps)"
    },
    "dockerfile": "1.0/runtime/jessie/amd64",
    "os": "linux",
    "tags": {
      "$(1.0-RuntimeVersion)-jessie": {},
      "1.0-jessie": {}
    },
    "customBuildLegGrouping": [
      {
        name: "ci"
        dependencies: [
          "$(Repo:sdk):1.1-jessie"
        ]
      }
    ]
  },
...
```

And for invoking the builder, you can still use any of the existing matrix types and the customization would be applied to any of them if specified:
`generateBuildManifest --type platformVersionedOs --customBuildLegGrouping ci`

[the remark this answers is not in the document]
MichaelSimons: I can get on-board with that.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -61,6 +63,27 @@ private static string[] GetDockerfilePaths(IEnumerable<PlatformInfo> platforms)
...
+                    {
+                        return customBuildLegGroupingInfo.DependencyImages
+                            .Select(image => Manifest.GetPlatformByTag(image));
+                    }
+                    else
+                    {
+                        return Enumerable.Empty<PlatformInfo>();
+                    }
+                })
+                .SelectMany(image => image));
```
MichaelSimons: I'm not seeing why `SelectMany` is needed here?  Am I missing something?

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -61,6 +63,27 @@ private static string[] GetDockerfilePaths(IEnumerable<PlatformInfo> platforms)
                 .ToArray();
         }
 
+        private string[] GetCustomLegGroupingDockerfilePaths(IEnumerable<PlatformInfo> platforms)
+        {
+            return GetDockerfilePaths(platforms
+                .GetCompleteSubgraphs(platform =>
```
MichaelSimons: This indentation confused me for a bit.  I thought `GetCompleteSubgraphs` was getting called on the result of `GetDockerfilePaths`.  IMO it would be easier to read as

```
IEnumerable<IEnumerable<PlatformInfo>> subgraphs = platforms.GetCompleteSubgraphs(...);
return GetDockerfilePaths(subgraphs);
```

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -61,6 +63,27 @@ private static string[] GetDockerfilePaths(IEnumerable<PlatformInfo> platforms)
...
+                {
+                    if (Options.CustomBuildLegGrouping != null &&
+                        platform.CustomLegGroupings.TryGetValue(
+                            Options.CustomBuildLegGrouping,
+                            out CustomBuildLegGroupingInfo customBuildLegGroupingInfo))
+                    {
+                        return customBuildLegGroupingInfo.DependencyImages
+                            .Select(image => Manifest.GetPlatformByTag(image));
+                    }
+                    else
```
MichaelSimons: The else isn't necessary with the return in the if body.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
mthalman: The delegate passed to GetCompleteSubgraphs is returning an enumerable of PlatformInfo, representing the set of dependency images.  GetCompleteSubgraphs is consuming this delegate and walking the graph to produce multiple sets of this PlatformInfo enumerables.  I'm just flattening those multiple sets into one set of platforms and getting their Dockerfiles.  It's basically the same as what is done in AddDockerfilePathLegs, for instance; the only difference is that the code there is for-each'ing the sets and I'm just flattening them.

Issue #165, opened 2019-03-20
Refactor publish commands to run faster and require fewer resources


MichaelSimons:
- [ ] CopyAcrImagesCommand could run many ACR Import requests in parallel.  As a result we should be able to utilize fewer agents to "publish" the images during the docker builds.
- [ ] PublishManifestCommand could run many `manifest-tool push` requests in parallel.

MichaelSimons:
Another point to investigate is whether it'd be beneficial to run all of the az acr import calls in a single call to docker run. Currently, we execute docker run ... az acr import for each tag. There's some extra overhead with doing that.

PR #171, opened 2019-04-18
Refactor PublishManifestCommand to utilize parallel processing

MichaelSimons:
In my local benchmark, this cut the Publish Manifest time for dotnet-docker from 2 min 35 sec down to 37 sec.

Related to #165

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
@@ -27,7 +27,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 IEnumerable<ImageInfo> multiArchImages = Manifest.FilteredRepos
                     .SelectMany(repo => repo.AllImages)
                     .Where(image => image.SharedTags.Any());
-                foreach (ImageInfo image in multiArchImages)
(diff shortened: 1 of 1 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
```diff
@@ -37,7 +37,7 @@ public override Task ExecuteAsync()
                     // ExecuteWithRetry because the manifest-tool fails periodically while communicating
```
mthalman: It seems unsafe to be calling `File.WriteAllText("manifest.yml", manifest);` in parallel.  Probably want to put a lock around that.

mthalman: Actually, I misunderstood what was happening.  You need to be using a unique filename for each image.  It'd also be nice to delete those files when finished with pushing them.

MichaelSimons: Yes you are correct, I realized that after your first comment.

MichaelSimons: Cleaning it up wasn't an issue in our usage because it is running within a container.  That being said it is best to cleanup.  I pushed an update if you want to take a look.

PR #176, opened 2019-04-25
Introduce build definition to validate remote Docker daemon config

MichaelSimons:
This is a build definition that will up us and DDFUN validate the build agents that are configured to use a remote ARM docker daemon.

The build definition validates the following:
1. Remote daemon is reachable (validates the cert configuration) 
1. Remote daemon can run a container and that it has network access.

We can add more to this over time as we see fit.

### DIFF
--- a/.vsts-pipelines/validate-remote-docker-config.yml
+++ b/.vsts-pipelines/validate-remote-docker-config.yml
@@ -1,0 +1,45 @@
+pr: none
+trigger: none
+
+jobs:
+- job: ValidateRemoteDockerConfig
+  pool:
+    ${{ if eq(variables['System.TeamProject'], 'public') }}:
+      name: DotNetCore-Docker-Public
+    ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+      name: DotNetCore-Docker
+    demands:
+    - agent.os -equals linux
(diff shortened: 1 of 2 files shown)


.vsts-pipelines/validate-remote-docker-config.yml
```diff
@@ -0,0 +1,45 @@
...
+      docker run
+      --rm
+      --entrypoint docker
+      -v /var/run/docker.sock:/var/run/docker.sock
+      -v $(DOCKER_CERT_PATH):/docker-certs
+      -e DOCKER_CERT_PATH=/docker-certs
+      -e DOCKER_TLS_VERIFY=1
+      -e DOCKER_HOST=tcp://$(DOCKER_HOST_IP):2376
+      $(imageNames.imageBuilder.linux)
+    imageNames.imageBuilder.linux: mcr.microsoft.com/dotnet-buildtools/prereqs:debian-stretch-docker-testrunner-63f2145-20184325094343
```
mthalman: Can you pull this out into a separate variables template file so we can easily find it when updating image names?

Same for nanoserver and debian images.

MichaelSimons: Added.

PR #189, opened 2019-05-30
Add support for generating MCR tags YAML and producing tags listing from it

MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker/issues/990

- Refactor the GenerateTagsReadmeCommand to generate the MCR tags metadata and use the MCR tool to generate the tags table off of the MCR metadata.
- Refactor the PublishMcrReadmeCommand (renamed to PublishMcrDocsCommand) to generate the MCR tags metadata, strip out the tags listing in the readme, and publish the artifacts to McrDocs
- In support of the previous two commands, two model schema changes were introduced.  First the osVersion is now a required property.  Second the McrTagsMetadataTemplatePath was added to the Repo.

A sample of the output can be view at https://hub.docker.com/_/microsoft-dotnet-core-nightly-aspnet/.
The repo changes to consume this can be viewed at https://github.com/dotnet/dotnet-docker/compare/nightly...MichaelSimons:mcr-tags-table-nightly.

I will be logging a follow-up issue to refactor the GenerateBuildMatricCommand to make use of the osVersion.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
@@ -1,0 +1,44 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Microsoft.DotNet.ImageBuilder.Commands;
+using Microsoft.DotNet.ImageBuilder.Model;
+using Microsoft.DotNet.ImageBuilder.ViewModel;
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    public class ReadmeHelper
+    {
+        private const string TagsSectionHeader = "# Full Tag Listing";
+
+        private static string NormalizeLineEndings(string value, string targetFormat)
+        {
+            string targetLineEnding = targetFormat.Contains("\r\n") ? "\r\n" : "\n";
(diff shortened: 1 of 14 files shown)


Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Microsoft.DotNet.ImageBuilder.Commands;
+using Microsoft.DotNet.ImageBuilder.Model;
+using Microsoft.DotNet.ImageBuilder.ViewModel;
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    public class ReadmeHelper
```
mthalman: >  [](start = 10, length = 1)

static

mthalman: Make this class static.

Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public static string UpdateTagsListing(string readme, string tagsListing)
+        {
+            tagsListing = $"{TagsSectionHeader}{Environment.NewLine}{Environment.NewLine}{tagsListing}{Environment.NewLine}{Environment.NewLine}";
+
+            // Normalize the line endings to match the readme.
+            tagsListing = NormalizeLineEndings(tagsListing, readme);
+
+            Regex regex = new Regex($"^{TagsSectionHeader}\\s*(^(?!# ).*\\s)*", RegexOptions.Multiline);
```
mthalman: Add a comment that provides an example of what the regex matches.  I hate trying to parse them.

PR #190, opened 2019-06-03
Update generateBuildMatrix to ignore OsVersion for grouping Linux entries

MichaelSimons:
1. With the recent changes to add the Linux distro variant info in the Manifest, the build matrix now inadvertently includes the distro in the matrix entry names.  This change will ignore the osVersion for Linux entries.
2. Modified the build matrix logic to retrieve the osVariant test variable from the Manifest's osVersion.


Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -33,11 +33,6 @@ public override Task ExecuteAsync()
         private void AddDockerfilePathLegs(
             MatrixInfo matrix, IEnumerable<string> matrixNameParts, IGrouping<PlatformId, PlatformInfo> platformGrouping)
         {
-            IEnumerable<string> platformNameParts = new string[] {
```
MichaelSimons: This is some general cleanup.  In practice the plaformNameParts are already part of the Dockerfile path so this logic wasn't changing the resulting matrix leg names.

Issue #59, opened 2017-11-07
Add support for only building Dockerfiles that have changed.


MichaelSimons:
Currently the official builds always build the entire repo when triggered.  This is inefficient and hurts the ability to have shared images/layers in production.  Tooling should be created that will detect which Dockerfiles should get rebuilt whenever Dockerfiles are changed as well as base images.

PR #191, opened 2019-06-04
Support for image data file production

mthalman:
Updates the Image Builder tool to define commands that aid in the production of a data file that contains info about images produced during a build.  The end goal of all this is to be able to automatically queue a build whenever a base image changes and have that build only target the images that reference that changed base image (related to #59).

Change details:
* Updates the Build command to write to a file the external base images digest values of the images that were built by the command.  The Build command gets run by many jobs each producing a small subset of the whole graph so each one of those jobs will be producing such a data file.
* Defines a new command named UpdateImageData.  This command consumes all of the image data files that were produced by the Build command and merges them all into a master file.  This master file, named image-data.json, is stored in the docker-tools repo and is intended to contain image data for all .NET Docker images.  This file has been included with these changes and is bootstrapped with the data produced by the dotnet-docker nightly build.  The UpdateImageData command is intended to be executed in a pipeline during the publish stage; once all the images have been published, we want to track the data associated with these images in this master file.
* Defines a new command named RebuildStaleImages.  This command consumes a subscriptions.json file which describes all of the .NET Docker image repos we want to track and the associated pipelines that produce those images.  The command will check whether those images are up-to-date with their referenced base image by comparing the digest of the currently available base image with the digest value stored in the master image data file stored in docker-tools.  If the digests don't match, the command will queue a build to rebuild the image that references that base image.  A pipeline is also included with these changes that invokes this command.  This pipeline is intended to be run on a scheduled basis to ensure the images are kept up-to-date.


Microsoft.DotNet.ImageBuilder/src/SubscriptionsModel/PipelineTrigger.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.DotNet.ImageBuilder.SubscriptionsModel
+{
+    public class PipelineTrigger
+    {
+        [JsonProperty(Required = Required.Always)]
+        public int Id { get; set; }
+
+        [JsonProperty(Required = Required.Always)]
+        public string PathArgsName { get; set; }
```
MichaelSimons: >PathArgsName [](start = 22, length = 12)

What do you think of renaming this in terms of the pipeline concept - `PathVariable`

Microsoft.DotNet.ImageBuilder/src/ManifestModel/Architecture.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-namespace Microsoft.DotNet.ImageBuilder.Model
+namespace Microsoft.DotNet.ImageBuilder.ManifestModel
```
MichaelSimons: >ManifestModel [](start = 40, length = 13)

What do you think of a more hierarchical namespace for the models?  

 Microsoft.DotNet.ImageBuilder.Models.Manifest
 Microsoft.DotNet.ImageBuilder.Models.Subscription
etc

Microsoft.DotNet.ImageBuilder/src/SubscriptionsModel/GitRepoInfo.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.ImageBuilder.SubscriptionsModel
+{
+    public class GitRepoInfo
```
MichaelSimons: >GitRepoInfo [](start = 17, length = 11)

I think we should avoid using the `Info` suffix within the Model Types.

Microsoft.DotNet.ImageBuilder/src/JsonHelper.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    internal static class JsonHelper
```
MichaelSimons: >internal [](start = 4, length = 8)

I have been following a different convention where all type are public.  I don't have a strong preference on this but I do feel strongly about being consistent.  If you would like to change the convention, I'd suggest logging an issue to convert the entire code base.

Microsoft.DotNet.ImageBuilder/src/DictionaryExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    internal static class DictionaryExtensions
+    {
+        public static Dictionary<TKey, TValue> Sort<TKey, TValue>(this IDictionary<TKey, TValue> dict)
```
MichaelSimons: >Dictionary [](start = 22, length = 10)

Would it be possible to use a SortedDictionary in the first place instead of re-allocating and sorting the dictionary?

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateVersionsOptions.cs
```diff
@@ -79,4 +79,4 @@ public override void ParseCommandLine(ArgumentSyntax syntax)
             GitAuthToken = gitAuthToken;
         }
     }
-}
+}
```
MichaelSimons: Lost EOF line ending

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateVersionsCommand.cs
```diff
@@ -99,4 +99,4 @@ private async Task<GitObject[]> GetUpdatedVerionInfo(GitHubClient client, GitHub
             return versionInfo.ToArray();
         }
     }
-}
+}
```
MichaelSimons: Lost EOF Line Ending

Microsoft.DotNet.ImageBuilder/src/Commands/ManifestCommand.cs
```diff
@@ -0,0 +1,30 @@
...
+
+            Manifest = ManifestInfo.Create(
+                Options.Manifest,
+                Options.GetManifestFilter(),
+                Options);
+
+            if (Options.IsVerbose)
+            {
+                Logger.WriteMessage(JsonConvert.SerializeObject(Manifest, Formatting.Indented));
+            }
```
MichaelSimons: I think this code path should be executed anytime the Manifest is loaded (e.g. RebuildStaleImagesCommand should be using this as well).  Maybe just move the LoadManifest as it was previously defined (taking in the context path) as a new method on ManifestInfo and make the existing Create method private.  The benefit is it separates the view model creation logic from the Model loading and validation.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -85,8 +85,14 @@ private static void AddImageBuilderPathsVariable(string[] dockerfilePaths, LegIn
...
 
-        private static void AddPlatformVariables(IGrouping<PlatformId, PlatformInfo> platformGrouping, LegInfo leg)
+        private static void AddCommonVariables(IGrouping<PlatformId, PlatformInfo> platformGrouping, LegInfo leg)
         {
+            string fullyQualifiedLegName =
+                (platformGrouping.Key.OsVersion ?? platformGrouping.Key.OS.GetDockerName()) +
+                platformGrouping.Key.Architecture.GetDisplayName(platformGrouping.Key.Variant) +
+                leg.Name;
+
+            leg.Variables.Add(("legName", fullyQualifiedLegName));
```
MichaelSimons: >legName [](start = 32, length = 7)

Can you explain how this is being used?

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -85,8 +85,14 @@ private static void AddImageBuilderPathsVariable(string[] dockerfilePaths, LegIn
             leg.Variables.Add(("imageBuilderPaths", pathArgs));
         }
 
-        private static void AddPlatformVariables(IGrouping<PlatformId, PlatformInfo> platformGrouping, LegInfo leg)
+        private static void AddCommonVariables(IGrouping<PlatformId, PlatformInfo> platformGrouping, LegInfo leg)
         {
+            string fullyQualifiedLegName =
+                (platformGrouping.Key.OsVersion ?? platformGrouping.Key.OS.GetDockerName()) +
```
MichaelSimons: >(platformGrouping.Key.OsVersion ?? platformGrouping.Key.OS.GetDockerName())  [](start = 16, length = 76)

There are going to be issues with this with my latest changes.  OsVersion is now required so GetDockerName will never get invoked.  OsVersion may not be what you want for linux legs.

image-data.json
```diff
@@ -0,0 +1,474 @@
+[
```
MichaelSimons: This file is being moved right?

Microsoft.DotNet.ImageBuilder/src/ImageModel/ImageData.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.ImageBuilder.ImageModel
```
MichaelSimons: >ImageModel [](start = 40, length = 10)

What do you think of using the nomenclature for this concept that the official Docker repos use - either TagDetails or RepoInfo

Microsoft.DotNet.ImageBuilder/src/ImageModel/ImageData.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public class ImageData
+    {
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public Dictionary<string, string> BaseImageDigests { get; set; }
```
MichaelSimons: >BaseImageDigests [](start = 42, length = 16)

Thoughts on dropping the `Digests` to keep the name shorter?

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataOptions.cs
```diff
@@ -0,0 +1,83 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.CommandLine;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class UpdateImageDataOptions : Options
```
MichaelSimons: >UpdateImageDataOptions [](start = 17, length = 22)

How about `PublishImageData`

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataOptions.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        protected override string CommandHelp => "Updates an image data file in GitHub by merging in the image data that was generated from a build.";
```
MichaelSimons: >CommandHelp [](start = 34, length = 11)

This is intended to be terse - `Publishes the merged build image data`

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
mthalman: It provides a unique value to identify the build job so that we can produce a pipeline artifact by that name and not have it conflict with any of the other jobs.

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataOptions.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        public string GitUsername { get; set; }
+        public string GitEmail { get; set; }
+        public string GitAuthToken { get; set; }
+        public string GitOwner { get; set; }
+        public string GitRepo { get; set; }
+        public string GitBranch { get; set; }
+        public string GitImageDataPath { get; set; }
```
MichaelSimons: I think these should be factored into a shared type following the patter used by ManifestFilterOptions.  These Options are being used with the PublishMcrDocsOptions type.

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataCommand.cs
```diff
@@ -0,0 +1,134 @@
...
+                string srcImageDataContents = File.ReadAllText(sourceImageDataPath);
+                RepoData[] srcRepos = JsonConvert.DeserializeObject<RepoData[]>(srcImageDataContents);
+                srcReposList.Add(srcRepos);
+            }
+
+            GitHubAuth githubAuth = new GitHubAuth(Options.GitAuthToken, Options.GitUsername, Options.GitEmail);
+            using (GitHubClient client = new GitHubClient(githubAuth))
+            {
+                GitHubProject project = new GitHubProject(Options.GitRepo, Options.GitOwner);
+                GitHubBranch branch = new GitHubBranch(Options.GitBranch, project);
```
MichaelSimons: It feels like there is an opportunity to share logic with the other cmds that are doing the same thing.  This code doesn't retry the git commit which I think it should.

```
                foreach (RepoData[] srcRepos in srcReposList)
                {
                    MergeRepos(srcRepos, targetRepos);
                }

                targetRepos = RepoData.SortRepoData(targetRepos);

                string newTargetImageDataContents = JsonHelper.SerializeObject(targetRepos.ToArray()) + Environment.NewLine;

                if (originalTargetImageDataContents != newTargetImageDataContents)
                {
                    GitObject imageDataGitObject = new GitObject
                    {
                        Path = Options.GitImageDataPath,
                        Type = GitObject.TypeBlob,
                        Mode = GitObject.ModeFile,
                        Content = newTargetImageDataContents
                    };
               }

               await GitHelper.PatchAsync(gitOptions, imageDataGitObject);
```

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataCommand.cs
```diff
@@ -0,0 +1,134 @@
...
+        private void MergeImages(RepoData srcRepo, RepoData targetRepo)
+        {
+            if (srcRepo.Images == null)
+            {
+                return;
+            }
+
+            if (srcRepo.Images.Any() && targetRepo.Images == null)
+            {
+                targetRepo.Images = new Dictionary<string, ImageData>();
```
MichaelSimons: Would the following work and be more efficient?

targetRepo.Images = srcRepo.Images;
return;

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataCommand.cs
```diff
@@ -0,0 +1,134 @@
...
+        private void MergeDigests(ImageData srcImage, ImageData targetImage)
+        {
+            if (srcImage.BaseImageDigests == null)
+            {
+                return;
+            }
+
+            if (srcImage.BaseImageDigests.Any() && targetImage == null)
+            {
+                targetImage.BaseImageDigests = new Dictionary<string, string>();
```
MichaelSimons: Same comment as with MergeImages

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataOptions.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            string gitImageDataPath = null;
+            syntax.DefineOption(
+                "git-image-data-path",
+                ref gitImageDataPath,
+                "Path to the image data file in GitHub.");
+            GitImageDataPath = gitImageDataPath;
+
+            string sourceImageDataFolderPath = null;
+            syntax.DefineOption(
```
MichaelSimons: It feels like these should be parameters not options.

Microsoft.DotNet.ImageBuilder/src/ImageModel/ImageData.cs
mthalman: Why does the name need to be shorter?

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+                ref buildOrganization,
+                "The name of the Azure DevOps organization where builds are queued.");
+            BuildOrganization = buildOrganization;
+
+            string buildProject = null;
+            syntax.DefineOption(
+                "build-project",
+                ref buildProject,
+                "The name of the Azure DevOps project where builds are queued.");
+            BuildProject = buildProject;
```
MichaelSimons: I was expecting these to be defined in the subscriptions file.

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+            const string DefaultImageDataPath = "image-data.json";
+            string imageDataPath = DefaultImageDataPath;
+            syntax.DefineOption(
+                "image-data-path",
+                ref imageDataPath,
+                $"Path to the file containing image metadata (defaults to '{DefaultImageDataPath}').");
+            ImageDataPath = imageDataPath;
+
+            string buildPersonalAccessToken = null;
+            syntax.DefineOption(
```
MichaelSimons: >DefineOption [](start = 19, length = 12)

Should this be a parameter.

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesOptions.cs
mthalman: We've only got a single PAT that gets passed in so we can't target arbitrary DevOps projects.  The PAT that gets passed in is tied to a specific organization and project so that's why they're passed in together.

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+            IEnumerable<string> pathsToRebuild = await GetPathsToRebuildAsync(subscription, repos);
+
+            string pathsStr = String.Join(", ", pathsToRebuild.ToArray());
+            Logger.WriteMessage(
+                $"The following images in subscription '{subscription.ToString()}' were determined to be using out-of-date base images: {pathsStr}");
+
+            await ExecuteBuild(subscription, pathsToRebuild);
+        }
+
+        private async Task ExecuteBuild(Subscription subscription, IEnumerable<string> pathsToRebuild)
```
MichaelSimons: >ExecuteBuild [](start = 27, length = 12)

What do you think about calling this QueueBuild?

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+        {
+            string formattedParameters = String.Join(" ", pathsToRebuild
+                .Select(path => $"--path '{path}'")
```
MichaelSimons: >path [](start = 36, length = 4)

What do you think about utilizing a shared const with the ManifestOptions?

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+                .ToArray());
```
MichaelSimons: >ToArray [](start = 17, length = 7)

Consider using LINQ throughout and avoid the ToArray call by using Aggregate.

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            string parameters = "{\"" + subscription.PipelineTrigger.PathArgsName + "\": \"" + formattedParameters + "\"}";
+
+            using (VssConnection connection = new VssConnection(
+                new Uri($"https://dev.azure.com/{Options.BuildOrganization}"),
+                new VssBasicCredential(String.Empty, Options.BuildPersonalAccessToken)))
+            {
+                using (ProjectHttpClient projectHttpClient = connection.GetClient<ProjectHttpClient>())
```
MichaelSimons: Suggestion:  use the multiple `using` syntax to avoid as much nesting.

```
            using (HttpClient client = new HttpClient())
            using (HttpResponseMessage response = await client.GetAsync(downloadUrl))
            {
```

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        private async Task ExecuteBuild(Subscription subscription, IEnumerable<string> pathsToRebuild)
+        {
+            string formattedParameters = String.Join(" ", pathsToRebuild
+                .Select(path => $"--path '{path}'")
+                .ToArray());
+
+            string parameters = "{\"" + subscription.PipelineTrigger.PathArgsName + "\": \"" + formattedParameters + "\"}";
+
+            using (VssConnection connection = new VssConnection(
```
MichaelSimons: This is a place that would benefit from supporting the IsDryRun option.  This option has proven to be resourceful IMO as a way to debug and test the commands.  To get the most of this, the code should be instrumented with more logging.  For example here it should emit a message indicating which build is being queued and the args it is being queued with.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
mthalman: OsVersion is required in the model but it doesn't get set on PlatformId if the OS is Linux.  So it is null in that case.

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+                        }
+
+                        if (hasDigestChanged)
+                        {
+                            pathsToRebuild.Add(platform.BuildContextPath);
+                        }
+                    }
+                    else
+                    {
+                        pathsToRebuild.Add(platform.BuildContextPath);
```
MichaelSimons: Isn't this an unexpected edge case?  Should this be treated as an exception?  If not, it feels like this is a case that should have message logged so that we can ensure it is not happening often.

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+                foreach (var platform in platforms)
+                {
+                    if (repoData != null &&
+                        repoData.Images != null &&
+                        repoData.Images.TryGetValue(platform.BuildContextPath, out ImageData imageData))
+                    {
+                        bool hasDigestChanged = false;
+                        foreach (string fromImage in platform.ExternalFromImages)
+                        {
+                            DockerHelper.PullImage(fromImage, Options.IsDryRun);
```
MichaelSimons: >PullImage [](start = 41, length = 9)

I think it is time we investigate a more performant alternative which utilizes the Docker registry API.  Can you please log an issue for this?

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+                    {
+                        pathsToRebuild.Add(platform.BuildContextPath);
+                    }
+                }
+            }
+
+            return pathsToRebuild;
+        }
+
+        private async Task<string> GetGitRepoPath(Subscription sub)
```
MichaelSimons: >sub [](start = 63, length = 3)

Consider spelling out.  Variable names should be descriptive regardless of a methods access level.

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataCommand.cs
```diff
@@ -0,0 +1,134 @@
...
+        public override async Task ExecuteAsync()
+        {
+            List<RepoData[]> srcReposList = new List<RepoData[]>();
+            
+            foreach (string sourceImageDataPath in Directory.EnumerateFiles(Options.SourceImageDataFolderPath, "*.json", SearchOption.AllDirectories))
+            {
+                string srcImageDataContents = File.ReadAllText(sourceImageDataPath);
+                RepoData[] srcRepos = JsonConvert.DeserializeObject<RepoData[]>(srcImageDataContents);
+                srcReposList.Add(srcRepos);
+            }
```
MichaelSimons: Consider utilizing LINQ here

IEnumerable<RepoData[]> srcReposList = Directory.EnumerateFiles(Options.SourceImageDataFolderPath, "*.json", SearchOption.AllDirectories)
  .Select(imageData =>  JsonConvert.DeserializeObject<RepoData[]>(File.ReadAllText(sourceImageDataPath)))

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -43,54 +44,104 @@ public override Task ExecuteAsync()
...
+            {
+                string digest = DockerHelper.GetImageDigest(fromImage, Options.IsDryRun);
+                baseImageDigestMappings[fromImage] = digest;
+            }
+
+            Dictionary<string, string> digests = new Dictionary<string, string>();
+            foreach (KeyValuePair<string, string> baseImageDigestMapping in baseImageDigestMappings)
+            {
+                digests.Add(baseImageDigestMapping.Key, baseImageDigestMapping.Value);
+            }
```
MichaelSimons: I am confused by this?  What is using digests?

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -43,54 +44,104 @@ public override Task ExecuteAsync()
...
             }
 
             BuiltTags = BuiltTags.ToArray();
+
+            reposList = RepoData.SortRepoData(reposList);
+            string digestsString = JsonHelper.SerializeObject(reposList.ToArray());
+            File.WriteAllText(Options.ImageDataOutputPath, digestsString);
+        }
+
+        private Dictionary<string, string> GetBaseImageDigests(PlatformInfo platform)
```
MichaelSimons: >GetBaseImageDigests [](start = 43, length = 19)

Consider using Linq

```
private Dictionary<string, string> GetBaseImageDigests(PlatformInfo platform) => platform.ExternalFromImages.ToDictionary(image => image, DockerHelper.GetImageDigest(image, Options.IsDryRun);
```

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
MichaelSimons: Have you considered using utilizing one of the built-in variable names instead (e.g - Agent.JobName)?  The reason I mention it is to limit the dependencies the image-data has on the rest of the system.

---
In reply to: [290462260](https://github.com/dotnet/docker-tools/pull/191#discussion_r290462260) [](ancestors = 290462260)

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
MichaelSimons: Oh sure - that's a duh on me :)

---
In reply to: [290488520](https://github.com/dotnet/docker-tools/pull/191#discussion_r290488520) [](ancestors = 290488520)

Microsoft.DotNet.ImageBuilder/src/ImageModel/ImageData.cs
MichaelSimons: To me the property contains more than just digests.  It is general info about the base images.

Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
mthalman: I wish we could use that but the Agent.JobName variable does not get set to the "expanded" name of the job as a result of applying the matrix.  For example, Agent.JobName will be set to "Build_NanoServer1803_amd64" for both of these jobs: "Build_NanoServer1803_amd64 2.2nanoserver-1803" and "Build_NanoServer1803_amd64 3.0nanoserver-1803".  I see this additional variable as being a general purpose mechanism that can be useful whenever we need a unique value for the job; it's not exclusively tied to this feature.

---
In reply to: [290508154](https://github.com/dotnet/docker-tools/pull/191#discussion_r290508154) [](ancestors = 290508154,290462260)

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
mthalman: Yes, it wouldn't be expected often.  The scenario would be:
1. We add a new Dockerfile path to manifest and commit that.
2. Scheduled build that runs this command gets executed.
3. Sees that no image data for that Dockerfile path has been written yet.
4. Queues a build to get it updated.

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
mthalman: https://github.com/dotnet/docker-tools/issues/192

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
mthalman: I originally had it written as 'subscription' but the usage of it within the method caused the lines to get so long because it was referenced so much.  I felt it was a pretty benign change to make considering its type is Subscription and all.  There's no question as to what it is.

---
In reply to: [290493915](https://github.com/dotnet/docker-tools/pull/191#discussion_r290493915) [](ancestors = 290493915)

Microsoft.DotNet.ImageBuilder/src/ImageModel/ImageData.cs
mthalman: I guess I prefer Image over Repo and Tag.  The whole point of the data that we're collecting is about the image.  The Repo is just an organizing aspect.  Tag is just one part of it and a piece of data we're not even collecting right now.  

---
In reply to: [290460329](https://github.com/dotnet/docker-tools/pull/191#discussion_r290460329) [](ancestors = 290460329)

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
mthalman: Disregard.  That's dead code.

---
In reply to: [290498370](https://github.com/dotnet/docker-tools/pull/191#discussion_r290498370) [](ancestors = 290498370)

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
mthalman: The problem with your code snippet is that the lambda passed to ToDictionary is the key selector and it's returning the image digest as the key.  We want image name to be the key, not the digest.  Leaving the code as is.

---
In reply to: [290499655](https://github.com/dotnet/docker-tools/pull/191#discussion_r290499655) [](ancestors = 290499655)

Microsoft.DotNet.ImageBuilder/src/SubscriptionsModel/PipelineTrigger.cs
mthalman: Agreed

Microsoft.DotNet.ImageBuilder/src/ManifestModel/Architecture.cs
mthalman: Sounds good

Microsoft.DotNet.ImageBuilder/src/DictionaryExtensions.cs
mthalman: Yes, that works quite nicely.

Microsoft.DotNet.ImageBuilder/src/Commands/ManifestCommand.cs
mthalman: I've refactored it to be in ManifestInfo.

image-data.json
mthalman: Yes, removed.

Microsoft.DotNet.ImageBuilder/src/ImageModel/ImageData.cs
mthalman: Naming the concept of the file as "image info".

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataCommand.cs
mthalman: Refactored to have a common method in GitHelper.

Microsoft.DotNet.ImageBuilder/src/Commands/UpdateImageDataCommand.cs
mthalman: Yep

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
mthalman: Moved as constant in ManifestFilterOptions.

Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
MichaelSimons: That is what I thought.  It seems like a cause that would be worthwhile to log.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
MichaelSimons: I typed the syntax wrong.  `image` here is the name of the from image...but since you are using SortedDictioinary now, this isn't as useful

```
private Dictionary<string, string> GetBaseImageDigests(PlatformInfo platform) => platform.ExternalFromImages.ToDictionary(image => image, image => DockerHelper.GetImageDigest(image, Options.IsDryRun);
```

---
In reply to: [291184185](https://github.com/dotnet/docker-tools/pull/191#discussion_r291184185) [](ancestors = 291184185,290499655)

Microsoft.DotNet.ImageBuilder/src/Commands/BuildOptions.cs
```diff
@@ -36,6 +37,10 @@ public override void ParseCommandLine(ArgumentSyntax syntax)
             bool isSkipPullingEnabled = false;
             syntax.DefineOption("skip-pulling", ref isSkipPullingEnabled, "Skip explicitly pulling the base images of the Dockerfiles");
             IsSkipPullingEnabled = isSkipPullingEnabled;
+
+            string imageInfoOutputPath = null;
+            syntax.DefineOption("image-info-output-path", ref imageInfoOutputPath, "Path to write the image info information to.");
```
MichaelSimons: >Path to write the image info information to. [](start = 84, length = 44)

(Nit) None of the option descriptions are complete sentences.  They are terse to save space.  At a minimum the punctuation should be removed.

Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.CommandLine;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class GitOptions
+    {
+        private readonly string defaultOwner;
+        private readonly string defaultRepo;
+        private readonly string defaultBranch;
+        private readonly string defaultPath;
```
MichaelSimons: It seems like you could easily get away without declaring these fields and just use the non-default fields to store the default value.  What is the value in declaring this extra state?

Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
```diff
@@ -34,43 +32,28 @@ public override async Task ExecuteAsync()
...
+                    GitTree tree = await client.PostTreeAsync(project, masterSha, gitObjects);
+                    string commitMessage = $"Mirroring {productRepo} readmes";
+                    GitCommit commit = await client.PostCommitAsync(
+                        project, commitMessage, tree.Sha, new[] { masterSha });
+
+                    // Only fast-forward. Don't overwrite other changes: throw exception instead.
+                    await client.PatchReferenceAsync(project, masterRef, commit.Sha, force: false);
                 }
-            }
+            });
```
MichaelSimons: It feels like there is more opportunity to factor out shared code here.

Caller await GitHelper.PushChangesAsync(Options.GitOptions, "Update Docker image digests", GetUpdatedVerionInfo);

public static async Task GitHelper.PushChangesAsync(
    GitOptions gitOptions,
    string commitMessage,
    Func<GitHubClient, GitHubBranch, Task<IEnumerable<GitObject>>> GetChanges,
    int maxTries = DefaultMaxTries,
    int retryMillisecondsDelay = DefaultRetryMillisecondsDelay)
{
    GitHubAuth githubAuth = new GitHubAuth(gitOptions.AuthToken, gitOptions.Username, gitOptions.Email);
    using (GitHubClient client = new GitHubClient(githubAuth))
    {
        for (int i = 0; i < maxTries; i++)
        {
            try
            {
                GitHubProject project = new GitHubProject(Options.GitOptions.Repo, Options.GitOptions.Owner);
                GitHubBranch branch = new GitHubBranch(Options.GitOptions.Branch, project);
                IEnumerable<GitObject> gitObjects = await GetChanges(client, branch);

                if (gitObjects.Any())
                {
                    string masterRef = $"heads/{Options.GitOptions.Branch}";
                    GitReference currentMaster = await client.GetReferenceAsync(project, masterRef);
                    string masterSha = currentMaster.Object.Sha;
                    GitTree tree = await client.PostTreeAsync(project, masterSha, gitObjects);
                    GitCommit commit = await client.PostCommitAsync(
                        project, commitMessage, tree.Sha, new[] { masterSha });

                    // Only fast-forward. Don't overwrite other changes: throw exception instead.
                    await client.PatchReferenceAsync(project, masterRef, commit.Sha, force: false);
                }
                break;
            }
            catch (HttpRequestException ex) when (i < (maxTries - 1))
            {
                Logger.WriteMessage($"Encountered exception interacting with GitHub: {ex.Message}");
                Logger.WriteMessage($"Trying again in {retryMillisecondsDelay}ms. {maxTries - i - 1} tries left.");
                await Task.Delay(retryMillisecondsDelay);
            }
        }
    }
}

MichaelSimons:
Looks good - I just had a few comments.

Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
mthalman: Done.  I've kept ExecuteGitOperationsWithRetryAsync as it's own method as I want the retry logic to be separate from the underlying operations.

Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
```diff
@@ -34,43 +32,14 @@ public override async Task ExecuteAsync()
 
             string productRepo = GetProductRepo();
 
-            GitHubAuth githubAuth = new GitHubAuth(Options.GitAuthToken, Options.GitUsername, Options.GitEmail);
-            using (GitHubClient client = new GitHubClient(githubAuth))
+            await GitHelper.ExecuteGitOperationsWithRetryAsync(Options.GitOptions, async client =>
             {
-                for (int i = 0; i < MaxTries; i++)
+                await GitHelper.PushChangesAsync(client, Options.GitOptions, $"Mirroring {productRepo} readmes", async branch =>
```
MichaelSimons: I like the sharing of code here, nice!

Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.CommandLine;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class GitOptions
+    {
+        private string owner;
```
MichaelSimons: This doesn't solve my original concern.  This options class is still caring around/allocating extra state unnecessarily.

My suggestion would be the following.
1. Remove the private fields.
2. Set the public property to the default value within the ctor.
   `Owner = defaultOwner ?? throw new ArgumentNullException(nameof(defaultOwner));`
3. Utilize the default value within ParseCommandLine
```
            string owner = Owner;
            syntax.DefineOption(
                "git-owner",
                ref owner,
                $"Owner of the GitHub repo to write to (defaults to {owner})");
            Owner = owner;
```

mthalman: Ah, I get what you're saying now.

PR #194, opened 2019-06-10
Add pipeline to trigger builds for updating base images

mthalman:
Defines a pipeline that invokes the rebuildStaleImages command of Image Builder.  This command will read the subscriptions.json file included in these changes to determine the images that are to be checked to ensure their base images are up-to-date with the base images defined in the image info file from the versions repo.  If not, Image Builder will kick off the pipeline defined in the subscriptions file so that the images get updated.


mthalman:
The most significant files here are:
* .vsts-pipelines/check-base-image-updates.yml
* .vsts-pipelines/steps/queue-build-for-stale-images.yml
* subscriptions.json

The rest are just infrastructure files copied verbatim from dotnet-docker.

.vsts-pipelines/check-base-image-updates.yml
```diff
@@ -0,0 +1,24 @@
...
+  # Use the most recent Windows version so we can pull all image versions of Windows
+  pool: # windows1903Amd64
+    name: DotNetCore-Docker
+    demands: VSTS_OS -equals Windows_Server_2019_Data_Center_1903
+  # Sequentially execute the jobs to avoid race condition and having multiple builds getting queued
+  dependsOn: Build_Linux
+  steps:
+  - template: steps/queue-build-for-stale-images.yml
+    parameters:
+      osType: windows
```
MichaelSimons: (Nit) There are a few files that are missing the EOF line ending.

.vsts-pipelines/steps/queue-build-for-stale-images.yml
```diff
@@ -0,0 +1,21 @@
+parameters:
+  osType: null
+  imageInfoRepoPath: 'build-info/docker/image-info.json'
+steps:
+  - script: |
+      git clone -n https://github.com/dotnet/versions.git --depth 1
```
MichaelSimons: This seems heavy handed.  Why not just download the single file via its raw url (e.g. https://raw.githubusercontent.com/dotnet/versions/master/build-info/docker/image-info.json)?

.vsts-pipelines/steps/cleanup-docker-windows.yml
```diff
@@ -0,0 +1,15 @@
+steps:
```
MichaelSimons: Have you considered to putting these common build pipeline artifacts into a common directory now in preparation for the upcoming shared common infrastructure work?  The only reason I mention it is that it would eliminate the need to update the references and then test the changes.

.vsts-pipelines/check-base-image-updates.yml
```diff
@@ -0,0 +1,24 @@
...
+  steps:
+  - template: steps/queue-build-for-stale-images.yml
+    parameters:
+      osType: linux
+- job: Build_Windows
+  # Use the most recent Windows version so we can pull all image versions of Windows
+  pool: # windows1903Amd64
+    name: DotNetCore-Docker
+    demands: VSTS_OS -equals Windows_Server_2019_Data_Center_1903
+  # Sequentially execute the jobs to avoid race condition and having multiple builds getting queued
```
MichaelSimons: How does executing this sequentially prevent multiple builds from getting queued?

mthalman: Image Builder will ensure that behavior by not queuing a second build if one is already running.  But if the jobs run in parallel, there's always the change that Image Builder gets invoked at the same time by each of the jobs in such a way that it thinks no build is running for each of them and it queues two builds.  By running the jobs sequentially, it prevents the chance of Image Builder getting invoked in that way.

.vsts-pipelines/steps/cleanup-docker-windows.yml
mthalman: The references are going to need to change anyway since they'll be in a directory above where .vsts-pipelines is currently located.  Plus, there's enough changes going on to hook up to the new eng folder structure that testing will be a requirement anyway.  So I'll just leave it as is.

.vsts-pipelines/steps/queue-build-for-stale-images.yml
mthalman: Updated to use curl.

PR #195, opened 2019-06-11
Enable manifest filtering for RebuildStaleImages command

mthalman:
The RebuildStaleImages command currently enumerates all images in the manifest and attempts to pull them to get their base image digest value.  This can't be done when there is a mix of Linux and Windows images because you can't pull a Windows image from a Linux Docker host, for example.  For that reason, there needs to be filtering options for the command that are applied to the manifests that are consumed.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
@@ -53,7 +53,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
             if (!pathsToRebuild.Any())
             {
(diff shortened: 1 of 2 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/RebuildStaleImagesCommand.cs
```diff
@@ -51,6 +51,12 @@ private async Task QueueBuildForStaleImages(Subscription subscription, RepoData[
         {
             IEnumerable<string> pathsToRebuild = await GetPathsToRebuildAsync(subscription, repos);
 
+            if (!pathsToRebuild.Any())
+            {
+                Logger.WriteMessage($"All images for subscription {subscription} are using up-to-date base images. No rebuild necessary.");
```
MichaelSimons:
```suggestion
                Logger.WriteMessage($"All images for subscription '{subscription}' are using up-to-date base images. No rebuild necessary.");
```

PR #198, opened 2019-06-14
Add build pipeline to cleanup old ACR images

### LINKED ISSUE #150: Implement cleanup logic for ACR build staging repositories
Each build in the Docker repositories uses the following pattern.

- Build Phase - build images and push to ACR staging location
- Test Phase - pull images from staging location and run tests
- Publish Phase - published images from staging location to MCR/DH

We don't have logic that cleans up the ACR staging location.  This location filled up recently and I ran a local script to cleanup old repos.  We should have automated tooling that cleans this up.

I don't think this is as simple as adding a cleanup phase to the build.  I think we may want to keep the images around for a time period in case there are issues downstream.  We also have scenarios where we want to support publishing to internal MCR and then release at a later time.  Being able to rerun the publish build phase to publish a specific build to public MCR would be desirable in this scenario.

mthalman:
Fixes #150

### DIFF
--- a/scripts/Invoke-AcrCleanup.ps1
+++ b/scripts/Invoke-AcrCleanup.ps1
@@ -61,3 +61,4 @@ foreach ($deletedRepo in $deletedRepos) {
 
 Write-Host ""
 Write-Host "Total repositories deleted: $($deletedRepos.Count)"
+Write-Host "Total repositories remaining: $($repos.Count - $deletedRepos.Count)"
--- a/scripts/Invoke-AcrCleanup.ps1
+++ b/scripts/Invoke-AcrCleanup.ps1
@@ -1,0 +1,68 @@
+[cmdletbinding(DefaultParameterSetName=$null)]
+param(
+    [Parameter(Mandatory = $true)]
+    [string]
+    $RegistryName,
+
+    [Parameter(Mandatory = $true)]
+    [string]
+    $SubscriptionId,
+
+    [switch]
+    $WhatIf,
+
+    [string]
+    $ServicePrincipalName,
+
+    [string]
+    $ServicePrincipalPassword,
+
+    [string]
+    $ServicePrincipalTenant
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+$suffixIndex = $RegistryName.IndexOf(".azurecr.io")
+if ($suffixIndex -ge 0) {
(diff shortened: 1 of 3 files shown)


scripts/Invoke-AcrCleanup.ps1
```diff
@@ -0,0 +1,61 @@
+[cmdletbinding(DefaultParameterSetName=$null)]
+param(
```
MichaelSimons: We are using a different style in our other PS scripts.  Thoughts on using a more compact consistent style for the parameters?

scripts/Invoke-AcrCleanup.ps1
```diff
@@ -0,0 +1,61 @@
...
+Write-Host ""
+$repos = az acr repository list --name $RegistryName | ConvertFrom-Json
+
+foreach ($repoName in $repos) {
+    Write-Host "Querying repository '$repoName'"
+    $repo = az acr repository show --name $RegistryName --repository $repoName | ConvertFrom-Json
+    $lastUpdateTime = [datetime]$repo.lastUpdateTime
+
+    # If the repo was last updated more than X days ago, delete it
+    if ($lastUpdateTime.AddDays(45) -lt (Get-Date)) {
```
MichaelSimons: 1. I don't think we want to ever delete any of the `public` repo.  We should be safeguarding against this. 
2. 45 days seems like a long time.  What is the size of the repositories less than 45 days old?

mthalman: Why not delete from the public repo?  How would it still be used after this time period?

MichaelSimons: MCR has a feature on their backlog to support deletes.  According to Rohit, the feature has a requirement that you have the image stored in your ACR to be able to delete it.

scripts/Invoke-AcrCleanup.ps1
```diff
@@ -43,26 +30,34 @@ if ($ServicePrincipalName) {
...
+Write-Host "SUMMARY"
+Write-Host "======="
+Write-Host "Deleted repositories:"
+
+foreach ($deletedRepo in $deletedRepos) {
+    Write-Host $deletedRepo
+}
+
+Write-Host ""
+Write-Host "Total repositories deleted: $($deletedRepos.Count)"
```
MichaelSimons: I could see it being useful to also emit the total repositories remaining.

PR #212, opened 2019-06-20
Add build schedule for ACR cleanup build

mthalman:
Set the build to run every night.

### DIFF
--- a/eng/pipelines/cleanup-acr-images.yml
+++ b/eng/pipelines/cleanup-acr-images.yml
@@ -4,6 +4,9 @@ pr: none
 schedules:
 - cron: "0 5 * * *"
   displayName: Nightly build
+  branches:
+    include:
+    - master
 
 variables:
 - group: DotNet-Docker-Common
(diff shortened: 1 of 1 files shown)


eng/pipelines/cleanup-acr-images.yml
```diff
@@ -1,6 +1,10 @@
 trigger: none
 pr: none
 
+schedules:
+- cron: "0 5 * * *"
```
MichaelSimons:
```suggestion
- cron: "0 0 * * *"
```

Consider pushing this into the night.  I'd be willing to bet the workload at 5 is pretty high.

mthalman: This is what I was unsure about.  I was assuming this time was UTC which means it would be midnight in Central time.

MichaelSimons: You are right it is UTC time - https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?tabs=yaml&view=azure-devops#scheduled-triggers

> The time zone for cron schedules is UTC, so in these examples, the midnight build and the noon build are at midnight and noon in UTC.

PR #214, opened 2019-06-20
Add more OS version display names

mthalman:
As part of the integration of the common engineering infrastructure into the dotnet-buildtools-prereqs-docker repo, that repo will now have a dependency on the latest version of Image Builder.  This new version requires the osVersion field to be set in the manifest for all platforms.  That repo makes use of many OS versions that are not covered in the Image Builder code that translates the OS version into a display name.  These changes update the code to include all the cases needed by the dotnet-buildtools-prereqs-docker repo.


mthalman:
This isn't yet required and may lead to a redesign based on the circumstances around the dotnet-buildtools-prereqs-docker repo.  Closing this for now.

PR #216, opened 2019-06-21
only output ImageInfoOutputPath when appropriate

mthalman:


mthalman:
> LGTM
> 
> On a related note, should the build definitions be updated to exclude this option for public CI? It seems like wasted overhead.

Yes, I'll log an issue.

mthalman:
> > LGTM
> > On a related note, should the build definitions be updated to exclude this option for public CI? It seems like wasted overhead.
> 
> Yes, I'll log an issue.

Oh wait, I guess I can just update that here now that we have eng\common.

eng/common/templates/jobs/build-images.yml
```diff
@@ -50,14 +40,29 @@ jobs:
...
+        echo "##vso[task.setvariable variable=imageBuilderBuildArgs]
+        --image-info-output-path "$(imageBuilderArtifactsPath)/$(legName)-image-info.json"
+        --registry-override $(acr.server)
+        --repo-prefix $(stagingRepoPrefix)
+        --push
+        --username $(acr.userName)
+        --password $(BotAccount-dotnet-docker-acr-bot-password)
+        $(imageBuilder.queueArgs)"
+      displayName: Define imageBuilderBuildArgs Variable
+  - ${{ if eq(variables['System.TeamProject'], 'public') }}:
```
MichaelSimons: This 'if' could simply be removed and the common `$(imageBuilder.queueArgs)` could be added to the very end of the `Build Images` step.  The benefit is it gets all of the common args declared in one spot.

mthalman: The downside to that is that imageBuilderBuildArgs would be an undeclared variable in the public build.  That produces an ignored error in Windows I think (or maybe it's Linux).  I want to avoid that because I don't like seeing such errors in the build output.

MichaelSimons: Ok that makes sense.

PR #217, opened 2019-06-24
Fix RebuildStaleImages command to generate dependency graph

### LINKED ISSUE #201: RebuildStaleImages command needs to rebuild graph, not just one image
Currently the RebuildStaleImages command will check whether an image is referencing a base image that is stale and then queue a build to rebuild just that image.  But it actually needs to build that image's dependencies as well.  For example, if runtime-deps gets triggered to be rebuilt, the SDK image also needs to get rebuilt because it depends on runtime-deps.  All those dependencies need to be added as additional `--path` arguments.

mthalman:
Ensures that the set of paths produced by the RebuildStaleImages command consists of the image whose base image had changed as well as its dependent images.

Also added a little bit of code to ensure we don't process base images that had already been processed.

Fixes #201


.gitignore
```diff
@@ -7,5 +7,6 @@
 .vscode/
 .vs
 
-# Visual Studio debug profile
+# Visual Studio user-specific files
 **/launchSettings.json
+**/*.user
```
MichaelSimons: Missing line ending.

Issue #218, opened 2019-06-25
Allow testing of images without pulling dependencies from staging


mthalman:
There are scenarios where we want to build an image but not build the images that it's dependent on for testing purposes.  An example of this is mcr.microsoft.com/dotnet/core/sdk:3.0-buster.  Building that image will cause the testing stage to attempt to pull the runtime and ASP.NET images from staging.  But those images will not have been built and so will not exist in the staging location.  Instead, the images should be pulled from the public MCR location.

PR #219, opened 2019-06-25
Add logic to support testing with partial graph from build

mthalman:
These changes update Image Builder to provide support for the scenario described by #218.  The goal is for the test stage to be able to run tests for an image that was built by the build but the test requires images that were not built during the build.  An example of this is provided in #218.

To solve this, the image info files produced by the build are being utilized to track image tag information that can then be read by the test project to determine which images were actually produced by the build.  For those images, the test code will pull them from the staging location, where they were pushed during the build stage.  For any other images that the test requires, it will pull them from the public MCR location.

To make things cleaner, a new Image Builder command is defined which handles the merging of image info files (from each of the build jobs) into a single file.  This file can then be used by both the tests and the publish stage to do their work.  This command is called in a newly defined Post_Build stage that gets run after the Build stage.


Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -110,6 +110,10 @@ private void BuildImages()
                         {
                             imageData.BaseImages = baseImageDigests;
                         }
+
+                        List<string> tags = platform.Tags.Select(tag => tag.Name).ToList();
+                        tags.Sort();
```
MichaelSimons: Curioius why not use OrderBy as part of the Linq expression?

Microsoft.DotNet.ImageBuilder/src/Commands/MergeImageInfoFilesCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.DotNet.ImageBuilder.Models.Image;
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class MergeImageInfoFilesCommand : Command<MergeImageInfoFilesOptions>
```
MichaelSimons: Thoughts on removing `Files` from the name?  It doesn't feel necessary from a conceptual level about what the command does and keeps the name of the command shorter.

Microsoft.DotNet.ImageBuilder/src/Commands/MergeImageInfoFilesCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        public override Task ExecuteAsync()
+        {
+            List<RepoData[]> srcReposList = Directory.EnumerateFiles(Options.SourceImageInfoFolderPath, "*.json", SearchOption.AllDirectories)
```
MichaelSimons: Nit - line length

Microsoft.DotNet.ImageBuilder/src/Commands/MergeImageInfoFilesCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+                    $"No JSON files found in source folder '{Options.SourceImageInfoFolderPath}'");
+            }
+
+            List<RepoData> combinedRepos = new List<RepoData>();
+            foreach (RepoData[] repos in srcReposList)
+            {
+                ImageInfoHelper.MergeRepos(repos, combinedRepos);
+            }
+
+            string destinationContents = JsonHelper.SerializeObject(combinedRepos.OrderBy(r => r.Repo).ToArray()) + Environment.NewLine;
```
MichaelSimons: Nit - line length

Microsoft.DotNet.ImageBuilder/src/Commands/MergeImageInfoFilesOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public string DestinationImageInfoPath { get; set; }
+
+        public override void ParseCommandLine(ArgumentSyntax syntax)
+        {
+            base.ParseCommandLine(syntax);
+
+            string sourceImageInfoFolderPath = null;
+            syntax.DefineParameter(
+                "source-image-info-folder-path",
```
MichaelSimons: I realize this is the name already being used in `PublishImageInfoOptions` but it feels like a shorted name would be more consumer friendly.  How around just `source` and `destination` or `source-path` and `destination-path` ?

Microsoft.DotNet.ImageBuilder/src/Models/Image/ImageData.cs
```diff
@@ -11,5 +11,6 @@ public class ImageData
     {
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public SortedDictionary<string, string> BaseImages { get; set; }
+        public List<string> Tags { get; set; } = new List<string>();
```
MichaelSimons: I am wondering if this should be named more precisely.  These aren't the full set of tags as they don't include the shared tags.  What do you think about naming this `SimpleTags`?

Microsoft.DotNet.ImageBuilder/tests/ImageInfoHelperTests.cs
```diff
@@ -0,0 +1,188 @@
...
+            List<RepoData> targetRepos = new List<RepoData>();
+            ImageInfoHelper.MergeRepos(repoDataSet, targetRepos);
+
+            CompareRepos(repoDataSet, targetRepos);
+        }
+
+        [Fact]
+        public void ImageInfoHelper_MergeRepos_ExistingTarget()
+        {
+            RepoData[] repoDataSet = new RepoData[]
```
MichaelSimons: Would it make sense to share the image instances between the source repos and expected repos?

mthalman:
I just realized I forgot to include the pipeline changes.  Added them now.

eng/common/templates/variables/docker-images.yml
```diff
@@ -1,6 +1,6 @@
 variables:
-  imageNames.imageBuilder.linux: mcr.microsoft.com/dotnet-buildtools/image-builder:debian-20190620175920
-  imageNames.imageBuilder.windows: mcr.microsoft.com/dotnet-buildtools/image-builder:nanoserver-20190620175920
+  imageNames.imageBuilder.linux: mthalman/test:imagebuilder
+  imageNames.imageBuilder.windows: mcr.microsoft.com/dotnet-buildtools/image-builder:nanoserver-20190606195358
```
MichaelSimons: Because of this, I wonder if the build pipeline changes shouldn't be a separate PR?

eng/common/templates/stages/build-test-publish-repo.yml
```diff
@@ -162,12 +162,18 @@ stages:
       dockerClientOS: windows
       buildJobTimeout: ${{ parameters.windowsAmdBuildJobTimeout }}
 
+- stage: Post_Build
```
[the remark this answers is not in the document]
mthalman: To me, `Finalize`  would imply the very last stage that would be run (i.e. after publish) which is not the case here.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
mthalman: Just wasn't thinking.  Changed to OrderBy.

Microsoft.DotNet.ImageBuilder/src/Commands/MergeImageInfoFilesCommand.cs
mthalman: Makes sense.

eng/common/templates/jobs/build-images.yml
```diff
@@ -40,36 +50,19 @@ jobs:
...
-        $(imageBuilder.queueArgs)"
-      displayName: Define imageBuilderBuildArgs Variable
   - script: >
       $(runImageBuilderCmd) build
       --manifest $(manifest)
       $(imageBuilderPaths)
       --os-type $(osType)
       --os-version "$(osVersion)"
       --architecture $(architecture)
+      --image-info-output-path "$(artifactsPath)/$(legName)-image-info.json"
```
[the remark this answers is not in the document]
mthalman: Updated so it only gets set for internal builds.

eng/common/templates/stages/build-test-publish-repo.yml
```diff
@@ -162,12 +162,18 @@ stages:
...
+  dependsOn: Build
+  condition: and(succeeded(), or(eq(variables['singlePhase'], ''), eq(variables['singlePhase'], 'build')))
```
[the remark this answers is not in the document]
mthalman: Updated so it only gets included for internal builds.

eng/common/templates/variables/docker-images.yml
mthalman: Agreed.  Reverting changes to the pipeline files for this PR.

PR #221, opened 2019-07-02
Pipeline changes to support testing partial graph

mthalman:
These changes update the build pipeline to provide support for the scenario described by #218. The goal is for the test stage to be able to run tests for an image that was built by the build but the test requires images that were not built during the build. An example of this is provided in #218.

To solve this, the image info files produced by the build are being utilized to track image tag information that can then be read by the test project to determine which images were actually produced by the build. For those images, the test code will pull them from the staging location, where they were pushed during the build stage. For any other images that the test requires, it will pull them from the public MCR location.

A new Post_Build stage has been added which calls Image Builder to consolidate all of the image info files produced by the build jobs into a single image info file that can be used by both test and publish jobs.

### DIFF
--- a/eng/common/templates/jobs/post-build.yml
+++ b/eng/common/templates/jobs/post-build.yml
@@ -1,0 +1,17 @@
+jobs:
+- job: Build
+  pool: # linuxAmd64Pool
+    name: Hosted Ubuntu 1604
+  steps:
+  - template: ../steps/init-docker-linux.yml
+  - template: ../steps/download-build-artifact.yml
+    parameters:
+      targetPath: $(Build.ArtifactStagingDirectory)
+  - script: >
+      $(runImageBuilderCmd) mergeImageInfo
+      $(artifactsPath)
+      $(artifactsPath)/image-info.json
+    displayName: Merge Image Info Files
+  - publish: $(Build.ArtifactStagingDirectory)/image-info.json
+    artifact: image-info
+    displayName: Publish Image Info File Artifact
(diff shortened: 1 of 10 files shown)


eng/common/templates/jobs/post-build.yml
```diff
@@ -0,0 +1,17 @@
+jobs:
+- job: Build
```
MichaelSimons: Would PostBuild be a better name and help distinguish it from the build jobs that run before it?

eng/common/templates/steps/test-images-linux-client.yml
```diff
@@ -10,21 +16,16 @@ steps:
       optionalTestArgs="-DisableHttpVerification"
     fi
     if [ "${{ eq(variables['System.TeamProject'], 'public') }}" == "False" ]; then
-      optionalTestArgs="$optionalTestArgs -Registry $(acr.server) -RepoPrefix $(stagingRepoPrefix)"
+      optionalTestArgs="$optionalTestArgs -Registry $(acr.server) -RepoPrefix $(stagingRepoPrefix) -ImageInfoPath $(artifactsPath)/image-info/image-info.json"
```
MichaelSimons: At first I was a little torn on requiring this additional script arg across all Docker repos.  Some of the full fx repos (asp.net and wcf) may not use it.  I however like being explicit over other solutions that utilize env vars to control behavior.

eng/common/templates/jobs/post-build.yml
mthalman: I'm not sure that's necessary since the label of the job shows up as "Post_Build Build".  There isn't really any ambiguity there.

PR #222, opened 2019-07-05
Add Base Image Dependency Flow documentation

### LINKED ISSUE #207: Document base image change detection system
The system defined by https://github.com/dotnet/docker-tools/issues/59 should be documented.

mthalman:
Fixes #207


documentation/base-image-dependency-flow.md
```diff
@@ -0,0 +1,27 @@
...
+4. If any images need to be rebuilt, the pipeline constructs a special build argument that describes the set of paths to the Dockerfiles that need to be built.  It then queues a build for the pipeline referenced by the subscription in the subscriptions file, passing it this build argument.  When that pipeline builds the images, it will ensure that the built images are making use of the latest base images, just by the nature of the build process pulling the base image before building.  You can see here that there's a virtuous cycle; when the pipeline publishes these updated images, it updates the master image info so that it now references the latest base image digests.  That is, until the next base image changes and the process starts all over again.
+
+This entire process is hands-off automation running 7 days a week, allowing consumers of .NET Docker images to be assured they're getting the most up-to-date version of the images they care about.
+
+## Workflow Diagram
+![Image](images/base-image-dependency-flow-diagram.png)
```
MichaelSimons: The line between `Base Image Subscriptions` and `Dockerfiles` is confusing.  The line type isn't described in the key.  

It seems like there is a Data Flow missing from Check Base Images Pipeline to the Base Image Subscriptions.

mthalman: Ok, I'll fix the arrow lines there.  The intent was that all of the lines from the 3 different files converge and all point to `Load Dockerfile metadata for each subscription`.  But I can see how this could be confusing.  I'll use separate lines or at least shift things so it doesn't look like one continuous line from the top to the bottom.

MichaelSimons: Looks good now!

PR #224, opened 2019-07-05
Documentation for manifest file

### LINKED ISSUE #9: Add manifest documentation
Create documentation that describes the purpose of the manifest and details of the schema.

Detailed docs should be placed in a new /Documentation folder.

mthalman:
Fixes #9


MichaelSimons:
One concern I have is keeping this up to date over time.  To date there are very few consumers wanting to know this info especially compared to the number of times the schema has changed.  While this format is nice, I am wondering if it wouldn't be better to capture the content (e.g. descriptions) in code and utilize a tool (e.g. https://www.newtonsoft.com/jsonschema/help/html/GenerateWithDescriptions.htm) to create a more friendly holistic view for consumers.  I feel like having the descriptions in code would lend itself better to being kept up to date as the code evolves.

MichaelSimons:
Another benefit of utilizing a tool like the JSchemaGenerator is it would capture the required state and any other constraints such as null handling.

documentation/manifest-file-schema-reference.md
```diff
@@ -0,0 +1,71 @@
...
+|mcrTagsMetadataTemplatePath|string|Relative path to the MCR tags template YAML file that is used by tooling to generate the tags section of the readme file.|
+|images|array|The set of images contained in this repository. See [Images](#Images).|
+
+## Images
+
+An image object contains metadata about a specific Docker image.
+
+|Attribute|Type|Description|
+|-|-|-|
+|sharedTags|object|The set of [tags](#Tags) that are shared amongst all platform-specific versions of the image. An example of a shared tag, including its repo name, is dotnet/core/runtime:2.2; running `docker pull mcr.microsoft.com/dotnet/core/runtime:2.2` on Windows will get me the default Windows-based tag whereas running it on Linux will get me the default Linux-based tag.|
```
MichaelSimons: Missing the Platforms array.

Microsoft.DotNet.ImageBuilder/src/Models/Manifest/Tag.cs
```diff
@@ -2,16 +2,38 @@
...
+        )]
     public class Tag
     {
+        [Description(
+            "An identifier used to conceptually group related tags in the readme " +
+            "documentation."
+            )]
         public string DocumentationGroup { get; set; }
 
+        // TODO: Is this needed anymore? Not seeing it used in manifest files.
```
MichaelSimons: This was originally added [here](https://github.com/dotnet/docker-tools/pull/70).  I think we are exclusively utilizing the "islocal" attribute rather that modeled args utilizing TagRef variables.  It does look like this could be cleaned up, but IMO should happened with a separate PR.

Microsoft.DotNet.ImageBuilder/src/Models/Manifest/Tag.cs
```diff
@@ -2,16 +2,38 @@
...
+        [Description(
+            "Indicates whether the image should only be tagged with this tag on the " +
+            "local machine that builds the image. The published image will not include " +
+            "this tag. This is only used for advanced build dependency scenarios."
+            )]
         public bool IsLocal { get; set; }
 
+        [Description(
+            "Indicates whether this tag should not be documented in the readme file. The " +
+            "image will still be tagged with this tag however. This is useful when " +
```
MichaelSimons: Consider calling out publish..."will still be tagged and published..."

documentation/manifest-file.md
```diff
@@ -0,0 +1,5 @@
+# Manifest File
+
+The manifest file is the primary source of metadata that drives the production of all .NET Docker images.  It describes various attributes of the Docker images that are to be produced by a given GitHub repo. .NET Docker's engineering system consumes this file in various ways as part of the automated build pipelines and other tools. It's intended to be product-agnostic meaning that it could be used to describe metadata for Docker image production of any product, not just .NET.
+
+For a description of the schema, see the source location at https://github.com/dotnet/docker-tools/tree/master/Microsoft.DotNet.ImageBuilder/src/Models/Manifest.
```
MichaelSimons: Do you see value in having the Image-builder tool be able to emit a schema that makes model easier to view as a single cohesive unit?  It would be easy to do with a simple command.

Microsoft.DotNet.ImageBuilder/src/Models/Manifest/Platform.cs
```diff
@@ -10,30 +10,60 @@
...
         public IDictionary<string, string> BuildArgs { get; set; }
 
+        [Description(
+            "Relative path to the associated Dockerfile."
+            )]
         [JsonProperty(Required = Required.Always)]
         public string Dockerfile { get; set; }
 
+        [Description(
+            "The generic name of the operating system associated with the image."
```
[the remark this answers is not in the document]
mthalman: Why are examples needed?  It's an enum.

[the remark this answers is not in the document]
MichaelSimons: I retract my comment :)

Microsoft.DotNet.ImageBuilder/src/Models/Manifest/Tag.cs
mthalman: Logged an issue to remove this: #232

documentation/manifest-file.md
MichaelSimons: I think it would be worthwhile to mention the `ShowManifestSchema` command here as an alternative that produces something that is IMO more concise/readable.

PR #231, opened 2019-07-17
Add unit tests for RebuildStaleImagesCommand

mthalman:
Adds unit tests for RebuildStaleImagesCommand.  This required refactoring the command to support dependency injection since it had several hard-wired dependencies on external services like Docker, GitHub, and Azure DevOps REST API.  At this point, this has been done piecemeal, enough to support just the APIs that this command uses.  As more tests get added for Image Builder, these APIs can be updated to handle more scenarios.


Microsoft.DotNet.ImageBuilder/tests/RebuildStaleImagesCommandTests.cs
```diff
@@ -0,0 +1,1457 @@
...
+                    RepoInfo = new GitRepo
+                    {
+                        Branch = "testBranch",
+                        Name = repo1,
+                        Owner = GetRepoOwner()
+                    }
+                }
+            };
+
+            Dictionary<Subscription, Manifest> subscriptionManifests =
```
MichaelSimons: Consider factoring out these structures that are declared multiple times.  They are significant in size.  I think this would help in maintenance and readability.

Microsoft.DotNet.ImageBuilder/tests/RebuildStaleImagesCommandTests.cs
```diff
@@ -0,0 +1,1457 @@
...
+                    {
+                        throw new ArgumentNullException(nameof(expectedPathsBySubscription));
+                    }
+
+                    foreach (KeyValuePair<Subscription, IList<string>> kvp in expectedPathsBySubscription)
+                    {
+                        if (kvp.Value.Any())
+                        {
+                            this.buildHttpClientMock
+                                .Verify(o => o.QueueBuildAsync(It.Is<Build>(build => FilterBuildToSubscription(build, kvp.Key, kvp.Value))));
```
MichaelSimons: (nit) There are several long lines in this file.  Consider wrapping to fit in GitHubs view port.

Microsoft.DotNet.ImageBuilder/tests/RebuildStaleImagesCommandTests.cs
```diff
@@ -0,0 +1,1457 @@
...
+                        new DockerfileInfo(sdkDockerfilePath, null, null, hasInternalFrom: true),
+                        new DockerfileInfo(aspnetDockerfilePath, null, null, hasInternalFrom: true)
+                    }
+                }
+            };
+
+            using (TestContext context =
+                new TestContext(imageInfoData, subscriptions, subscriptionManifests, dockerfileInfos))
+            {
+                await context.ExecuteAsync();
```
MichaelSimons: I find the readability of these test cases to be rather difficult.  

1. The core test logic is dwarfed by the many lines that initialize data structures.
2. The TestContext.ExecuteAsync method doesn't convey what is being executed/tested without digging into the implementation.

Microsoft.DotNet.ImageBuilder/tests/RebuildStaleImagesCommandTests.cs
```diff
@@ -0,0 +1,1457 @@
...
+
+                Dictionary<Subscription, IList<string>> expectedPathsBySubscription =
```
MichaelSimons: It feels like it would be more interesting if the manifest included DockerfilePaths that weren't included in the expected.

Microsoft.DotNet.ImageBuilder/tests/RebuildStaleImagesCommandTests.cs
mthalman: Any suggestions?  There are a lot of data dependencies here so that's going to be required regardless.  The intent here was to simplify the amount of code in each test case and to factor things out into common code.  There's a lot of mocks to wire up and have them return the appropriate data.  I feel that the current code is preferable than moving all of that common code into each of the test cases.

Microsoft.DotNet.ImageBuilder/tests/RebuildStaleImagesCommandTests.cs
```diff
@@ -0,0 +1,1176 @@
...
+                Mock<IVssConnectionFactory> connectionFactoryMock = CreateVssConnectionFactoryMock(
+                    projectHttpClientMock, this.buildHttpClientMock);
+
+                this.httpClientFactory = CreateHttpClientFactory(subscriptions, subscriptionManifests, dockerfileInfos);
+
+                this.DockerServiceMock = this.CreateDockerServiceMock();
+                this.command = this.CreateCommand(connectionFactoryMock);
+            }
+
+            public Task ExecuteAsync()
```
MichaelSimons: I think it would help the readability to rename this to something like `ExecuteCommandAsync`.  `ExecuteAsync` in conjunction with the class name 'TestContext' doesn't provide any insight on what is being executed.

Microsoft.DotNet.ImageBuilder/tests/RebuildStaleImagesCommandTests.cs
MichaelSimons: Yes this is a hard one.  The main thing I was wondering if there was a way to shorten the data initialization.  Your recent changes help but it is still a lot to parse through on the first read.

Microsoft.DotNet.ImageBuilder/src/ImageBuilder.cs
```diff
@@ -4,33 +4,53 @@
...
+            }
+        }
+
         public static int Main(string[] args)
         {
             int result = 0;
 
             try
             {
                 ICommand[] commands = {
```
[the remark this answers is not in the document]
mthalman: Yes, indeed. Good suggestion

Microsoft.DotNet.ImageBuilder/src/ILoggerService.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public interface ILoggerService
+    {
+        void WriteError(string error);
+        void WriteHeading(string heading);
+        void WriteMessage();
+        void WriteMessage(string message);
+        void WriteSubheading(string subheading);
+    }
+}
```
MichaelSimons: EOF

mthalman: Darn VS-generated files.

PR #234, opened 2019-07-23
Include simple tags when merging image info

### LINKED ISSUE #233: MergeImageInfoCommand needs to account for SimpleTags property
The [ImageData.SimpleTags](https://github.com/dotnet/docker-tools/blob/a3e3204614c88a906f89118462acca4ed29db808/Microsoft.DotNet.ImageBuilder/src/Models/Image/ImageData.cs#L14) property is not being accounted for in the MergeImageInfoCommand class when it merges the image info result of a build into https://github.com/dotnet/versions/commits/master/build-info/docker/image-info.json.  As a result, the output always contains a set of empty tags for all the images.

mthalman:
Ensures that simple tags gets merged appropriately when merging image info data.  I've generalized the algorithm to use reflection to ensure that every property is accounted for to prevent issues like this in the future.  Updates unit tests to verify the scenario.

Fixes #233

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
@@ -2,8 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using Microsoft.DotNet.ImageBuilder.Models.Image;
 
 namespace Microsoft.DotNet.ImageBuilder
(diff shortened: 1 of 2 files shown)


Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
```diff
@@ -21,53 +24,88 @@ public static void MergeRepos(RepoData[] srcRepos, List<RepoData> targetRepos)
...
+            IList<string> srcList = (IList<string>)property.GetValue(srcObj);
+            if (srcList == null)
             {
                 return;
             }
 
-            if (srcImage.BaseImages.Any() && targetImage.BaseImages == null)
+            IList<string> targetList = (IList<string>)property.GetValue(targetObj);
+
+            if (srcList.Any() && targetList == null)
```
MichaelSimons: The srcList.Any() condition seems strange.  If srcList is empty, then we go through the union path?  What happens if targetList is null in that case?

```
if (srcList.Any())
{
    if (targetList != null)
    {
        targetList = targetList
                .Union(srcList)
                .OrderBy(element => element)
                .ToList();
    }
    else
    {
        targetList = srcList;
    }

    property.SetValue(targetObj, targetList);
}
```

Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
```diff
@@ -21,53 +24,88 @@ public static void MergeRepos(RepoData[] srcRepos, List<RepoData> targetRepos)
...
+            IDictionary srcDict = (IDictionary)property.GetValue(srcObj);
+            if (srcDict == null)
             {
-                targetImage.BaseImages[srcKvp.Key] = srcKvp.Value;
+                return;
+            }
+
+            IDictionary targetDict = (IDictionary)property.GetValue(targetObj);
+
+            if (srcDict.Cast<object>().Any() && targetDict == null)
```
MichaelSimons: Same comment as above on the source being an empty dictionary.

Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
```diff
@@ -21,53 +24,88 @@ public static void MergeRepos(RepoData[] srcRepos, List<RepoData> targetRepos)
...
+                    MergeDictionaries(property, srcObj, targetObj);
+                }
+                else if (typeof(IList<string>).IsAssignableFrom(property.PropertyType))
+                {
+                    MergeLists(property, srcObj, targetObj);
                 }
                 else
                 {
-                    targetRepo.Images.Add(srcKvp.Key, srcKvp.Value);
+                    throw new NotSupportedException("Unsupported model property type:" + property.PropertyType.FullName);
```
MichaelSimons:
```suggestion
                    throw new NotSupportedException($"Unsupported model property type: 'property.PropertyType.FullName'");
```

PR #239, opened 2019-07-31
Image info file support for CopyAcrImages cmd

mthalman:
Updates the CopyAcrImages command to optionally incorporate the data from the image info file.  This is necessary to handle image tags that make use of dynamically generated values, such as timestamps.  Without these changes, Image Builder would calculate a new tag value each time it is invoked.  This is obviously not good when we're trying to reference the tag of an image that was built in a separate build step.

By making use of the image info file that was generated when the image was built, we know the exact static value of the tag that needs to be referenced.  So when the CopyAcrImages command runs and finds a Platform to process, it looks up the tag value in the image info file instead of using the value that was calculated when parsing the manifest.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
@@ -52,48 +52,45 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 .Select(repo =>
                     repo.FilteredImages
                         .SelectMany(image => image.FilteredPlatforms)
-                        .Select(platform => GetDestinationTagNames(repo, platform))
-                        .Select(tags => ImportImage(azure, tags, registryName)))
+                        .SelectMany(platform => GetDestinationTagNames(repo, platform))
+                        .Select(tag => ImportImage(azure, tag, registryName)))
                 .SelectMany(tasks => tasks);
 
             await Task.WhenAll(importTasks);
         }
 
-        private async Task ImportImage(IAzure azure, IEnumerable<string> destTagNames, string registryName)
+        private async Task ImportImage(IAzure azure, string destTagName, string registryName)
         {
-            foreach (string destTagName in destTagNames)
+            string sourceTagName = destTagName.Replace(Options.RepoPrefix, Options.SourceRepoPrefix);
+            ImportImageParametersInner importParams = new ImportImageParametersInner()
             {
-                string sourceTagName = destTagName.Replace(Options.RepoPrefix, Options.SourceRepoPrefix);
-                ImportImageParametersInner importParams = new ImportImageParametersInner()
-                {
-                    Mode = "Force",
-                    Source = new ImportSource(
-                        sourceTagName,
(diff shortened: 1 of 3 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
```diff
@@ -22,46 +26,95 @@ public CopyAcrImagesCommand() : base()
...
+                .SelectMany(tasks => tasks);
 
-            Logger.WriteMessage($"Importing '{destTagName}' from '{sourceTagName}'");
+            await Task.WhenAll(importTasks);
+        }
 
-            if (!Options.IsDryRun)
+        private async Task ImportImage(IAzure azure, RepoInfo repo, PlatformInfo platform, string registryName)
+        {
+            List<string> destTagNames = null;
```
MichaelSimons: From a readability perspective, this is a rather large method that has two distinct responsibilities.  How about breaking out the logic that retrieves the `destTagNames` into its own method?  This could be invoked as part of the calling LINQ query.  

```
repo.FilteredImages
    .SelectMany(image => image.FilteredPlatforms)
    .SelectMany(platform => GetTags(repo, platform)
    .Select(tag=> ImportImage(azure, tag, registryName))
```

Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
```diff
@@ -22,46 +26,95 @@ public CopyAcrImagesCommand() : base()
...
+
+            // If an image info file was provided, use the tags defined there rather than the manifest. This is intended
+            // to handle scenarios where the tag's value is dynamic, such as a timestamp, and we need to know the value
+            // of the tag for the image that was actually built rather than just generating new tag values when parsing
+            // the manifest.
+            if (!String.IsNullOrEmpty(Options.ImageInfoPath))
             {
-                try
+                RepoData[] repos = JsonConvert.DeserializeObject<RepoData[]>(File.ReadAllText(Options.ImageInfoPath));
```
MichaelSimons: Reading and deserializing the image info for each platform feels like a pretty significant waste of CPU.  Can this only be done once?

Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
```diff
@@ -41,31 +52,63 @@ public override async Task ExecuteAsync()
                 .Select(repo =>
                     repo.FilteredImages
                         .SelectMany(image => image.FilteredPlatforms)
-                        .Select(platform => ImportImage(azure, repo, platform, registryName)))
+                        .Select(platform => GetDestinationTagNames(repo, platform))
```
MichaelSimons: Why not SelectMany and avoid the loop (and nesting) within ~~ImportImage~~ImportTag?  If you do this, GetDestinationTagNames should return IEnumerable with deferred execution.

mthalman: Ah, I see what you're saying.

PR #245, opened 2019-08-02
Support providing additional test args.

StephenBonikowsky:
* This is a port of changes made in https://github.com/microsoft/dotnet-framework-docker with the addition of the WCF and Asp.Net repos to be able to use a test argument to filter and only run applicable tests.

### DIFF
--- a/eng/common/templates/steps/test-images-windows-client.yml
+++ b/eng/common/templates/steps/test-images-windows-client.yml
@@ -11,6 +11,9 @@ steps:
     } else {
       $optionalTestArgs="-IsLocalRun"
     }
+    if ($env:REPOTESTARGS) {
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
Looks like this file also has an additional 'Docker login' step that does not exist in dotnet-framework-docker. I assume this is expected.

StephenBonikowsky:
@mthalman @MichaelSimons @HongGit

mthalman:
> Looks like this file also has an additional 'Docker login' step that does not exist in dotnet-framework-docker. I assume this is expected.

It has it right here: https://github.com/microsoft/dotnet-framework-docker/blob/6b69b568d888ee80f08bc5a3e3c9fbfb46020358/eng/common/templates/steps/test-images-windows-client.yml#L15

PR #247, opened 2019-08-02
Exclude IsLocal tags from image info output

### LINKED ISSUE #246: Build command includes image tags marked as IsLocal in image info data
When a tag in the manifest is marked as IsLocal, it means that tag will not be pushed when running the Build command with the --push option. When the Build command is executed, it produces an image info output file containing the tags of the image that was built.  But for tags marked as IsLocal, these tags were never pushed and so they should not be included in the image info file.  Currently, they are being included.

mthalman:
Fixes #246

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -113,7 +113,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                             imageData.BaseImages = baseImageDigests;
                         }
 
-                        imageData.SimpleTags = platform.Tags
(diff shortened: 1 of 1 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -289,6 +288,11 @@ private void PushImages()
             }
         }
 
+        private static IEnumerable<TagInfo> GetPushTags(IEnumerable<TagInfo> buildTags)
```
MichaelSimons: This is a good place to consider using an expression bodied function.  It is a preference.

```
private static IEnumerable<TagInfo> GetPushTags(IEnumerable<TagInfo> buildTags) =>
    buildTags.Where(tag => !tag.Model.IsLocal);
```

PR #248, opened 2019-08-05
Include platform dependencies for different .NET versions

### LINKED ISSUE #243: platformVersionedOs matrix is not including platform dependencies
The platformVersionedOs matrix type is used for PR builds but Image Builder has an issue with generating the correct matrix when a Dockerfile has a dependency on a Dockerfile from a different .NET version.  It doesn't include that dependency in the `imageBuilderPaths` attribute.  The consequence of this is that the dependent image doesn't exist locally so it gets automatically pulled down when building the depending image.  This means that we're testing with the wrong base image (remember this is for PR builds only so this doesn't actually get published).

An example of this scenario is the Dockerfile at `2.2/runtime/alpine3.9`.  This image depends on `runtime-deps:2.2-alpine3.9` which is actually built from the Dockerfile located at `2.1/runtime-deps/alpine3.9`.  Notice that that Dockerfile paths are from different .NET versions.  This is what causes this to break.  The `imageBuilderPaths` that get generated for this are `--path 2.2/runtime/alpine3.9/amd64 --path 2.2/aspnet/alpine3.9/amd64 --path 2.2/sdk/alpine3.9/amd64`.  Notice that `2.1/runtime-deps/alpine3.9` is not included in there.

mthalman:
Ensures that platform dependencies are included appropriately for the platformVersionedOs matrix type.

Fixes #243

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/LegInfo.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/LegInfo.cs
@@ -1,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class LegInfo
+    {
+        public string Name { get; set; }
+        public List<(string Name, string Value)> Variables { get; } = new List<(string, string)>();
+    }
+}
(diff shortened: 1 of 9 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/LegInfo.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class LegInfo
```
MichaelSimons: To me it feels like this type name should be qualified/more descriptive.  When it was a nested type, the containing class gave it context.  On its own the name is pretty ambiguous.  I could see the name causing confusion over time.  The same comment applies to `MatrixInfo`.  Maybe `BuildMatrixInfo` and `BuildLegInfo` would be better?

PR #269, opened 2019-08-09
Replace tags when merging image info

mthalman:
There's an issue with the image info merging logic that was exposed when moving a shared tag.  In this case, the shared tag `3.0-alpine` was moved from the Alpine 3.9-based images to Alpine 3.10 in the manifest.  The images were tagged appropriately but when the image info file was updated during the publish stage, the Alpine 3.9-based images in the image info continued to list `3.0-alpine` as a tag in addition to the Alpine 3.10-based images correctly listing `3.0-alpine`.  See this [commit](https://github.com/dotnet/versions/blob/31981d050666cd82e6dc472223ebf2908a916041/build-info/docker/image-info.dotnet-dotnet-docker-nightly.json#L185) for the issue.

To fix this, the merge logic is updated to always replace the `SimpleTags` list such that the source image's tag always overwrite the target's tags in the image info.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
@@ -73,10 +73,8 @@ namespace Microsoft.DotNet.ImageBuilder
             }
         }
 
-        private static void ReplaceValue(PropertyInfo property, object srcObj, object targetObj)
-        {
+        private static void ReplaceValue(PropertyInfo property, object srcObj, object targetObj) =>
             property.SetValue(targetObj, property.GetValue(srcObj));
-        }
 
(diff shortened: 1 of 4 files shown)


Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
```diff
@@ -57,6 +73,11 @@ private static void MergeData(object srcObj, object targetObj)
             }
         }
 
+        private static void ReplaceValue(PropertyInfo property, object srcObj, object targetObj)
```
MichaelSimons: For brevity consider using an expression bodied method

```
private static void ReplaceValue(PropertyInfo property, object srcObj, object targetObj) =>
    property.SetValue(targetObj, property.GetValue(srcObj));
```

Issue #274, opened 2019-08-15
check-base-image-updates pipeline should consolidate results and queue build once


mthalman:
Currently, the pipeline executes 4 jobs which each independently will queue a build when necessary for the images that it's responsible for.  It also has logic to prevent queuing multiple builds if a build is currently running.  This leads to inefficiency when two or more of the jobs encounter images that need to be rebuilt.  In that case, only first job will queue the build.  The other job will be skipped.  Not until the scheduled build runs again will it have an opportunity to pick up those image updates that we skipped.  This can be made more efficient by have each job simply collect the Dockerfile paths that need to be rebuilt.  Then have a single job at the end actually queue the build if necessary.

PR #275, opened 2019-08-16
Separate RebuildStaleImagesCommand into two separate commands

mthalman:
These changes break apart the RebuildStaleImagesCommand into two separate commands: GetStaleImagesCommand and QueueBuildCommand.  This is necessary for the check-base-image-updates pipeline so that it can determine which images need to be rebuilt across both Windows and Linux as well as the different architectures before actually queuing a build.  Once all the image paths have been determined, then the final part of the build should actually queue the necessary builds.

A given job that executes GetStaleImagesCommand will be looking for changes to images across all the subscribing Git repos/branches.  And the command will produce an output variable to be consumed by the pipeline.  So there will be four such variables whose data represents the same sets of subscriptions amongst all of them.  It's up to the QueueBuildCommand variable to take those variables as input and consolidate the data according to subscription and then queuing a build for that subscription if necessary.

Related to #274

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
@@ -57,11 +57,20 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                         ImagePaths = (await GetPathsToRebuildAsync(s, repos)).ToArray()
                     }));
 
+                // Filter out any results that don't have any images to rebuild
+                results = results
+                    .Where(result => result.ImagePaths.Any())
+                    .ToArray();
+
                 string outputString = JsonConvert.SerializeObject(results);
 
                 this.loggerService.WriteMessage(
                     PipelineHelper.FormatOutputVariable(Options.VariableName, outputString)
                         .Replace("\"", "\\\"")); // Escape all quotes
+
+                string formattedResults = JsonConvert.SerializeObject(results, Formatting.Indented);
+                this.loggerService.WriteMessage(
+                    $"Image Paths to be Rebuilt:{Environment.NewLine}{formattedResults}");
             }
             finally
             {
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
@@ -51,16 +51,13 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             try
             {
                 var results = await Task.WhenAll(
-                    subscriptions.Select(async s => new
+                    subscriptions.Select(async s => new SubscriptionImagePaths
                     {
-                        Subscription = s,
-                        Paths = await GetPathsToRebuildAsync(s, repos)
+                        SubscriptionId = s.Id,
+                        ImagePaths = (await GetPathsToRebuildAsync(s, repos)).ToArray()
                     }));
 
-                Dictionary<string, string> outputDictionary =
-                    results.ToDictionary(result => result.Subscription.Id, result => FormatPaths(result.Paths));
-
-                string outputString = JsonConvert.SerializeObject(outputDictionary);
+                string outputString = JsonConvert.SerializeObject(results);
 
                 this.loggerService.WriteMessage(
                     PipelineHelper.FormatOutputVariable(Options.VariableName, outputString)
@@ -75,18 +72,6 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                     Directory.Delete(new DirectoryInfo(repoPath).Parent.FullName, true);
                 }
             }
-        }
-
-        private static string FormatPaths(IEnumerable<string> paths)
-        {
-            if (!paths.Any())
-            {
-                return String.Empty;
-            }
-
-            return paths
-                .Select(path => $"{ManifestFilterOptions.FormattedPathOption} '{path}'")
-                .Aggregate((p1, p2) => $"{p1} {p2}");
         }
 
         private async Task<IEnumerable<string>> GetPathsToRebuildAsync(Subscription subscription, RepoData[] repos)
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
@@ -13,37 +13,29 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DotNet.ImageBuilder.Models.Image;
 using Microsoft.DotNet.ImageBuilder.Models.Subscription;
-using Microsoft.DotNet.ImageBuilder.Services;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
-using Microsoft.TeamFoundation.Build.WebApi;
-using Microsoft.TeamFoundation.Core.WebApi;
-using Microsoft.VisualStudio.Services.Common;
-using Microsoft.VisualStudio.Services.WebApi;
 using Newtonsoft.Json;
 
 namespace Microsoft.DotNet.ImageBuilder.Commands
 {
     [Export(typeof(ICommand))]
-    public class RebuildStaleImagesCommand : Command<RebuildStaleImagesOptions>, IDisposable
+    public class GetStaleImagesCommand : Command<GetStaleImagesOptions>, IDisposable
     {
         private readonly Dictionary<string, string> gitRepoIdToPathMapping = new Dictionary<string, string>();
         private readonly Dictionary<string, string> imageDigests = new Dictionary<string, string>();
         private readonly SemaphoreSlim gitRepoPathSemaphore = new SemaphoreSlim(1);
         private readonly SemaphoreSlim imageDigestsSemaphore = new SemaphoreSlim(1);
         private readonly IDockerService dockerService;
-        private readonly IVssConnectionFactory connectionFactory;
         private readonly ILoggerService loggerService;
         private readonly HttpClient httpClient;
 
         [ImportingConstructor]
-        public RebuildStaleImagesCommand(
+        public GetStaleImagesCommand(
             IDockerService dockerService,
-            IVssConnectionFactory connectionFactory,
             IHttpClientFactory httpClientFactory,
             ILoggerService loggerService)
         {
             this.dockerService = dockerService;
-            this.connectionFactory = connectionFactory;
             this.loggerService = loggerService;
             this.httpClient = httpClientFactory.GetClient();
         }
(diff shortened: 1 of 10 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesOptions.cs
```diff
@@ -0,0 +1,49 @@
...
+                "image-info-path",
+                ref imageDataPath,
+                $"Path to the file containing image info (defaults to '{DefaultImageInfoPath}').");
+            ImageInfoPath = imageDataPath;
+
+            string variableName = null;
+            syntax.DefineParameter(
+                "variable",
+                ref variableName,
+                "The variable name to assign the image paths to");
```
MichaelSimons: The terminology/description here is ambiguous.  In the context of ImageBuilder, `variable` has it's own meaning.  This command is intended to be used withing an AzDO pipeline and this is an AzDO variable.  I think this should be called out in the help output.

Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                if (consolidatedSubscriptionsData.Any())
+                {
+                    await Task.WhenAll(
+                        consolidatedSubscriptionsData.Select(
+                            kvp => QueueBuildForStaleImages(subscriptions.First(sub => sub.Id == kvp.Key), kvp.Value)));
+                }
+                else
+                {
+                    this.loggerService.WriteMessage($"None of the subscriptions have base images that are out-of-date. No rebuild necessary.");
```
MichaelSimons: Long line

Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            this.connectionFactory = connectionFactory;
+            this.loggerService = loggerService;
+        }
+
+        public override async Task ExecuteAsync()
+        {
+            string subscriptionsJson = File.ReadAllText(Options.SubscriptionsPath);
+            Subscription[] subscriptions = JsonConvert.DeserializeObject<Subscription[]>(subscriptionsJson);
+            Dictionary<string, string> consolidatedSubscriptionsData = GetConsolidatedSubscriptions();
```
MichaelSimons: I was confused by this when I first look at the code.  To me the `consolidatedSubscriptions` terminology doesn't help the readability of this logic.   The first part is that it is more information than just subscriptions.  Second, from this method, the primary piece to me is that this is only the data about the builds to queue, e.g. the stale subscriptions info.  The details of the consolidation don't seem particularly important here.

Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
```diff
@@ -0,0 +1,148 @@
...
+                            kvp => QueueBuildForStaleImages(subscriptions.First(sub => sub.Id == kvp.Key), kvp.Value)));
+                }
+                else
+                {
+                    this.loggerService.WriteMessage($"None of the subscriptions have base images that are out-of-date. No rebuild necessary.");
+                }
+            }
+            finally
+            {
+                foreach (string repoPath in gitRepoIdToPathMapping.Values)
```
MichaelSimons: Is this being used?  Why does this command need the repo?

Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
```diff
@@ -58,7 +50,21 @@ public override async Task ExecuteAsync()
 
             try
             {
-                await Task.WhenAll(subscriptions.Select(s => QueueBuildForStaleImages(s, repos)));
+                var results = await Task.WhenAll(
+                    subscriptions.Select(async s => new
```
MichaelSimons: It feels like the data _contract_ should be well defined between the GetStaleImages and QueueBuild commands.  Using an anonymous type here doesn't feel like the best.

Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
```diff
@@ -58,7 +50,21 @@ public override async Task ExecuteAsync()
...
+                    {
+                        Subscription = s,
+                        Paths = await GetPathsToRebuildAsync(s, repos)
+                    }));
+
+                Dictionary<string, string> outputDictionary =
+                    results.ToDictionary(result => result.Subscription.Id, result => FormatPaths(result.Paths));
```
MichaelSimons: Wouldn't it be good to filter out subscriptions with no paths (no stale images)?  It feels like this would be a good optimization to avoid unnecessary overhead in serializing/deserializing.

Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildOptions.cs
```diff
@@ -32,13 +30,13 @@ public override void ParseCommandLine(ArgumentSyntax syntax)
...
             SubscriptionsPath = subscriptionsPath;
 
-            const string DefaultImageInfoPath = "image-info.json";
-            string imageDataPath = DefaultImageInfoPath;
-            syntax.DefineOption(
-                "image-info-path",
-                ref imageDataPath,
-                $"Path to the file containing image info (defaults to '{DefaultImageInfoPath}').");
-            ImageInfoPath = imageDataPath;
+            // These values come from the output variable of GetStaleImagesCommand
```
MichaelSimons: Why not include this in the help text?

Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildOptions.cs
```diff
@@ -32,13 +30,13 @@ public override void ParseCommandLine(ArgumentSyntax syntax)
...
+            IReadOnlyList<string> subscriptions = Array.Empty<string>();
+            syntax.DefineOptionList(
+                "subscriptions",
+                ref subscriptions,
+                "Subscriptions data describing paths to be built");
```
MichaelSimons: This is where a type that establishes the data contract would be helpful IMO.  `Subscriptions data` does not describe the structure of the data that is expected.  Lastly the overloading of `subscriptions` is confusing.  Can the reader understand the difference between what is specified here versus what is read from the `subscriptions-path` option?

Microsoft.DotNet.ImageBuilder/src/Models/Subscription/Subscription.cs
```diff
@@ -17,9 +17,11 @@ public class Subscription
         [JsonProperty(Required = Required.Always)]
         public PipelineTrigger PipelineTrigger { get; set; }
 
+        public string Id => $"{RepoInfo.Owner}/{RepoInfo.Name}/{RepoInfo.Branch}/{ManifestPath}";
+
         public override string ToString()
         {
-            return $"{RepoInfo.Owner}/{RepoInfo.Name}/{RepoInfo.Branch}/{ManifestPath}";
+            return Id;
```
MichaelSimons: This would be a good place for an expression-bodied method.

Microsoft.DotNet.ImageBuilder/src/PipelineHelper.cs
```diff
@@ -0,0 +1,14 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    public static class PipelineHelper
+    {
+        public static string FormatOutputVariable(string variableName, string value)
+        {
+            return $"##vso[task.setvariable variable={variableName};isoutput=true]{value}";
```
MichaelSimons: Another opportunity to use an expression-bodied method.

Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
```diff
@@ -0,0 +1,148 @@
...
+            // of the images that need to be built. A given subscription may have images that are spread across Linux/Windows, AMD64/ARM
+            // which means that the paths collected for that subscription were spread across multiple jobs.  Each of the items in the 
+            // Enumerable here represents the data collected by one job.  We need to consolidate the paths for a given subscription since
+            // they could be spread across multiple items in the Enumerable.
+            IEnumerable<Dictionary<string, string>> allSubscriptionsData = Options.Subscriptions
+                .Select(subscriptionsData => JsonConvert.DeserializeObject<Dictionary<string, string>>(subscriptionsData))
+                .ToList();
+
+            Dictionary<string, string> consolidatedSubscriptionsData = new Dictionary<string, string>();
+            foreach (Dictionary<string, string> subscriptionData in allSubscriptionsData)
```
MichaelSimons: Using Linq could be utilized more here to avoid the nested foreach loops

```
            IEnumerable<KeyValuePair<string, string>> subscriptionData = Options.Subscriptions
                .Select(subscription => JsonConvert.DeserializeObject<Dictionary<string, string>>(subscription))
                .SelectMany(data => data.Where(kvp => !String.IsNullOrEmpty(kvp.Value)));
```

Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
```diff
@@ -71,61 +77,16 @@ public override async Task ExecuteAsync()
             }
         }
 
-        private async Task QueueBuildForStaleImages(Subscription subscription, RepoData[] repos)
+        private static string FormatPaths(IEnumerable<string> paths)
         {
-            IEnumerable<string> pathsToRebuild = await GetPathsToRebuildAsync(subscription, repos);
-
-            if (!pathsToRebuild.Any())
+            if (!paths.Any())
```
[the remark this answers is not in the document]
mthalman: Yes, it's necessary.  The call to `Aggregate` fails if the collection is empty.

Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
mthalman: Nope, that was leftover from the original implementation.  Removed.

Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
```diff
@@ -58,7 +50,18 @@ public override async Task ExecuteAsync()
...
+                var results = await Task.WhenAll(
+                    subscriptions.Select(async s => new SubscriptionImagePaths
+                    {
+                        SubscriptionId = s.Id,
+                        ImagePaths = (await GetPathsToRebuildAsync(s, repos)).ToArray()
+                    }));
+
+                string outputString = JsonConvert.SerializeObject(results);
+
+                this.loggerService.WriteMessage(
```
MichaelSimons: Thoughts on logging more info for diagnostic purposes? The variable output gets masked by AzDO and doesn't appear in the output unless debug is turned on.  This is not very helpful IMO when something goes wrong/unexpected with a build because you have to reproduce the issue with debug=true.

PR #276, opened 2019-08-16
Update check-base-image-updates pipeline

### LINKED ISSUE #274: check-base-image-updates pipeline should consolidate results and queue build once

mthalman:
These are the corresponding pipeline changes for https://github.com/dotnet/docker-tools/pull/275.

Fixes #274


eng/pipelines/templates/steps/queue-build.yml
```diff
@@ -0,0 +1,17 @@
+parameters:
```
MichaelSimons: What value does encapsulating this within a template provide?

eng/pipelines/templates/steps/queue-build.yml
```diff
@@ -0,0 +1,17 @@
...
+  subscriptions: []
```
MichaelSimons: This doesn't appear to be used.  Am I missing something?

eng/pipelines/templates/steps/queue-build.yml
```diff
@@ -0,0 +1,17 @@
...
+steps:
+  - template: ../../../common/templates/steps/init-docker-linux.yml
+  - script: >
+      $(runImageBuilderCmd)
+      queueBuild
+      $(System.AccessToken)
+      dnceng
+      internal
+      --subscriptions-path eng/check-base-image-subscriptions.json
```
MichaelSimons: Consider encapsulating the subscriptions value in a shared variable since it is referenced in multiple locations now.

eng/pipelines/check-base-image-updates.yml
```diff
@@ -4,15 +4,15 @@ pr: none
 variables:
 - template: templates/variables/common.yml
 
-# Sequentially execute the jobs to avoid race condition and having multiple builds getting queued
 jobs:
 - job: Build_Linux_AMD
```
MichaelSimons: The "Build" job names seem out of date now.

eng/pipelines/check-base-image-updates.yml
```diff
@@ -49,10 +49,25 @@ jobs:
...
   steps:
-  - template: templates/steps/queue-build-for-stale-images.yml
+  - template: templates/steps/get-stale-images.yml
     parameters:
       osType: windows
       dockerClientOS: linux
       useRemoteDockerServer: true
+      staleImagePathsVariableName: windows-arm-stale-image-paths
+
+- job: Rebuild_Stale_Images
```
MichaelSimons: This name is a little deceptive when looking at the build output.  To me, the name should convey it simply queues the builds and doesn't actually do the work to rebuild images.

eng/pipelines/templates/steps/queue-build.yml
mthalman: Good point

eng/pipelines/templates/steps/queue-build.yml
mthalman: Nope, it was leftover from an experiment.  Removed

eng/pipelines/check-base-image-updates.yml
```diff
@@ -4,55 +4,83 @@ pr: none
...
+  steps:
+  - template: ../common/templates/steps/init-docker-linux.yml
+  - script: >
+      $(runImageBuilderCmd)
+      queueBuild
+      $(System.AccessToken)
+      dnceng
+      internal
+      --subscriptions-path $(checkBaseImageSubscriptionsPath)
+      --image-paths "$(imagePaths1)"
```
MichaelSimons: Is the variable necessary?  Can you directly reference the dependency here?

mthalman: No, I tried and got syntax errors.

PR #277, opened 2019-08-20
Build matrix fix: Don't include dependencies if not in context

### LINKED ISSUE #186: Allow leaf-node image in dependency graph to be built/tested/published on its own
Currently all the images in a dependency graph must be built, tested, and published as a set.  As an example, this means we can't make a change to the SDK image without also building the runtime-deps image and that means the runtime-deps image will also get published even though it hasn't necessarily changed.  Similarly, the SDK image can't be tested without also building the ASP.NET image because the tests require it.  This prevents us from making more fine-grained releases.  It forces us to publish images that haven't even been changed.

This should be fixed by allowing a leaf-node image to be built on its own.  Any dependencies it has should be retrieved from the publicly released images in MCR.

mthalman:
Allows images to be built without forcing the images they depend on to be built.  For example, we sometimes want to be able to publish an update to the SDK image but do not need to update the runtime image it depends upon.  This currently isn't possible because the matrix that gets generated will automatically include the base image(s).

These changes fix this by only including the base image graph in the outputted matrix if they match the path filter that was provided as input to the generateBuildMatrix command.

Fixes #186

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
@@ -36,7 +36,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             BuildMatrixInfo matrix, IEnumerable<string> matrixNameParts, IGrouping<PlatformId, PlatformInfo> platformGrouping)
         {
             IEnumerable<IEnumerable<PlatformInfo>> subgraphs = platformGrouping.GetCompleteSubgraphs(
-                platform => GetPlatformDependencies(platform).Intersect(platformGrouping));
+                platform => GetPlatformDependencies(platform, platformGrouping));
 
             foreach (IEnumerable<PlatformInfo> subgraph in subgraphs)
             {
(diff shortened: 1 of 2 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -35,7 +35,9 @@ public override Task ExecuteAsync()
         private void AddDockerfilePathLegs(
             BuildMatrixInfo matrix, IEnumerable<string> matrixNameParts, IGrouping<PlatformId, PlatformInfo> platformGrouping)
         {
-            IEnumerable<IEnumerable<PlatformInfo>> subgraphs = platformGrouping.GetCompleteSubgraphs(GetPlatformDependencies);
+            IEnumerable<IEnumerable<PlatformInfo>> subgraphs = platformGrouping.GetCompleteSubgraphs(
```
MichaelSimons: Since GetCompleteSubgraphs is utilized in `AddVersionedOsLegs` I was wondering if it too should change as well.  My cursory search doesn't show any need for the `imageBuilderPaths` in the `PlatformVersionedOs` therefore the call wouldn't be needed.  Can you follow up to see if this can indeed be cleaned up and if not, should this change be reflected?

mthalman: `imageBuilderPaths` is used for `PlatformVersionedOs` when that matrix type is used for PR builds.    I've updated the logic so that it's consistent between the two matrix types.

PR #302, opened 2019-09-12
Use Dockerfile path for image info

### LINKED ISSUE #299: Image info should be based on Dockerfile path, not build context path
An issue recently occurred when generating the image info command for the .NET Fx sample images.  The WCF samples define in the manifest multiple images that share the same build context path but make use of different Dockerfiles.  The image info file that gets generated only stores the data based on the build context path (the folder where the Dockerfile is located) so it was misrepresenting the data and not separating out the data for each of the multiple images.  This ends up omitting the tags from the images produced by the additional Dockerfiles.  When the CopyAcrImages command is called, it will use the tag data in the image info file to publish the images.  Since not all the tags are being represented in the image info file, those images don't get published.

mthalman:
Changes the way that images are referenced in the image info file.  Previously, the images were identified by their build context path (the folder where the Dockerfile is located).  But this doesn't work in cases where we build images from Dockerfiles where multiple Dockerfiles are located in the same folder, such as with .NET Fx sample images (see #299).  So this is now changed to use the Dockerfile path that is set in the manifest file.

I've added tests for all the commands that were changed.  This caused a little bit of churn to the code to account for this.  DockerService has been extended to include some additional operations needed by BuildCommand.

A testing issue was also identified that caused tests to fail when run in parallel.  This was due to the test setting the current working directory.  Since the current working directory applies to the entire process, this caused issues when multiple tests were running in parallel and changing the current working directory.  So that pattern has been removed and the code updated to use absolute paths so that the current working directory is not relied upon.

Fixes #299


Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -64,7 +68,7 @@ private void BuildImages()
                     foreach (PlatformInfo platform in image.FilteredPlatforms)
                     {
                         ImageData imageData = new ImageData();
-                        images.Add(platform.BuildContextPath, imageData);
+                        images.Add(platform.Model.Dockerfile, imageData);
```
MichaelSimons: Would it be better to use the platform.DockerFilePath property instead?  This property returns a normalized path e.g. a path that includes the file name.  The Model.Dockerfile path may be a directory or a path to a folder with a file named `Dockerfile`.  This is done for convenience reasons for manifest authors and aligns with how the docker CLI will default to use a `Dockerfile`.  The image-info files should be explicit IMO and fully qualified IMO.

Microsoft.DotNet.ImageBuilder/src/DockerServiceExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.DotNet.ImageBuilder.Commands;
+using Microsoft.DotNet.ImageBuilder.ViewModel;
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    public static class DockerServiceExtensions
+    {
+        public static void PullBaseImages(this IDockerService dockerService, ManifestInfo manifest, ManifestOptions options)
```
MichaelSimons: I think it is a desirable design goal to not add dependencies to the command types outside of the Commands namespace.  In this particular, case, I think passing the ManifestOptions should be replaced with a single `bool isDryRun` parameter.  This would align the method with the other IDockerService methods as well.

Microsoft.DotNet.ImageBuilder/src/Services/IAzureManagementFactory.cs
```diff
@@ -0,0 +1,14 @@
...
+using Microsoft.Azure.Management.Fluent;
+using Microsoft.Azure.Management.ResourceManager.Fluent.Authentication;
+
+namespace Microsoft.DotNet.ImageBuilder.Services
+{
+    public interface IAzureManagementFactory
+    {
+        IAzure CreateAzureManager(AzureCredentials credentials, string subscriptionId);
+    }
+}
```
MichaelSimons: Missing line feed.

Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -50,7 +50,12 @@ public static PlatformInfo Create(Platform model, string fullRepoModelName, stri
             platformInfo.FullRepoModelName = fullRepoModelName;
             platformInfo.VariableHelper = variableHelper;
 
-            if (File.Exists(model.Dockerfile))
+            // Ensure that we construct an absolute path here to check for the file. Using a relative path
+            // has a dependency on the current working directory which is set for the entire process and
+            // is incompatible with the parallelism that is used in the GetStaleImagesCommand.  That command
+            // processes multiple manifests in varying directories in parallel, so relying on current working
+            // directory is not possible in that case.
+            if (File.Exists(Path.Combine(baseDirectory, model.Dockerfile)))
```
MichaelSimons: Should this be checking for a rooted path first?  The manifest should be flexible in the type of paths it supports.

Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -68,7 +68,7 @@ private void BuildImages()
                     foreach (PlatformInfo platform in image.FilteredPlatforms)
                     {
                         ImageData imageData = new ImageData();
-                        images.Add(platform.Model.Dockerfile, imageData);
+                        images.Add(platform.DockerfilePath, imageData);
```
MichaelSimons: Does this cause the base working directory to get published in the image-info within the versions repo?  If so, it feels like it should be stripped.

mthalman: Not in practice because we don't provide absolute paths for the manifest.json file.  The comments in `PlatformInfo` were incorrect when they said the convert to an absolute path; I've corrected them.  I did, however, make some changes to provide consistency in how the paths are set depending on whether the Dockerfile filename was left out of the path in the manifest or not.

PR #307, opened 2019-09-17
Filter by OSType for GetStaleImages subscriptions

### LINKED ISSUE #288: GetStaleImages command should provide a way to exclude an OS type
We'll need a way for the GetStaleImages command to ignore images of a given OS type for a certain subscription.  For example, this will be needed in the case of dotnet-docker repo with its Windows images.  On Patch Tuesday, we want to control the release of when the images are published and not have the build automatically get kicked off should a single Windows image change be detected.

mthalman:
As described in #288, we need a way for the call to GetStaleImages to filter platforms by OS type with each subscription optionally declaring which OS type it wants to filter on.  

Fixes #288

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
@@ -9,6 +9,7 @@ using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Net.Http;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DotNet.ImageBuilder.Models.Image;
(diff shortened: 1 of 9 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
```diff
@@ -9,6 +9,7 @@
 using System.IO.Compression;
 using System.Linq;
 using System.Net.Http;
+using System.Text.RegularExpressions;
```
MichaelSimons: What changes warrant this new using?

mthalman: Mistakenly added, removed.

mthalman:
There's actually an issue with this implementation that I'm addressing. The `GetStaleImages` command gets executed by multiple agents with different OS types.  So the command itself is filtered with OsType which needs to be respected as well.  The current implementation is essentially overwriting that OsType with the OsType defined in the subscription.  This means if the command is executed on a Windows agent and is processing a subscription that is configured to filter to only the Linux OS type, the logic would end up trying to inspect Linux images which is obviously wrong.  I'm fixing this so that if the OS type of the subscription doesn't match the OS type of the command's filter, then it'll skip processing for that subscription.

PR #318, opened 2019-10-01
Consume repo-specific image info files for GetStaleImagesCommand

### LINKED ISSUE #281: Consume repo-specific image info files
Now that the single image info file has been broken into repo/branch-specific files (#265), the infra needs to be updated so that it can consume these individual files rather than just the single image info file.

mthalman:
Fixes #281

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
@@ -15,6 +15,8 @@ using System.Threading.Tasks;
 using Microsoft.DotNet.ImageBuilder.Models.Image;
 using Microsoft.DotNet.ImageBuilder.Models.Subscription;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
+using Microsoft.DotNet.VersionTools.Automation;
+using Microsoft.DotNet.VersionTools.Automation.GitHubApi;
 using Newtonsoft.Json;
 
 namespace Microsoft.DotNet.ImageBuilder.Commands
@@ -28,16 +30,19 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         private readonly SemaphoreSlim imageDigestsSemaphore = new SemaphoreSlim(1);
         private readonly IDockerService dockerService;
         private readonly ILoggerService loggerService;
+        private readonly IGitHubClientFactory gitHubClientFactory;
         private readonly HttpClient httpClient;
 
         [ImportingConstructor]
         public GetStaleImagesCommand(
             IDockerService dockerService,
             IHttpClientFactory httpClientFactory,
-            ILoggerService loggerService)
+            ILoggerService loggerService,
+            IGitHubClientFactory gitHubClientFactory)
         {
             this.dockerService = dockerService;
             this.loggerService = loggerService;
+            this.gitHubClientFactory = gitHubClientFactory;
             this.httpClient = httpClientFactory.GetClient();
         }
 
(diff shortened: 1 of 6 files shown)


Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
```diff
@@ -173,6 +176,22 @@ private async Task<IEnumerable<string>> GetPathsToRebuildAsync(Subscription subs
...
+        private async Task<RepoData[]> GetImageInfoForSubscriptionAsync(Subscription subscription)
+        {
+            string imageDataJson;
+            using (IGitHubClient gitHubClient = this.gitHubClientFactory.GetClient(Options.GitOptions.ToGitHubAuth()))
+            {
+                GitHubProject project = new GitHubProject(Options.GitOptions.Repo, Options.GitOptions.Owner);
+                GitHubBranch branch = new GitHubBranch(Options.GitOptions.Branch, project);
+
+                GitRepo repo = subscription.RepoInfo;
+                string imageInfoPath = $"{Options.GitOptions.Path}/image-info.{repo.Owner}-{repo.Name}-{repo.Branch}.json";
```
MichaelSimons: It feels strange that the file naming pattern is hard coded here (on read) but it isn't something that is embedding into image-builder on writes (e.g. publishImageInfo).  The publishImageInfo build step is utilizing the git-path option to specify the full path `--git-path build-info/docker/image-info.$(Build.Repository.Name)-$(publicSourceBranch).json`

Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesOptions.cs
```diff
@@ -6,22 +6,21 @@
...
 namespace Microsoft.DotNet.ImageBuilder.Commands
 {
-    public class GetStaleImagesOptions : Options, IFilterableOptions
+    public class GetStaleImagesOptions : Options, IFilterableOptions, IGitOptionsHost
     {
         protected override string CommandHelp => "Gets paths to images whose base images are out-of-date";
 
         public ManifestFilterOptions FilterOptions { get; } = new ManifestFilterOptions();
 
+        public GitOptions GitOptions { get; } = new GitOptions("dotnet", "versions", "master", "build-info/docker");
```
MichaelSimons: Consider refactoring to share the `new GitOptions("dotnet", "versions", "master", "build-info/docker");` logic with the `PublishImageInfoOptions` class as a way to ensure they stay in-sync.

Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
mthalman: This actually adds a lot of complexity to support this because the pipeline logic doesn't have access to the GitHub repo name and owner (it's actually using the internal Azure DevOps repo, which has its own formatting convention).  Things are also complicated by the need for ImageBuilder needing to perform the formatting of the filename because the names are specific to each subscription so it's not something that the pipeline itself knows.  So unfortunately, it's just simpler to keep things the way they are.

MichaelSimons: Per our conversation, I agree.

Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
```diff
@@ -7,12 +7,16 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using Microsoft.DotNet.ImageBuilder.Commands;
```
MichaelSimons: Nit - It's desirable IMO to not add this reference.  From an architectural standpoint, the Command types should not leak outside of that namespace.  This could easily be solved by making this a static member of the GitOptions class.

PR #320, opened 2019-10-03
Update Image Builder tags

mthalman:


MichaelSimons:
Is this to uptake the changes for consuming repo specific image-info (https://github.com/dotnet/docker-tools/issues/281)?  If so, should this PR be updated to also stop the production of the singe image-info file (https://github.com/dotnet/docker-tools/blob/master/eng/common/templates/jobs/publish.yml#L56)?

mthalman:
> Is this to uptake the changes for consuming repo specific image-info (#281)? If so, should this PR be updated to also stop the production of the singe image-info file (https://github.com/dotnet/docker-tools/blob/master/eng/common/templates/jobs/publish.yml#L56)?

Ah yes, thanks for the reminder.

eng/pipelines/templates/steps/get-stale-images.yml
```diff
@@ -4,18 +4,16 @@ parameters:
...
   - template: ${{ format('../../../common/templates/steps/init-docker-{0}.yml', parameters.dockerClientOS) }}
     parameters:
       setupRemoteDockerServer: ${{ parameters.useRemoteDockerServer}}
   - script: >
       $(runImageBuilderCmd)
       getStaleImages
+      $(dotnetBot-userName)
+      $(dotnetBot-email)
+      $(dotnet-bot-user-repo-adminrepohook-pat)
       ${{ parameters.staleImagePathsVariableName }}
```
MichaelSimons: Is the git-path being specified here?  Doesn't it need to be set to something like `build-info/docker`?

mthalman: It's using the default values configured by the command.  But I guess I should be explicit.

PR #321, opened 2019-10-10
Pipeline to push common updates to other repos

mthalman:
This defines a build pipeline which will run the FilePusher utility that pushes any updates to the eng/common to the subscribing repos.  The build gets automatically triggered whenever a change is made within the eng/common folder.  I'm proposing that we have just one build definition that references this pipeline and, by default, it will use an empty `$(filters)` variable.  If there's ever an issue with one of the subscribing repos that requires temporarily disabling syncing, we can just modify the variable appropriately.

### DIFF
--- a/eng/pipelines/push-common-updates.yml
+++ b/eng/pipelines/push-common-updates.yml
@@ -1,0 +1,30 @@
+trigger:
+  batch: true
+  branches:
+    include:
+    - master
+  paths:
+    include:
+    - eng/common/*
+pr: none
+
+variables:
+- group: DotNet-Docker-Common
+- group: DotNet-Docker-Secrets
+- group: DotNet-Maestro
+
+jobs:
+- job: Build
+  pool: Hosted Ubuntu 1604
+  steps:
+  - script: >
+      docker build . -f ./eng/file-pusher/Dockerfile -t file-pusher
+    displayName: Build File Pusher
+  - script: >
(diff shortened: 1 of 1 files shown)


eng/pipelines/push-common-updates.yml
```diff
@@ -0,0 +1,30 @@
...
+  pool: Hosted Ubuntu 1604
+  steps:
+  - script: >
+      docker build . -f ./eng/file-pusher/Dockerfile -t file-pusher
+    displayName: Build File Pusher
+  - script: >
+      docker run --rm file-pusher
+      $(filters)
+      ./eng/eng-common-file-pusher-config.json
+      $(dotnetBot-userName)
```
MichaelSimons: This doesn't look like the correct username and password for the maestro user.  I think it should be dotnet-maestro-bot and dotnet-maestro-bot@microsoft.com

eng/pipelines/push-common-updates.yml
```diff
@@ -0,0 +1,30 @@
...
+  branches:
+    include:
+    - master
+  paths:
+    include:
+    - eng/common/*
+pr: none
+
+variables:
+- group: DotNet-Docker-Common
```
MichaelSimons: Is there a need for these first two groups?

mthalman: Nope, removed

PR #324, opened 2019-10-22
Allow multiple pipeline stages to be explicitly defined to run

### LINKED ISSUE #295: Allow multiple pipeline stages to be explicitly defined to run
Currently the pipeline template allows either all stages to be run or a single stage to be run.  This is based off the `singlePhase` parameter.  If left blank, it runs everything; if it's set, it runs the specified stage.  The parameter should be changed to take a list of stages that can be executed rather than just a single one.  This would be useful for .NET Core releases where we pre-build the images prior to tic-toc but then need to test and publish; it would be good to run test and publish in one build.

mthalman:
Currently there is a queue-time variable named `singlePhase` that can be used to run specific pipeline stages.  When left empty, it runs all the stages.  The problem is that there's not a way to run more than one stage other than all of them (e.g. run just the test _and_ publish phases).

These changes replace the `singleStage` variable with a new one called `stages` that can used to list which stages the pipeline should run.  By default, the pipelines will predefine this variable with all of the stages listed: "build;test;publish".  At queue time, a person can modify this to specify which stages should be run.

Fixes #295

### DIFF
--- a/eng/common/templates/stages/build-test-publish-repo.yml
+++ b/eng/common/templates/stages/build-test-publish-repo.yml
@@ -20,7 +20,7 @@ stages:
 # Build Images
 ################################################################################
 - stage: Build
-  condition: and(succeeded(), or(eq(variables['singlePhase'], ''), eq(variables['singlePhase'], 'build')))
+  condition: and(succeeded(), contains(variables['stages'], 'build'))
   jobs:
   - template: ../jobs/generate-matrix.yml
     parameters:
@@ -186,7 +186,7 @@ stages:
   ################################################################################
   - stage: Post_Build
     dependsOn: Build
-    condition: and(succeeded(), or(eq(variables['singlePhase'], ''), eq(variables['singlePhase'], 'build')))
+    condition: and(succeeded(), contains(variables['stages'], 'build'))
     jobs:
     - template: ../jobs/post-build.yml
 
@@ -197,12 +197,12 @@ stages:
     dependsOn: Post_Build
     condition: "
       and(
-        not(contains(variables['manifest'], 'samples')),
+        contains(variables['stages'], 'test'),
         or(
           and(
             succeeded(),
-            eq(variables['singlePhase'], '')),
-          eq(variables['singlePhase'], 'test')))"
+            contains(variables['stages'], 'build')),
+          not(contains(variables['stages'], 'build'))))"
     jobs:
     - template: ../jobs/generate-matrix.yml
       parameters:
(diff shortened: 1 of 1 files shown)


eng/common/templates/stages/build-test-publish-repo.yml
```diff
@@ -320,15 +320,20 @@ stages:
   - stage: Publish
     dependsOn: Test
     condition: "
-      or(
-        and(
-          eq(variables['singlePhase'], ''),
+      and(
```
MichaelSimons: Am I reading/thinking about this correctly?  If `stages=build;test;publish` and the build passes but the tests fail, doesn't this condition allow publish to proceed?

```
    condition: "
      and( --> TRUE
        contains(variables['stages'], 'publish'), --> TRUE
        or(--> TRUE
          or( --> TRUE
            and( --> TRUE
              contains(variables['stages'], 'build'),
              succeeded('build')),
            and( -- FALSE
              contains(variables['stages'], 'test'),
              succeeded('test'))),
          not( --> FALSE
            or(
              contains(variables['stages'], 'build'),
              contains(variables['stages'], 'test')))))"

**Comment by mthalman** [comment]
> One behavior this allows which I am not crazy about is it allows for build;publish for any repo. Granted this could have been achieved before with multiple builds, this just makes it easy.

I'm not too concerned by that.  As you said, it was possible to do before and it's a pretty explicit act for someone to make to cause this to happen.  In my mind, the danger is the same as it's always been.

File: eng/common/templates/stages/build-test-publish-repo.yml (hunk 1 above)
Comment (MichaelSimons) [reply eng/common/templates/stages/build-test-publish-repo.yml]: Does this handle publish only?  This complexity is why I shied away from supporting this scenario.

Comment (mthalman) [reply eng/common/templates/stages/build-test-publish-repo.yml]: Yes, this is just for publish.  It's the complex one because it needs to be able to run after either build or test.  Test stage is simple(r) because it only cares about build.  If build fails, then test won't run.

Comment (mthalman) [reply eng/common/templates/stages/build-test-publish-repo.yml]: My pseudo code is a replacement for just this portion of the expression tree:
          or(
            and(
              contains(variables['stages'], 'build'),
              succeeded('build')),
            and(
              contains(variables['stages'], 'test'),
              succeeded('test'))),

The rest of the conditions are the same and handle the publish-only scenario.  Pushing an update shortly.

Comment (MichaelSimons) [reply eng/common/templates/stages/build-test-publish-repo.yml]: Is the following equivalent?  It seems simpler.  Not sure I have the parenthesis alignment but hopefully you will get the idea.

```
      and(
        contains(variables['stages'], 'publish'),
        and(
          or(
            and(
              contains(variables['stages'], 'build'),
              succeeded('Post_Build')),
            not(contains(variables['stages'], 'build'))),
          or(
            and(
              contains(variables['stages'], test
              succeeded('Test')),
            not(contains(variables['stages'], 'test')))))
```

File: eng/common/templates/stages/build-test-publish-repo.yml [hunk 2]
```diff
@@ -320,15 +320,32 @@ stages:
...
+        or(
           or(
-            succeeded(),
             and(
-              contains(variables['manifest'], 'samples'),
-              succeeded('Build')))),
-        eq(variables['singlePhase'], 'publish'))"
+              and(
+                contains(variables['stages'], 'build'),
+                succeeded('Post_Build')),
```
MichaelSimons: As I was reading this, I think it would be good to use a consistent ordering of the `contains` and `succeeded` checks.  This was an inconsistency that existed previously but never the less would be good to fix.  Here `contains` is first but in the build and test stages, `succeeded` is first.

eng/common/templates/stages/build-test-publish-repo.yml
mthalman: Yes, that will do nicely.

Issue #79, opened 2018-01-19
Provide detailed tag info


ravimeda:
## Steps to reproduce the issue
1. I would like to try out the .NET Core SDK Docker image. So, I visit https://hub.docker.com/r/microsoft/dotnet/
2. I try to get image details such as the Dockerfile that was used to assemble the image, the version of .NET Core SDK and Shared Framework available on the image.

## Expected behavior

Detailed tag information is available.  Information should include -

- Dockerfile. Not just a link to latest version. But the specific version (commit) that was used to build the image
- Platforms
- Layers 
- Date-time when the image was created
- Docker version that was used to assemble the image

Similar to https://github.com/docker-library/repo-info/blob/master/repos/golang/remote/windowsservercore-1709.md

## Actual behavior

Only a link to the latest Dockerfile is posted. For example,  https://github.com/dotnet/dotnet-docker/blob/master/2.0/sdk/nanoserver-1709/amd64/Dockerfile

ravimeda:
Updated title and description to more accurately describe the issue.

mthalman:
This is the tooling issue associated with https://github.com/dotnet/dotnet-docker/issues/212, which had been [closed](https://github.com/dotnet/dotnet-docker/issues/212#issuecomment-1371334509). The same applies to this issue that we have no plans to implement this.

PR #325, opened 2019-10-23
Include digest of built image in image info file

mthalman:
In order to support #79, we need to be tracking the digest values of the images that get published.  These changes update the Build command to include the digest value in the image info file it outputs.  

Since the digest value isn't known until the image is pushed, the ordering of the logic had to be changed.  Previously, the image info file was written immediately after the images had been built.  Now we need to wait until after the images have been pushed so that we can retrieve the digest values and update the image info model appropriately.

### DIFF
--- a/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -67,6 +67,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 {
                     Repo = repoInfo.Model.Name
                 };
+                reposList.Add(repoData);
 
                 SortedDictionary<string, ImageData> images = new SortedDictionary<string, ImageData>();
 
@@ -130,7 +131,6 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 if (images.Any())
                 {
                     repoData.Images = images;
-                    reposList.Add(repoData);
                 }
             }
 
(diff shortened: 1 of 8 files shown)


MichaelSimons:
Have you validated the digest remains the same as the images get ingested into MCR?  My understanding is a repo digest is repo specific.  Technically the staging repo is a different repo than the MCR repo.  Even if it the same today, is it something that will be guaranteed?  Official Docker repos collect the digest but they do it via pulling published images.  What they collect from the image is the ID.  I feel like this might be a better approach for us.  e.g. we should just collect the ID from our build process. 

Do these changes handle the multi-platform tags?

mthalman:
> Have you validated the digest remains the same as the images get ingested into MCR? My understanding is a repo digest is repo specific. Technically the staging repo is a different repo than the MCR repo. Even if it the same today, is it something that will be guaranteed? Official Docker repos collect the digest but they do it via pulling published images. What they collect from the image is the ID. I feel like this might be a better approach for us. e.g. we should just collect the ID from our build process.
> 
> Do these changes handle the multi-platform tags?

Yes, the digest is the same between registries.  I can investigate how things would work to defer the retrieval until later in the pipeline.

mthalman:
> Do these changes handle the multi-platform tags?

Yes, because the build is platform-specific.  It's probably actually simpler to handle it this way than deferring it to later in the pipeline because of that fact.

MichaelSimons:
> Do these changes handle the multi-platform tags?
> 
> Yes, because the build is platform-specific. It's probably actually simpler to handle it this way than deferring it to later in the pipeline because of that fact.

I'm not seeing how this is done, what am I missing?  The multi-platform tags that I'm referring to are the ones produced by the PublishManifestCommand.

MichaelSimons:
Closing until we settle on an overall plan for tag details.

PR #328, opened 2019-11-08
Include Git commit hash in Dockerfile links

mthalman:
Updates `McrTagsMetadataGenerator` to optionally output a Dockerfile link with a Git commit hash if no source branch is specified.  This allows for different behavior depending on whether the generator is being called by `PublishMcrDocsCommand` (for Docker Hub readme changes) or `GenerateTagsReadmeCommand` (for GitHub readme changes).  For the readmes on Docker Hub, the Dockerfile link should include the Git commit hash.  But for the readmes on GitHub, it should include either the `master` or `nightly` branch, not a Git commit.

Fixes https://github.com/dotnet/dotnet-docker/issues/1436


Microsoft.DotNet.ImageBuilder/src/Microsoft.DotNet.ImageBuilder.csproj
```diff
@@ -15,5 +15,6 @@
     <PackageReference Include="Newtonsoft.Json.Schema" Version="3.0.11" />
     <PackageReference Include="System.CommandLine" Version="0.1.0-e170603-2" />
     <PackageReference Include="System.ComponentModel.Composition" Version="4.5.0" />
+    <PackageReference Include="YamlDotNet" Version="8.0.0" />
```
MichaelSimons: It feels odd to me that we are taking a dependency on two different json libraries.  Can you log an issue to consolidate on one?

Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs
```diff
@@ -0,0 +1,244 @@
...
+                    string missingTags = string.Join(
+                        Environment.NewLine, _imageDocInfos.Select(info => info.FormattedDocumentedTags));
+                    throw new InvalidOperationException(
+                        $"The following tags are not included in the tags metadata: {Environment.NewLine}{missingTags}");
+                }
+
+                string metadata = yaml.ToString();
+
+                // Validate that the YAML is in a valid format
+                new DeserializerBuilder()
```
MichaelSimons: This caught me by surprise initially.  It makes me wonder if there isn't a better pattern to use here that would avoid the cost of serializing and then deserializing.  I realize the templates are what throw a wrench into the scenario.

Microsoft.DotNet.ImageBuilder/src/IMcrTagsMetadataGenerator.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.ImageBuilder.ViewModel;
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    public interface IMcrTagsMetadataGenerator
```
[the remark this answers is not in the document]
mthalman: It had sort of evolved to this point during my implementation.  There is nothing that requires the generator to be defined as a MEF component at this point in time with these changes.  Although, it will be desirable should we ever add unit tests for `GenerateTagsReadmeCommand` or `PublishMcrDocsCommand`.  I can revert this if you'd prefer.

If we were to proceed with this as a MEF component, I don't see the nested class as being unappealing. I actually see it better than the alternative which would be something like defining a factory class to create new instances of `McrTagsMetadataGenerator`.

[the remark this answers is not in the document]
MichaelSimons: My preference is to make the reasonable set of changes to address the problem at hand.  To me this change is not related and since there is no usage it should not be made at this time.  It definitely makes the changes more difficult to review, for example, McrTagsMetadataGenerator isn't detected as a move and therefore it isn't obvious to the reader what was changed.

Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs
mthalman: I thought it for a while when implementing this and, yeah, the templates are what make it rather unwieldy.  This was the lowest cost solution I could think of.  The only reason I added this was because I wanted to have model classes which the tests could make use of to easily grab the values for validation.  And since I added the model classes, I figured we'd might as well have a simple validation check that uses them.  Any other solution is going to be more involved and get outside the scope of these changes.

MichaelSimons: I agree and am fine with it.

Microsoft.DotNet.ImageBuilder/src/Microsoft.DotNet.ImageBuilder.csproj
mthalman: It was as easy as just removing `Newtonsoft.Json` reference since it's a dependency of `Newtonsoft.Json.Schema`.

PR #335, opened 2019-11-20
Consolidate Windows-based job pools in pipeline

### LINKED ISSUE #278: Combine Nano Server and Server Core job types into one set
Currently the pipeline has separate jobs defined depending on whether the matrix specifies "NanoServer" or "WindowsServerCore" as the OS type.  This leads to a lot of duplication since the agent pool requirements are the same between those two OS types.  The matrix output should be updated to specify one common value for a given server version, regardless of whether it's for Nano Serer or Server Core.  This will simplify the job definitions in the pipeline template.

mthalman:
There is essentially duplication in the pipeline in the build and test stages for the Windows jobs because there are sections for handling both a NanoServer-labeled matrix and ServerCore-labeled matrix.  Both of these use the same type of agent pool for their respective Windows version.  These are being consolidate with these changes.  For 1809 amd64, this will now be labeled as ltsc2019.

What this looks like from a consuming perspective is that the manifest must change from using `nanoserver` or `windowsservercore` to just `windows`.

Example (nanoserver):

* Old: `"osVersion": "nanoserver-1909"`
* New `"osVersion": "windows-1909"`

Example (servercore using 1809):

* Old: `"osVersion": "windowsservercore-1809"`
* New `"osVersion": "windows-ltsc2019"`

Fixes #278

### DIFF
--- a/eng/common/templates/stages/build-test-publish-repo.yml
+++ b/eng/common/templates/stages/build-test-publish-repo.yml
@@ -33,7 +33,7 @@ stages:
       name: Build_Linux_amd64
       pool: # linuxAmd64Pool
         name: Hosted Ubuntu 1604
-      matrix: dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixLinuxAmd64', parameters.buildMatrixType) }}']
+      matrix: dependencies.GenerateBuildMatrix.outputs['matrix.LinuxAmd64']
       dockerClientOS: linux
       buildJobTimeout: ${{ parameters.linuxAmdBuildJobTimeout }}
       customInitSteps: ${{ parameters.customBuildInitSteps }}
@@ -49,7 +49,7 @@ stages:
         - agent.os -equals linux
         - RemoteDockerServerOS -equals linux
         - RemoteDockerServerArch -equals aarch64
-      matrix: dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixLinuxArm64v8', parameters.buildMatrixType) }}']
+      matrix: dependencies.GenerateBuildMatrix.outputs['matrix.LinuxArm64v8']
       useRemoteDockerServer: true
       dockerClientOS: linux
       buildJobTimeout: ${{ parameters.linuxArmBuildJobTimeout }}
@@ -66,10 +66,23 @@ stages:
         - agent.os -equals linux
         - RemoteDockerServerOS -equals linux
         - RemoteDockerServerArch -equals aarch64
-      matrix: dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixLinuxArm32v7', parameters.buildMatrixType) }}']
+      matrix: dependencies.GenerateBuildMatrix.outputs['matrix.LinuxArm32v7']
       useRemoteDockerServer: true
       dockerClientOS: linux
       buildJobTimeout: ${{ parameters.linuxArmBuildJobTimeout }}
+      customInitSteps: ${{ parameters.customBuildInitSteps }}
+  - template: ../jobs/build-images.yml
+    parameters:
+      name: Build_NanoServer1809_amd64
+      pool: # windows1809Amd64
+        ${{ if eq(variables['System.TeamProject'], 'public') }}:
+          name: DotNetCore-Docker-Public
+        ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+          name: DotNetCore-Docker
+        demands: VSTS_OS -equals Windows_Server_2019_Data_Center_RS5
+      matrix: dependencies.GenerateBuildMatrix.outputs['matrix.Windows1809Amd64']
+      dockerClientOS: windows
+      buildJobTimeout: ${{ parameters.windowsAmdBuildJobTimeout }}
       customInitSteps: ${{ parameters.customBuildInitSteps }}
   - template: ../jobs/build-images.yml
     parameters:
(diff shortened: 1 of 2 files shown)


MichaelSimons:
`Windows-1909` feels redundant given the `os` property is already set to Windows.  I wonder if we will ever need to distinguish between nano and servercore images.  For example if we ever change the tag details rendering from Windows Server to Nano Server/Windows Server Core.  Have you considered leaving the manifest as is and changing the `GenerateBuildMatrixCommand` to transform the OsVersion segment for Windows as you have suggested when constructing the matrix name?

mthalman:
> Have you considered leaving the manifest as is and changing the `GenerateBuildMatrixCommand` to transform the OsVersion segment for Windows as you have suggested when constructing the matrix name?

That was my first approach but then I didn't want Image Builder to have such a dependency on the name here.

MichaelSimons:
The more I think about this the more it does not feel correct.  If I understand the proposal correctly, `osVersion` would get set to `windows-ltsc2019` for a `nanoserver 1809` based image.  This feels wrong because Nano Server is not an LTSC.   It feels like this change is changing the meaning of `osVersion` to be the OS needed to build the image.

Have you considered leaving the manifest as is and utilizing conditionals in the pipeline?

```
        ${{ if ne(dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixNanoserver1809Arm32', parameters.buildMatrixType) }}'], '') }}:
                matrix: dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixNanoserver1809Arm32', parameters.buildMatrixType) }}']
        ${{ if ne(dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixServercoreltsc2019Arm32', parameters.buildMatrixType) }}'], '') }}:
                matrix: dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixServercoreltsc2019Arm32', parameters.buildMatrixType) }}']
```

The downside to this is it wouldn't support building both servercore and nanoserver images within the same manifest but I don't recall a need for this as of now.

mthalman:
> It feels like this change is changing the meaning of osVersion to be the OS needed to build the image.

Is it changing the meaning though? That's essentially what `osVersion` is.  That's all this value is used for, in addition to filtering.

mthalman:
> > It feels like this change is changing the meaning of osVersion to be the OS needed to build the image.
> 
> Is it changing the meaning though? That's essentially what `osVersion` is. That's all this value is used for, in addition to filtering.

Ah, I guess it is also used for generating the readme tags.

MichaelSimons:
> Ah, I guess it is also used for generating the readme tags.

Yes.

> Is it changing the meaning though? That's essentially what osVersion is. That's all this value is used for, in addition to filtering

Filtering is pretty important especially for the local dev experience.  --osVersion windows-ltsc2016 within the context of dotnet-docker is going to be confusing and not intuitive.

mthalman:
> Have you considered leaving the manifest as is and utilizing conditionals in the pipeline?
> 
> ```
>         ${{ if ne(dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixNanoserver1809Arm32', parameters.buildMatrixType) }}'], '') }}:
>                 matrix: dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixNanoserver1809Arm32', parameters.buildMatrixType) }}']
>         ${{ if ne(dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixServercoreltsc2019Arm32', parameters.buildMatrixType) }}'], '') }}:
>                 matrix: dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixSer
> ```

Pretty sure that's not possible to do because when ${{ ... }} is evaluated, it is done when processing the pipeline and before it actually has executed (i.e. it's a preprocessor).  So accessing `dependencies.GenerateBuildMatrix` wouldn't work since that's not set until the job has been run.

mthalman:
I see two options:
1. Update Image Builder to parse out `osVersion` to produce a generically named matrix value as you had said.
1. Define another value in the manifest model named something like `osBuildVersion` to indicate which version of the OS is used to build the image.

I prefer the first option.

MichaelSimons:
I agree, option 1 seems like the preferred solution.

eng/common/templates/stages/build-test-publish-repo.yml
```diff
@@ -96,86 +96,60 @@ stages:
...
         demands: VSTS_OS -equals Windows_Server_2016_Data_Center_with_Containers
-      matrix: dependencies.GenerateBuildMatrix.outputs['${{ format('matrix.{0}MatrixWindowsservercoreLtsc2016Amd64', parameters.buildMatrixType) }}']
+      matrix: dependencies.GenerateBuildMatrix.outputs['matrix.WindowsLtsc2016Amd64']
       dockerClientOS: windows
       buildJobTimeout: ${{ parameters.windowsAmdBuildJobTimeout }}
       customInitSteps: ${{ parameters.customBuildInitSteps }}
   - template: ../jobs/build-images.yml
     parameters:
-      name: Build_WindowsServerCoreLtsc2019_amd64
+      name: Build_WindowsLtsc2019_amd64
```
MichaelSimons: Why not include mapping info in the build matrix logic for ltsc versions?  It would consolidate the duplicated build jobs.  The build definition needs to be updated when new windows versions come out anyway so it feels like updating image-builder is not much more work.  The fix only partially fixes https://github.com/dotnet/docker-tools/issues/278,  what is the value in that versus leaving it as is?

mthalman: It was due to the existence of the 1809 ARM leg which would not want to be translated to ltsc2019.  In order to avoid that, it would have required the Image Builder code to specifically handle that edge case which I wanted to avoid.  To workaround this, it's now using 1809 as the job label instead of ltsc2019.  So any ltsc2019 references are being translated to 1809.

Issue #305, opened 2019-09-16
Define a pipeline that runs a fake build through the common pipeline infra


mthalman:
The docker-tools is the main repo for the common pipeline infra but there's no verification of that infra as part of the PR validation for the repo.  It would be nice to run a build with a fake manifest and Dockerfile(s) to verify the common pipeline infrastructure.

PR #379, opened 2020-01-17
Support non-root folder for manifest file

### LINKED ISSUE #370: ImageBuilder has path issues when manifest is not located at current working directory
Let's say you run the `build` command on ImageBuilder with a current working directory of `.` and a manifest located at `./test/manifest.json`.  The Dockerfile paths specified in the manifest are relative to the manifest, so they could be paths like `3.1/runtime/alpine3.10/amd64`.  When the `build` command runs, it will use the Dockerfile path specified in the manifest and generate a Docker command like `docker build -t tag -f 3.1/runtime/alpine3.10/amd64/Dockerfile 3.1/runtime/alpine3.10/amd64`.  Since the current working directory is `.`, this command will fail because the Dockerfile does not exist at `./3.1/runtime/alpine3.10/amd64`; it exists at `./test/3.1/runtime/alpine3.10/amd64`.

ImageBuilder should be updated across all of its commands to support this scenario and generate Docker commands that reference the correct paths.

Two requirements on this:
* Dockerfile paths specified in the manifest files should be relative to the manifest file.
* When referencing paths with the `--path` filtering option that various commands support, those paths should also be relative to the manifest file.

mthalman:
Updates ImageBuilder logic to support a manifest file being located in a different folder from the current working directory (i.e. the root of the repo in most cases).  This is needed in order to support #305 where we want to make use of a fake manifest file located in a test folder.

All file paths referenced from within the manifest file (e.g. readme files) should be relative to the location of the manifest file.  Similarly, the `--path` filter options that can be passed to various ImageBuilder commands should also be relative to the manifest file.

Fixes #370

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
@@ -96,8 +96,6 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
         private IEnumerable<string> GetDestinationTagNames(RepoInfo repo, PlatformInfo platform)
         {
-            string lookupDockerfilePath = PathHelper.StripBaseDirectory(Manifest.BaseDirectory, platform.DockerfilePath);
-
             IEnumerable<string> destTagNames = null;
 
             // If an image info file was provided, use the tags defined there rather than the manifest. This is intended
@@ -109,7 +107,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 RepoData repoData = imageInfoRepos.Value.FirstOrDefault(repoData => repoData.Repo == repo.Model.Name);
                 if (repoData != null)
                 {
-                    if (repoData.Images.TryGetValue(lookupDockerfilePath, out ImageData image))
+                    if (repoData.Images.TryGetValue(platform.DockerfilePathRelativeToManifest, out ImageData image))
                     {
                         destTagNames = image.SimpleTags
                             .Select(tag => TagInfo.GetFullyQualifiedName(repo.Name, tag));
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
@@ -25,13 +25,14 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     public class CopyAcrImagesCommand : ManifestCommand<CopyAcrImagesOptions>
     {
         private Lazy<RepoData[]> imageInfoRepos;
-        private string baseDirectory;
         private readonly IAzureManagementFactory azureManagementFactory;
+        private readonly IEnvironmentService environmentService;
 
         [ImportingConstructor]
-        public CopyAcrImagesCommand(IAzureManagementFactory azureManagementFactory) : base()
+        public CopyAcrImagesCommand(IAzureManagementFactory azureManagementFactory, IEnvironmentService environmentService) : base()
         {
             this.azureManagementFactory = azureManagementFactory ?? throw new ArgumentNullException(nameof(azureManagementFactory));
+            this.environmentService = environmentService ?? throw new ArgumentNullException(nameof(environmentService));
             this.imageInfoRepos = new Lazy<RepoData[]>(() =>
             {
                 if (!String.IsNullOrEmpty(Options.ImageInfoPath))
@@ -46,8 +47,6 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         public override async Task ExecuteAsync()
         {
             Logger.WriteHeading("COPYING IMAGES");
-
-            this.baseDirectory = Path.GetDirectoryName(Options.Manifest);
 
             string registryName = Manifest.Registry.TrimEnd(".azurecr.io");
 
@@ -97,6 +96,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
         private IEnumerable<string> GetDestinationTagNames(RepoInfo repo, PlatformInfo platform)
         {
+            string lookupDockerfilePath = PathHelper.StripBaseDirectory(Manifest.BaseDirectory, platform.DockerfilePath);
+
             IEnumerable<string> destTagNames = null;
 
             // If an image info file was provided, use the tags defined there rather than the manifest. This is intended
@@ -108,7 +109,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 RepoData repoData = imageInfoRepos.Value.FirstOrDefault(repoData => repoData.Repo == repo.Model.Name);
                 if (repoData != null)
                 {
-                    if (repoData.Images.TryGetValue(platform.DockerfilePath, out ImageData image))
+                    if (repoData.Images.TryGetValue(lookupDockerfilePath, out ImageData image))
                     {
                         destTagNames = image.SimpleTags
                             .Select(tag => TagInfo.GetFullyQualifiedName(repo.Name, tag));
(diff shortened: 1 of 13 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
```diff
@@ -97,6 +96,8 @@ private async Task ImportImageAsync(string destTagName, string registryName)
 
         private IEnumerable<string> GetDestinationTagNames(RepoInfo repo, PlatformInfo platform)
         {
+            string lookupDockerfilePath = PathHelper.StripBaseDirectory(Manifest.BaseDirectory, platform.DockerfilePath);
```
MichaelSimons: There are enough places that perform this same logic that it feels beneficial to create a property or method on PlatformInfo for it.  The naming between this proposed property/method and the existing DockerfilePath is tricky.

I'm not in love with any of these but it is what come to mind at this moment:
`QualifiedDockerfilePath` / `RelativeDockerfilePath`
`ManifestQualifiedDockerfilePath` / `ManifestRelativeDockerfilePath`
`DockerfilePathWithManifestPath` / `DockerfilePath`

mthalman: How about `DockerfilePathRelativeToManifest`?  It's a little bit verbose but it's only referenced 5 times and the descriptiveness is necessary.

MichaelSimons: I agree the descriptiveness is necessary.  I like your suggestion.

mthalman:
> Should we add validation logic to ensure all paths specified in the manifest are not rooted? Am I correct in understanding that with this change the readmePath and mcrTagsMetadataTemplatePath must be relative but the Dockerfiles can be rooted? I think it would be best for everything to be relative in reference to the manifest and we should validate it.

This makes sense.  Support for absolute paths are really only needed for testing purposes and the `GetStaleImages` command, but only for manifest files.  Once you have the path to the manifest file, everything should be relative from that; there's not a need for the manifest to reference an absolute path.  I'll update the code to validate this.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -42,6 +43,21 @@ public static void Validate(this Manifest manifest)
             foreach (Repo repo in manifest.Repos)
             {
                 ValidateUniqueTags(repo);
+
+                if (Path.IsPathRooted(repo.ReadmePath))
```
MichaelSimons: Thoughts on having this validation logic also call `File.Exists`?  It seems like it would yield a better user experience.

src/Microsoft.DotNet.ImageBuilder/src/PathHelper.cs
```diff
@@ -2,10 +2,35 @@
...
     public static class PathHelper
     {
         public static string NormalizePath(string path) => path.Replace(@"\", "/");
+
+        /// <summary>
+        /// Trims the <paramref name="trimPath"/> string from <paramref name="path"/>.
+        /// </summary>
+        /// <param name="trimPath">The path segment to remove from <paramref name="path"/>.</param>
+        /// <param name="path">The path to be trimmed.</param>
+        public static string TrimPath(string trimPath, string path)
```
MichaelSimons: Did you consider making this an extension method to simplify the calling code?

mthalman: I don't like extension methods for primitive types unless they're relevant to any kind of value for that type. This is path-specific, so it doesn't fit with that.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -37,11 +38,40 @@ public static string GetDisplayName(this Architecture architecture, string varia
...
+
+                if (Path.IsPathRooted(repo.McrTagsMetadataTemplatePath))
+                {
+                    throw new ValidationException($"Tags template path '{repo.McrTagsMetadataTemplatePath}' for repo {repo.Name} must be a relative path.");
+                }
+
+                EnsureFileExists(repo.McrTagsMetadataTemplatePath, manifestDirectory);
+            }
+
+            if (Path.IsPathRooted(manifest.ReadmePath))
```
MichaelSimons: Thoughts on creating a single helper method that validates the path completely?

```
        private static void ValidatePath(string path, string manifestDirectory)
        {
            if (path != null)
            {
                if (Path.IsPathRooted(path))
                {
                    throw new ValidationException($"All paths specified in the manifest file must be relative paths.  Invalid Path: `{path}`");
                }
                if (!File.Exists(Path.Combine(manifestDirectory, path)))
                {
                    throw new FileNotFoundException("Path specified in manifest file does not exist.", path);
                }
            }
        }
```

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -44,11 +45,13 @@ private PlatformInfo(string baseDirectory)
...
-            platformInfo.VariableHelper = variableHelper;
+            PlatformInfo platformInfo = new PlatformInfo(baseDirectory)
+            {
+                Model = model,
+                RepoName = repoName,
+                FullRepoModelName = fullRepoModelName,
+                VariableHelper = variableHelper
+            };
 
             // Ensure that we construct a path using the base directory here to check for the file. In cases like
```
MichaelSimons: The IsPathRooted check should be removed here in order to treat all manifest paths consistently.  It also feels like the Dockerfile paths should be validated with the path validation logic you added.  This is a little more difficult because of the filename defaulting logic.  That logic could be factored into a common helper.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -53,24 +53,22 @@ public static PlatformInfo Create(Platform model, string fullRepoModelName, stri
...
 
-            if (File.Exists(dockerfileWithBaseDir))
+            try
             {
-                platformInfo.DockerfilePath = PathHelper.NormalizePath(dockerfileWithBaseDir);
-                platformInfo.BuildContextPath = Path.GetDirectoryName(dockerfileWithBaseDir);
+                PathHelper.ValidateFileReference(dockerfilePath, baseDirectory);
             }
-            else
+            catch (FileNotFoundException)
```
MichaelSimons: I'm not crazy about the try catch usage here.  The typical usage in our repos is to specify the directory without the implied Dockerfile.  Loading the dotnet-docker manifest would cause 130 exceptions to be raised.  This is going to have an impact on performance.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -53,24 +53,22 @@ public static PlatformInfo Create(Platform model, string fullRepoModelName, stri
...
-            string dockerfileWithBaseDir = Path.IsPathRooted(model.Dockerfile) ?
-                model.Dockerfile : Path.Combine(baseDirectory, model.Dockerfile);
+            string dockerfilePath = model.Dockerfile;
 
-            if (File.Exists(dockerfileWithBaseDir))
+            try
             {
-                platformInfo.DockerfilePath = PathHelper.NormalizePath(dockerfileWithBaseDir);
-                platformInfo.BuildContextPath = Path.GetDirectoryName(dockerfileWithBaseDir);
+                PathHelper.ValidateFileReference(dockerfilePath, baseDirectory);
```
MichaelSimons: I have mixed feelings about putting the Dackerfile validation logic here when the other paths are validated in the model validation logic.

src/Microsoft.DotNet.ImageBuilder/src/PathHelper.cs
```diff
@@ -32,5 +35,18 @@ public static string GetNormalizedDirectory(string path)
         {
             return PathHelper.NormalizePath(Path.GetDirectoryName(path));
         }
+
+        public static void ValidateFileReference(string path, string manifestDirectory)
```
MichaelSimons: The method implies this is a generic method but the exception messages ties the logic to manifest file references.

PR #395, opened 2020-01-29
Upgraded projects to .NET Core 3.1

### LINKED ISSUE #375: Update ImageBuilder to .NET Core 3.1
It's currently configured to use .NET Core 3.0:
https://github.com/dotnet/docker-tools/blob/c3e76ad8cd58865e8418330c6dc5c7ed10f6f283/src/Microsoft.DotNet.ImageBuilder/src/Microsoft.DotNet.ImageBuilder.csproj#L6

MichaelSimons:
Fixes #375

### DIFF
--- a/eng/file-pusher/Dockerfile
+++ b/eng/file-pusher/Dockerfile
@@ -1,7 +1,7 @@
 # This Dockerfile is intended to be built at the root of the repo.
 
 # build image
-FROM mcr.microsoft.com/dotnet/core/sdk:3.0-alpine AS build-env
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1-alpine AS build-env
 
 WORKDIR /file-pusher
 
@@ -16,7 +16,7 @@ RUN dotnet publish -c Release -o out --no-restore
 
 
 # runtime image
-FROM mcr.microsoft.com/dotnet/core/runtime:3.0-alpine
+FROM mcr.microsoft.com/dotnet/core/runtime:3.1-alpine
 
 # copy file-pusher
 WORKDIR /file-pusher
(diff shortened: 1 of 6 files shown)


eng/file-pusher/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # This Dockerfile is intended to be built at the root of the repo.
 
 # build image
-FROM mcr.microsoft.com/dotnet/core/sdk:3.0-alpine AS build-env
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1-alpine AS build-env
```
mthalman: Just wondering whether we can have a single point of configuration for this.  Can we add an `ARG` to the Dockerfile for the .NET version that is referenced both for the image tags as well as passed to the build of the project as a build property?

Something like:
```
ARG DOTNET_VERSION=3.1
FROM mcr.microsoft.com/dotnet/core/sdk:$DOTNET_VERSION-alpine AS build-env
...
RUN dotnet publish -c Release -o out --no-restore /p:DotnetVersion=$DOTNET_VERSION
```

```
<PropertyGroup>
  <TargetFramework>netcoreapp$(DotnetVersion)</TargetFramework>
</PropertyGroup>
```

MichaelSimons: I had experimented with this.  The problem I kept running into is that I felt it hurt the developer experience.  e.g. you can't just open the project and build; you can't just build the Dockerfile.  If you default the values in the Dockerfile and project then you can but you have negated the reason for this change as they would need to be updated each time the version revs.  Given this I decided to use the search and replace approach and move on.

mthalman: Makes sense.  Seems like there's an opportunity for a preprocessor tool for Dockerfiles that can allow you to include values from a separate file.  Apparently, direct support for this from Docker has been rejected: https://github.com/moby/moby/issues/735.

PR #399, opened 2020-01-31
Delete image after its size is validated

mthalman:
To avoid requiring a lot of disk space to store all the images that are pulled for image size validation, these changes remove any images that are pulled after each has been validated.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/ImageSizeCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/ImageSizeCommand.cs
@@ -40,16 +40,32 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
         protected long GetImageSize(string tagName)
         {
-            if (Options.IsPullEnabled)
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/ImageSizeCommand.cs
```diff
@@ -40,16 +40,32 @@ protected void ProcessImages(ImageHandler processImage)
...
+                return DockerService.GetImageSize(tagName, Options.IsDryRun);
             }
-            else if (!DockerService.LocalImageExists(tagName, Options.IsDryRun))
+            finally
             {
-                throw new InvalidOperationException($"Image '{tagName}' not found locally");
+                // If we had to pull the image because it didn't exist locally, be sure to clean it up
+                if (!localImageExists)
+                {
+                    string imageId = DockerService.GetImageId(tagName, Options.IsDryRun);
```
MichaelSimons: Why retrieve the image ID?  Why not just delete the image via tagName?

mthalman:
I agree that we should avoid pulling extra data unnecessarily.  I'm abandoning these changes and will pursue a different approach.

PR #400, opened 2020-02-03
Option to skip baseline integrity check for size validation

mthalman:
Adds an option to the `validateImageSize` command to skip the baseline integrity check.  This is needed to support https://github.com/dotnet/dotnet-docker/issues/1588 so that a build job can validate a partial set of images and not run into issues with the baseline integrity check.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
@@ -43,10 +43,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             // This handler will be invoked for each image defined in the manifest
             void processImage(string repoId, string imageId, string tagName)
             {
-                // If the CheckBaselineIntegrityOnly option is enabled, we want to skip the retrieval
-                // of the image size.
                 long? currentSize = null;
-                if (!Options.CheckBaselineIntegrityOnly)
+                if (Options.Mode.HasFlag(ImageSizeValidationMode.Size))
                 {
                     currentSize = GetImageSize(tagName);
                 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeOptions.cs
```diff
@@ -12,6 +12,8 @@ public class ValidateImageSizeOptions : ImageSizeOptions
 
         public bool CheckBaselineIntegrityOnly { get; set; }
 
+        public bool SkipBaselineIntegrityCheck { get; set; }
```
MichaelSimons: The usability of this new option is a little strange in that it completely overrides the `CheckBaselineIntegrityOnly` option.  Have you considered defining a single enum option?  Mode=All, Size, Integrity

mthalman: Yes, that's better.  Fixed

src/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeCommand.cs
```diff
@@ -85,7 +83,7 @@ private ImageSizeValidationResults ValidateImages(IEnumerable<ImageSizeInfo> ima
             IEnumerable<ImageSizeInfo> imagesWithNoSizeChange;
             IEnumerable<ImageSizeInfo> imagesWithAllowedSizeChange;
             IEnumerable<ImageSizeInfo> imagesWithDisallowedSizeChange;
-            if (Options.CheckBaselineIntegrityOnly)
+            if (Options.Mode == ImageSizeValidationMode.Integrity)
```
MichaelSimons: Doesn't this need to use HasFlag?  The same comment applies to the change where the `ImageSizeValidationResults` is constructed.

mthalman: No, we only want to execute the logic when it's specifically limited to `integrity` (or `size` in the other case).  In the case when it's `all`, we do want to compute all the size results.

MichaelSimons: Yes - I understand now.

PR #404, opened 2020-02-05
Fix multi-stage Dockerfile support in getStaleImage and enhance logging

### LINKED ISSUE #205: Fix tracking of base image digests for multi-stage Dockerfiles
Some of our Dockerfiles make use of multi-stage builds and we're tracking the image digests of all the images referenced in the FROM statements.  That means if any of those base images changes, it'll cause a build to be kicked off to rebuild the image.  But we don't actually care about _all_ of those base images.  We only care about the base image of the image _produced_ by the Dockerfile, not the intermediate images that were used during the build process.

We should update Image Builder's Build command to only output the image digest for the "final" base image.

### LINKED ISSUE #343: Add better logging in GetStaleImagesCommand
The GetStaleImagesCommand class in Image Builder could benefit from better logging.  It'd be good to see what digest values are being compared for each iteration.

MichaelSimons:
Fixes #205  
Fixes #343 

* Tweaked getStaleImages to only inspect the final stage of multi-stage Dockerfiles
  * Added unit tests to cover multi-stage Dockerfile scenarios
* Enhanced the logging of getStaleImages

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
@@ -97,6 +97,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 return Enumerable.Empty<string>();
             }
 
+            this.loggerService.WriteMessage($"Processing subscription:  {subscription.Id}");
+
             RepoData[] repos = await GetImageInfoForSubscriptionAsync(subscription);
 
             string repoPath = await GetGitRepoPath(subscription);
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
```diff
@@ -126,38 +128,36 @@ private async Task<IEnumerable<string>> GetPathsToRebuildAsync(Subscription subs
                         repoData.Images != null &&
                         repoData.Images.TryGetValue(platform.DockerfilePathRelativeToManifest, out ImageData imageData))
                     {
-                        bool hasDigestChanged = false;
+                        string fromImage = platform.ExternalFromImages.Last();
+                        string currentDigest;
 
-                        foreach (string fromImage in platform.ExternalFromImages)
+                        await this.imageDigestsSemaphore.WaitAsync();
```
mthalman: I realized that there's probably no benefit to executing this method in parallel because most of the time is spent blocking here when needing to pull an image. Just a comment, not expecting that to be changed here.

MichaelSimons: Yes, I had thought the same when I was making these changes.  It certainly makes the log output difficult to follow.  I decided this was beyond the scope of this change and really doesn't have business value.

PR #408, opened 2020-02-07
Cleanup untagged nightly images

### LINKED ISSUE #378: The cleanup ACR images pipeline should be cleaning up the nightly repos.
Our build ACR is ever increasing in size.  We should have cleanup logic for the public nightly repos.  There is no reason to keep every image we produce.  I think it would be reasonable to delete all untagged images from these repos which should dramatically reduce the size of the repos.

mthalman:
Updates the script to cleanup ACR by including untagged images in the nightly repos.

Fixes #378

### DIFF
--- a/eng/Invoke-AcrCleanup.ps1
+++ b/eng/Invoke-AcrCleanup.ps1
@@ -1,10 +1,28 @@
+#!/usr/bin/env pwsh
+
+<#
+    .SYNOPSIS
+        Cleans up repos and images in ACR that are no longer needed.
+#>
+
 [cmdletbinding(DefaultParameterSetName=$null)]
 param(
+    # Name of the ACR
     [Parameter(Mandatory = $true)][string]$RegistryName,
+
+    # Azure subscription ID
     [Parameter(Mandatory = $true)][string]$SubscriptionId,
+
+    # Shows what would happen if deletions occurred but doesn't actually delete anything.
     [switch]$WhatIf,
+
+    # Name of the ACR service principal
     [string]$ServicePrincipalName,
+
+    # Password of the ACR service principal
     [string]$ServicePrincipalPassword,
+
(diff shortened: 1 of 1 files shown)


eng/Invoke-AcrCleanup.ps1
```diff
@@ -27,22 +45,43 @@ az account set --subscription $SubscriptionId
...
+
+            $deletedRepos += $repoName
         }
+    }
+    else {
+        Write-Host "Querying manifests"
+        $manifests = az acr repository show-manifests --name $RegistryName --repository $repoName | ConvertFrom-Json
+
+        # Delete all of the untagged images
+        $untaggedImages = $manifests | Where-Object { $_.tags.Count -eq 0 }
```
MichaelSimons: Should this also include a date component?  e.g. delete all untagged images older than expiration period used in the staging repos cleanup logic.

eng/Invoke-AcrCleanup.ps1
```diff
@@ -57,5 +96,12 @@ foreach ($deletedRepo in $deletedRepos) {
 }
 
 Write-Host ""
+Write-Host "Deleted images:"
+foreach ($deletedImage in $deletedImages) {
+    Write-Host $deletedImage
```
MichaelSimons: Does this display information about what the parent repo is?  I think this should be displayed.

mthalman: Yes, an example is `public/dotnet/core-nightly/sdk@sha256:fff9f9ae1e00b84421588c37650aa7109e4c95d222007f2c72ad00d8a6409830`

eng/Invoke-AcrCleanup.ps1
mthalman: Good point

PR #410, opened 2020-02-10
Command to display base image status

mthalman:
Defines an ImageBuilder command that will display the "created date" status for external base images referenced by the Dockerfiles.

Related to #402

PR #411, opened 2020-02-11
Common script to get base image status

### LINKED ISSUE #402: Script to determine readiness of Windows images
On Patch Tuesday, we need to get positive confirmation that the base Windows images have been updated prior to producing the .NET Docker images.  The https://github.com/dotnet/dotnet-docker and https://github.com/Microsoft/dotnet-framework-docker repo each have their own set of images that they're dependent.  A common script should be defined that can be used to determine the status of these images and whether they've been updated.  This would save time from needing to manually run these actions and reduce the risk of potentially missing the checking of an image.

mthalman:
Adds a wrapper script for the `getBaseImageStatus` command (see https://github.com/dotnet/docker-tools/pull/410) to allow callers to get the status of base images.

Fixes #402

### DIFF
--- a/eng/common/Get-BaseImageStatus.ps1
+++ b/eng/common/Get-BaseImageStatus.ps1
@@ -16,24 +16,18 @@ param(
 
     # A value indicating whether to run the script continously
     [switch]
-    $Continuous
+    $Continuous,
+
+    # Number of seconds to wait between each iteration
+    [int]
+    $ContinuousDelay = 10
 )
 
 Set-StrictMode -Version Latest
 
-function RunCommand() {
-    $imageBuilderArgs = "getBaseImageStatus --manifest $Manifest --architecture $Architecture"
-    & "$PSScriptRoot/Invoke-ImageBuilder.ps1" -ImageBuilderArgs $imageBuilderArgs
-}
-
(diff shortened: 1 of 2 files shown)


eng/common/Get-BaseImageStatus.ps1
```diff
@@ -0,0 +1,39 @@
...
+
+    # A value indicating whether to run the script continously
+    [switch]
+    $Continuous
+)
+
+Set-StrictMode -Version Latest
+
+function RunCommand() {
+    $imageBuilderArgs = "getBaseImageStatus --manifest $Manifest --architecture $Architecture"
```
MichaelSimons: This has a lot of overhead.  For example it issues a docker pull each time it is invoked.  Have you considered moving the continuous implementation into the image-builder cmd?

mthalman: Well it's going to need to pull each time to get an accurate created date.  Or are you referring to pulling of the ImageBuilder image?

MichaelSimons: Sorry, my comment was not very clear.  I meant the overhead of the `Invoke-ImageBuilder.ps1` script.  It pulls the base image and builds a docker image which includes the repo each time it is run.

PR #422, opened 2020-03-02
Include digest of built image in image info file

mthalman:
There are a variety of scenarios that require the need for tracking the image digests of the images we publish and associating them with their corresponding Dockerfile.  These scenarios include:
* Mapping a digest to telemetry pivot data
* Support for #79 
* Identifying the tag associated with a digest for troubleshooting purposes

These changes update the Build command to include the digest value in the image info file it outputs.

Since the digest value isn't known until the image is pushed, the ordering of the logic had to be changed. Previously, the image info file was written immediately after the images had been built. Now we need to wait until after the images have been pushed so that we can retrieve the digest values and update the image info model appropriately.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -58,13 +58,9 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 return;
             }
 
-            var pushedImages = this.repoList
-                .Where(repoData => repoData.Images != null)
-                .SelectMany(repoData => repoData.Images.Values);
-
-            foreach (var image in pushedImages)
+            foreach (var image in GetBuiltImages())
             {
-                foreach (string tag in image.AllTags)
+                foreach (string tag in image.FullyQualifiedSimpleTags)
                 {
                     // The digest of an image that is pushed to ACR is guaranteed to be the same when transferred to MCR
                     // It is output in the form of <repo>@<sha> but we only want the sha.
@@ -155,6 +151,9 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                         imageData.SimpleTags = GetPushTags(platform.Tags)
                             .Select(tag => tag.Name)
                             .OrderBy(name => name)
+                            .ToList();
+                        imageData.FullyQualifiedSimpleTags = imageData.SimpleTags
+                            .Select(tag => TagInfo.GetFullyQualifiedName(repoInfo.Name, tag))
                             .ToList();
                         imageData.AllTags = allTags;
                     }
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -23,15 +23,16 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     public class BuildCommand : DockerRegistryCommand<BuildOptions>
     {
         private readonly IDockerService dockerService;
+        private readonly ILoggerService loggerService;
+        private readonly IEnvironmentService environmentService;
         private readonly List<RepoData> repoList = new List<RepoData>();
 
-        private IEnumerable<TagInfo> BuiltTags { get; set; } = Enumerable.Empty<TagInfo>();
-        
-
         [ImportingConstructor]
-        public BuildCommand(IDockerService dockerService)
+        public BuildCommand(IDockerService dockerService, ILoggerService loggerService, IEnvironmentService environmentService)
         {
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -23,8 +23,10 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     public class BuildCommand : DockerRegistryCommand<BuildOptions>
     {
         private readonly IDockerService dockerService;
+        private readonly List<RepoData> reposList = new List<RepoData>();
```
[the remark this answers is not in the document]
mthalman: `BuiltTags` is actually a superset of the tags contained in `reposList` because `reposList` only contains simple tags.  Once we support persisting _all_ tags in the the image info file, then we can get rid of `BuiltTags`.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -270,11 +273,34 @@ private void PushImages()
...
+
+                    foreach (var image in imagesToPush)
                     {
-                        this.dockerService.PushImage(tag, Options.IsDryRun);
+                        foreach (string tag in image.FullyQualifiedSimpleTags)
+                        {
+                            
+                            this.dockerService.PushImage(tag, Options.IsDryRun);
+
+                            // Only bother getting the image digest if we're going to output the image info
```
MichaelSimons: I like this implementation but I would prefer to see it encapsulated within a PublishImageInfo method along side the logic that writes the info to a file.  The advantage is centralizing more of the ImageInfo logic within as few of places as possible.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -23,41 +23,73 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
...
 
-            WriteBuildSummary();
+            var pushedImages = this.repoList
+                .Where(repoData => repoData.Images != null)
+                .SelectMany(repoData => repoData.Images.Values);
 
-            return Task.CompletedTask;
+            foreach (var image in pushedImages)
+            {
+                foreach (string tag in image.AllTags)
```
MichaelSimons: This doesn't look right to me as the way to handle the shared tags.  When building on a particular platform, ImageBuilder will tag the concrete images with the shared tags for convenience purposes only in-case your want to use these tags locally or in testing.  They are not the actual multi-platform tags.  Those don't get produced by the `BuildCommand` rather that is the responsibility of the `PublishManifestCommand`.

mthalman: Yep, that wasn't intentional.  Just an oversight on my part.  Updated the tests to catch this.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -262,19 +294,26 @@ private void PullBaseImages()
             }
         }
 
+        private IEnumerable<ImageData> GetBuiltImages() => this.repoList
```
MichaelSimons: Might be useful to also have a `GetBuiltTags` method as well.  It could be used from `PushImages` and `WriteBuildSummary`

src/Microsoft.DotNet.ImageBuilder/src/ExecuteHelper.cs
```diff
@@ -29,15 +29,24 @@ public static class ExecuteHelper
...
             bool isDryRun,
             string errorMessage = null,
             string executeMessageOverride = null)
         {
-            Execute(
-                new ProcessStartInfo(fileName, args),
+            return ExecuteWithRetry(new ProcessStartInfo(fileName, args), isDryRun, errorMessage, executeMessageOverride);
+        }
+
+        public static Process ExecuteWithRetry(
```
MichaelSimons: Is this change necessary anymore?

mthalman: Removed

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -298,6 +298,9 @@ private void PullBaseImages()
             .Where(repoData => repoData.Images != null)
             .SelectMany(repoData => repoData.Images.Values);
 
+        private IEnumerable<string> GetBuildTags() =>
```
MichaelSimons: Naming pattern inconsistency with `GetBuiltImages`
```suggestion
        private IEnumerable<string> GetBuiltTags() =>
```

PR #441, opened 2020-03-13
Store a schema version in the image info file

mthalman:
Stores a schema version in the image info file.  The previous root element of the image info file was a `RepoData` array which prevents a scalar property from being set on it.  So a new `ImageArtifactDetails`  element was defined as the root which contains a list of `RepoData` objects and the schema version.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Models/Image/ImageArtifactDetails.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Models/Image/ImageArtifactDetails.cs
@@ -1,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.ImageBuilder.Models.Image
+{
+    public class ImageArtifactDetails
+    {
+        private const string schemaVersion = "1.0";
+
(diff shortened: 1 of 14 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Models/Image/ImageArtifactDetails.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Microsoft.DotNet.ImageBuilder.Models.Image
+{
+    public class ImageArtifactDetails
+    {
+        private const string schemaVersion = "1.0";
+
+        public string SchemaVersion
+        {
+            get { return schemaVersion; }
+            set { }
```
MichaelSimons: Shouldn't this deserialize the value?  We could validate it in `LoadFromContent` any throw if the value is not the current version?

mthalman: I thought about that for a while but I'm not sure what purpose that would serve.  If the schema of the file that is being loaded is incompatible with the currently defined schema, the deserialization will fail anyway before we even get a chance to check the version value.  If the schema is different but compatible, I'm not sure we want to fail on load.  There's been a lot of cases where new fields have been added to the schema and I don't want to have to update the files stored in the versions repo so that their schema version matches just because of that change.

MichaelSimons: That makes sense.

Issue #438, opened 2020-03-12
Track additional data in image info


mthalman:
In order to accurately provide telemetry data on image pulls, it's necessary to be able to correlate Docker image digests with the data we want to pivot on.  We need to collect this data within the image info file for this purpose:

- [x] Shared tags
- [x] Architecture
- [x] OS Type [Linux|Windows]
- [x] OS version (friendly name)
- [x] .NET product version (major.minor.patch)
- [x] Commit SHA
- [x] Created date
- [x] Manifest Tag Digests

mthalman:
https://github.com/dotnet/docker-tools/pull/437 added the tracking of the following data:
* Architecture
* OS Type
* OS version

mthalman:
All this data is now being collected by the nightly branch.  Keeping this open until it's been rolled out across the board.

mthalman:
This has now been rolled out to all the repos.  Closing.

PR #442, opened 2020-03-17
Store product version in image info

mthalman:
This allows us to associate each image defined in the image info file with a specific product version.  Having this data will allow for telemetry correlation between image pulls and product version.  The product version will need to be defined in the manifest which will then flow into the outputted image info file.

Related to #438

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
@@ -32,7 +32,6 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
         public IEnumerable<string> ExternalFromImages { get; private set; }
         public IEnumerable<string> InternalFromImages { get; private set; }
         public Platform Model { get; private set; }
-        public ImageInfo Image { get; private set; }
         public IEnumerable<string> OverriddenFromImages { get => _overriddenFromImages; }
         private string FullRepoModelName { get; set; }
         private string RepoName { get; set; }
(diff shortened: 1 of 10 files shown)


MichaelSimons:
We parse out the "dotnetVersion" from the path in GenerateBuildMaxtrixCommand.  Would it make sense to utilize the major.minor parts of this new product version?

mthalman:
> We parse out the "dotnetVersion" from the path in GenerateBuildMaxtrixCommand. Would it make sense to utilize the major.minor parts of this new product version?

Ah, excellent point.  It'd be great to simplify that.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -33,23 +32,20 @@ public class PlatformInfo
         public IEnumerable<string> ExternalFromImages { get; private set; }
         public IEnumerable<string> InternalFromImages { get; private set; }
         public Platform Model { get; private set; }
+        public ImageInfo Image { get; private set; }
```
MichaelSimons: I've tried hard to keep circular references out of the view model.  By that I mean the children don't have a reference to their parents.  I have waffled on this decision as in some areas it may help the usage patterns.  In other cases it has been helpful such as when shared tags were introduced.  One of the problems these references introduce is that the `parent` properties aren't valid to call within the instantiation process of the view model itself because the parent isn't fully instantiated.  I don't know if you thought about this deeply or not.

An alternative to the new usage of this property would be to lookup the ImageInfo for the specified PlatformInfo.

mthalman: Makes sense.  I've updated with a lookup method.

PR #444, opened 2020-03-17
Store Created date in image info output

mthalman:
Stores the Created date in the image file for each platform to be used for tracking telemetry data of the image pulls.

Related to #438


mthalman:
@MichaelSimons - Would using a name of `Updated` instead of `Created` be better for the shared tags since they can be updated over time?  Or would it even make sense to just use `Updated` across the board for consistency?  LMK

MichaelSimons:
> Would using a name of Updated instead of Created be better for the shared tags since they can be updated over time? Or would it even make sense to just use Updated across the board for consistency? LMK

I look at it from the perspective that this is image information.  Every build is producing new images.  Therefore in this context, `created` feels appropriate to me.  The tags are what float between images so they are the only thing that get `updated`.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestOptions.cs
```diff
@@ -12,6 +12,10 @@ public class PublishManifestOptions : DockerRegistryOptions, IFilterableOptions
 
         public ManifestFilterOptions FilterOptions { get; } = new ManifestFilterOptions();
 
+        public string ImageInfoPath { get; set; }
```
MichaelSimons: What is the scenario for wanting different paths here?

src/Microsoft.DotNet.ImageBuilder/src/Models/Image/SharedTag.cs
```diff
@@ -0,0 +1,26 @@
...
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.DotNet.ImageBuilder.Models.Image
+{
+    public class SharedTag : IComparable<SharedTag>
+    {
+        public string Name { get; set; }
+
+        public DateTime Created { get; set; }
```
MichaelSimons: Isn't it inconsistent that we are tracking the `Created` timestamp at the tag level for manifests but not for images?  Is there a need since the manifests are all created together?

mthalman:
> > Would using a name of Updated instead of Created be better for the shared tags since they can be updated over time? Or would it even make sense to just use Updated across the board for consistency? LMK
> 
> I look at it from the perspective that this is image information. Every build is producing new images. Therefore in this context, `created` feels appropriate to me. The tags are what float between images so they are the only thing that get `updated`.

So are you saying that it would be appropriate to change the name to `Updated` for the `SharedTags`?

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestOptions.cs
mthalman: There is none.  When I was coding this, I was thinking the pipeline would need to keep them separate as separate files for some reason.  But that's not needed.  I'll change this to just update the file in place.

PR #449, opened 2020-03-25
Helper script for pulling internal Windows images

mthalman:


mthalman:
Moving to internal repo instead.

PR #457, opened 2020-03-30
Store commit URL in image info file

### LINKED ISSUE #438: Track additional data in image info

mthalman:
Includes a `commitUrl` field in the image info file to provide a link to the specific Dockerfile the image was built from.

Fixes #438

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/GitServiceExtensions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/GitServiceExtensions.cs
@@ -1,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.ImageBuilder.ViewModel;
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    public static class GitServiceExtensions
+    {
+        public static string GetDockerfileCommitUrl(
+            this IGitService gitService,
+            PlatformInfo platform,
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.ImageBuilder/src/GitServiceExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+            this IGitService gitService,
+            PlatformInfo platform,
+            string sourceRepoUrl,
+            string sourceBranch = null)
+        {
+            string branchOrShaPathSegment = sourceBranch ??
+                gitService.GetCommitSha(platform.DockerfilePath, useFullHash: true);
+
+            string dockerfileRelativePath = PathHelper.NormalizePath(platform.DockerfilePathRelativeToManifest);
+            return $"{sourceRepoUrl}/blob/{branchOrShaPathSegment}/{dockerfileRelativePath}";
```
MichaelSimons: Should this fail if the sourceRepoUrl isn't specified?  It seems like it silently fails which seems like something we should get a notification of.

mthalman: Yes, added a check.

Issue #436, opened 2020-03-09
PublishImageInfoCommand should remove out of date image info


MichaelSimons:
The [current Image Info data](https://github.com/dotnet/versions/blob/master/build-info/docker/image-info.dotnet-dotnet-docker-master.json) for the .NET Docker repos contains image info for images that are no longer supported e.g. .NET Core 2.2 and 3.0).  The PublishImageInfoCommand should be removing the image info once an image reaches EOL.  The Manifest file is the source of truth so any image in the image info that is not in the manifest should be removed.

mthalman:
This logic needed to be temporarily disabled due to it corrupting the image info content: https://github.com/dotnet/docker-tools/issues/436.  Need to get this fixed and re-enabled.

PR #459, opened 2020-04-01
Clean old data from image info file on publish

mthalman:
We want to ensure the published image info files contain data only for currently supported images.  The manifest represents what is currently supported.  This updates the publish logic to remove any content from the target image info file that isn't represented in the manifest.

Related to #436

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
@@ -133,7 +133,6 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                     for (int platformIndex = imageData.Platforms.Count - 1; platformIndex >= 0; platformIndex--)
                     {
                         PlatformData platformData = imageData.Platforms[platformIndex];
-
                         PlatformInfo manifestPlatform = manifestImage.AllPlatforms
                             .FirstOrDefault(manifestPlatform => platformData.Equals(manifestPlatform));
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
```diff
@@ -101,5 +103,60 @@ public override async Task ExecuteAsync()
...
+                    if (manifestImage is null)
+                    {
+                        repoData.Images.Remove(imageData);
+                        continue;
+                    }
+                    
+                    for (int platformIndex = imageData.Platforms.Count - 1; platformIndex >= 0; platformIndex--)
+                    {
+                        PlatformData platformData = imageData.Platforms[platformIndex];
+
```
MichaelSimons, line 136: nit: remove extra line to be consistent in formatting with other areas like the repo and image.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
```diff
@@ -101,5 +103,60 @@ public override async Task ExecuteAsync()
...
+                            .FirstOrDefault(manifestPlatform => platformData.Equals(manifestPlatform));
+
+                        // If there doesn't exist a matching platform in the manifest, remove it from the image info
+                        if (manifestPlatform is null)
+                        {
+                            imageData.Platforms.Remove(platformData);
+                        }
+                    }
+
+                    // If all of the image's platforms were removed, then remove the image too
```
MichaelSimons, line 147: Would this only happen in the following situations?
1. The model is defined this way - a repo with no images or an image with no platforms.  Maybe the model validation should prevent this situation?  IDK - It seems harmless.
1. A build hasn't happened which builds all of the manifest images.

I am wondering if this path is necessary?

mthalman: That's true.  Looking at it again, I don't think it's possible for this condition to ever occur.  Since the correlation of an `ImageData` to it's corresponding manifest `ImageInfo` is solely based on finding an equivalent platform, you could never have a case where the correlation was able to be done and yet still have all the platforms removed.  If the correlation isn't made, the image would get deleted up above before enumerating the platforms.  So yeah, I'll get rid of this.

Issue #461, opened 2020-04-03
Be more aggressive on cleaning up images stored in the staging and testing locations in ACR


MichaelSimons:
1. Currently all staging repositories are [kept for 30 days](https://github.com/dotnet/docker-tools/blob/master/eng/Invoke-AcrCleanup.ps1#L32).  I don't see a need to keep these so long.  We should be way more aggressive on cleaning these up.  Keep them for 1 week, 2 weeks at the most.
1. The testing repositories should have a retention policy set on them to keep images for 1 week.

mthalman:
This has been integrated into the cleanup pipeline as part of https://github.com/dotnet/docker-tools/pull/506.

Issue #419, opened 2020-02-24
dotnet-docker-tools-cleanup-acr-images pipeline needs to be more efficient


MichaelSimons:
The dotnet-docker-tools-cleanup-acr-images pipeline has been timing out ever since [this fix was submitted to cleanup the nightly repos](https://github.com/dotnet/docker-tools/issues/378).  In this [recent run of the build](https://dev.azure.com/dnceng/internal/_build/results?buildId=533918), the build timed out after 60 minutes of running.  It queried 217 repos and only deleted 87 images.  In 1 hour, I expect the pipeline would complete all of the  cleanup tasks. Consider running the cleanup logic asynchronously so that we can make better use of our build agents.

PR #462, opened 2020-04-03
Command to delete unnecessary images from ACR

mthalman:
Adds a command that is intended to replace the logic in https://github.com/dotnet/docker-tools/blob/master/eng/Invoke-AcrCleanup.ps1.  It provides a parallel implementation to query and delete all unnecessary images from ACR in a much faster execution time.

There doesn't exist a .NET library for communicating with the ACR REST API so I had to implement my own.  Note that the [Container Registry REST API](https://docs.microsoft.com/en-us/rest/api/containerregistry/) isn't actually sufficient because that API only provides access for managing Azure resources.  But the things that we need to access are within the registry itself which is outside the jurisdiction of Azure resources.

Related to #419
Related to #461

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
@@ -1,0 +1,185 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.DotNet.ImageBuilder.Models.Acr;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    [Export(typeof(ICommand))]
+    public class CleanAcrImagesCommand : Command<CleanAcrImagesOptions>
+    {
+        private readonly IAcrClientFactory acrClientFactory;
+        private readonly ILoggerService loggerService;
+
+        [ImportingConstructor]
+        public CleanAcrImagesCommand(IAcrClientFactory acrClientFactory, ILoggerService loggerService)
+        {
+            this.acrClientFactory = acrClientFactory ?? throw new ArgumentNullException(nameof(acrClientFactory));
+            this.loggerService = loggerService ?? throw new ArgumentNullException(nameof(loggerService));
+        }
+
+        public override async Task ExecuteAsync()
+        {
+            this.loggerService.WriteHeading("FINDING IMAGES TO CLEAN");
+
+            this.loggerService.WriteSubheading($"Connecting to ACR '{Options.RegistryName}'");
+            using IAcrClient acrClient = await this.acrClientFactory.CreateAsync(
+                Options.RegistryName, Options.Tenant, Options.Username, Options.Password);
+
+            this.loggerService.WriteSubheading($"Querying catalog of ACR '{Options.RegistryName}'");
+            Catalog catalog = await acrClient.GetCatalogAsync();
+
+            this.loggerService.WriteSubheading($"Querying repository details of ACR '{Options.RegistryName}'");
+            IEnumerable<Task<Repository>> getRepositoryTasks = catalog.RepositoryNames
+                .Where(repoName => !IsPublicRepo(repoName) || IsNightlyRepo(repoName))
+                .Select(repoName => acrClient.GetRepositoryAsync(repoName))
+                .ToArray();
+            await Task.WhenAll(getRepositoryTasks);
+            IEnumerable<Repository> repositories = getRepositoryTasks
+                .Select(task => task.Result)
+                .ToArray();
+
+            List<string> deletedRepos = new List<string>();
+            List<string> deletedImages = new List<string>();
+
+            this.loggerService.WriteHeading("DELETING IMAGES");
+            await Task.WhenAll(GetRepoProcessingTasks(acrClient, repositories, deletedRepos, deletedImages));
+
+            LogSummary(catalog, deletedRepos, deletedImages);
+        }
+
+        private IEnumerable<Task> GetRepoProcessingTasks(
+            IAcrClient acrClient, IEnumerable<Repository> repositories, List<string> deletedRepos, List<string> deletedImages)
+        {
+            int publicCount = 0;
+            int nonPublicCount = 0;
+            foreach (Repository repository in repositories)
+            {
+                if (IsPublicRepo(repository.Name))
+                {
+                    publicCount++;
+                    if (publicCount <= 2)
+                    {
+                        yield return ProcessPublicRepoAsync(acrClient, deletedImages, repository);
+                    }
+                }
+                else
+                {
+                    nonPublicCount++;
+                    if (nonPublicCount <= 20)
+                    {
+                        yield return ProcessNonPublicRepoAsync(acrClient, deletedRepos, repository);
+                    }
+                }
+            }
+        }
+
+        private void LogSummary(Catalog catalog, List<string> deletedRepos, List<string> deletedImages)
+        {
+            this.loggerService.WriteHeading("SUMMARY");
+
+            this.loggerService.WriteSubheading("Deleted repositories:");
+            foreach (string deletedRepo in deletedRepos)
+            {
+                this.loggerService.WriteMessage($"\t{deletedRepo}");
+            }
+
+            this.loggerService.WriteMessage();
+
+            this.loggerService.WriteSubheading("Deleted untagged images:");
+            foreach (string deletedImage in deletedImages)
+            {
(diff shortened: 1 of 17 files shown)


mthalman:
I got some testing code I need to clean up.

MichaelSimons:
What do you think about making the `CleanAcrImagesCommand` more metadata/config driven?  By that I mean the caller would specify the options on what should be cleaned up.  For example delete all repos with prefix X older than Y days.  Another would be delete all images from repos with prefix S older than T days.

The cleanup build would then invoke the cmd three times for our cleanup policy.  Once to cleanup staging report, once to cleanup old test images, and once more to cleanup old nightly images.

This would support more flexibility over time without having to modify the command's source.

mthalman:
> What do you think about making the `CleanAcrImagesCommand` more metadata/config driven? 

While that could be useful, I don't really see it as something that's necessary at this point.  I doubt that it would be something we take advantage of very often.  Either way, a change to the repo has to be made.

MichaelSimons:
> Either way, a change to the repo has to be made.
Technically changes to image-builder require two commits today.

> While that could be useful, I don't really see it as something that's necessary at this point.
I can't argue that it is necessary.  I was viewing this as something more generally useful for example you could use it in a dev scenario to delete some of the straggler test repos - e.g. `test/dotnet/test`

mthalman:
I'm not even sure how this would represented in a simple way with command options.  The rules are not as simple as "delete all repos with prefix X older than Y days".

These are the rules that currently exist:
* Delete any manifest that is not tagged, is older than 30 days, and exists in repo with prefix "public".
* Delete any manifest that is older than 7 days and exists in repo with prefix "test".
* Delete any repo that is not prefixed with "public" or "test" and is older than 15 days.

Sure, it could be done, but it's going to be complicated. Is it really worth that?

MichaelSimons:
This is what I had in mind for options.

repo = name of the repo(s) to cleanup - wildcard chars * and ? are supported - similar to the `path` argument.
action = the action to perform on the repo (delete, pruneAll, pruneDangling).
days = how old an image/repo must be when performing the action.

cleanAcrImages -repo staging/* -action delete -days 15
cleanAcrImages -repo test/* -action pruneAll -days 7
cleanAcrImages -repo dotnet/core-nightly/* -action pruneDangling -days 30

That being said, I have no strong objection to the hard coded approach.  I will review your changes.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -0,0 +1,194 @@
...
+            {
+                if (IsPublicRepo(repository.Name))
+                {
+                    yield return ProcessPublicRepoAsync(acrClient, deletedImages, repository);
+                }
+                else if (IsTestRepo(repository.Name))
+                {
+                    yield return ProcessTestRepoAsync(acrClient, deletedImages, repository);
+                }
+                else
```
MichaelSimons, line 71: For the same paranoid reason above, I think it would be better to explicitly check for `staging` here and have an else that throws an exception.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -0,0 +1,194 @@
...
+                    foreach (string tag in deleteResponse.TagsDeleted)
+                    {
+                        messageBuilder.AppendLine($"\t{tag}");
+                    }
+
+                    this.loggerService.WriteMessage(messageBuilder.ToString());
+                }
+                else
+                {
+                    this.loggerService.WriteMessage($"Deleted repository '{repository.Name}'");
```
MichaelSimons, line 179: I think dry run should be showing everything that is being deleted just like what is done for the real operation.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -95,43 +94,52 @@ private void LogSummary(Catalog catalog, List<string> deletedRepos, List<string>
...
+            IAcrClient acrClient, List<string> deletedImages, Repository repository, IEnumerable<ManifestAttributes> manifests)
         {
             List<Task> tasks = new List<Task>();
-            foreach (Manifest manifest in manifests)
+            foreach (ManifestAttributes manifest in manifests)
             {
                 tasks.Add(DeleteManifestAsync(acrClient, deletedImages, repository, manifest));
             }
-
+
```
MichaelSimons, line 137:
```suggestion

```

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -95,43 +94,52 @@ private void LogSummary(Catalog catalog, List<string> deletedRepos, List<string>
...
-            RepositoryManifests repoManifests = await acrClient.GetRepositoryManifests(repository.Name);
-            IEnumerable<Manifest> expiredTestImages = repoManifests.Manifests
+            RepositoryManifests repoManifests = await acrClient.GetRepositoryManifestsAsync(repository.Name);
+            IEnumerable<ManifestAttributes> expiredTestImages = repoManifests.Manifests
                 .Where(manifest => IsExpired(manifest.LastUpdateTime, 7));
             await DeleteManifestsAsync(acrClient, deletedImages, repository, expiredTestImages);
         }
 
-        private async Task ProcessPublicRepoAsync(IAcrClient acrClient, List<string> deletedImages, Repository repository)
+        private async Task ProcessPublicRepoAsync(
```
MichaelSimons, line 120:
```suggestion
        private async Task ProcessPublicNightlyRepoAsync(
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -95,43 +94,52 @@ private void LogSummary(Catalog catalog, List<string> deletedRepos, List<string>
...
-        private async Task ProcessTestRepoAsync(IAcrClient acrClient, List<string> deletedImages, Repository repository)
+        private async Task ProcessTestRepoAsync(
+            IAcrClient acrClient, List<string> deletedImages, Repository repository)
         {
-            RepositoryManifests repoManifests = await acrClient.GetRepositoryManifests(repository.Name);
-            IEnumerable<Manifest> expiredTestImages = repoManifests.Manifests
+            RepositoryManifests repoManifests = await acrClient.GetRepositoryManifestsAsync(repository.Name);
+            IEnumerable<ManifestAttributes> expiredTestImages = repoManifests.Manifests
                 .Where(manifest => IsExpired(manifest.LastUpdateTime, 7));
             await DeleteManifestsAsync(acrClient, deletedImages, repository, expiredTestImages);
```
MichaelSimons, line 117: What do you think about checking if the repo is empty and deleting it if it is?  The reason I mention this is I noticed a couple test repos that should be cleaned up - `test/test` and `test/dotnet/test`

mthalman: Yep, that makes sense to clean that up.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -109,15 +109,21 @@ private async Task LogSummaryAsync(IAcrClient acrClient, List<string> deletedRep
         }
 
         private async Task ProcessTestRepoAsync(
-            IAcrClient acrClient, List<string> deletedImages, Repository repository)
+            IAcrClient acrClient, List<string> deletedImages, List<string> deletedRepos, Repository repository)
         {
             RepositoryManifests repoManifests = await acrClient.GetRepositoryManifestsAsync(repository.Name);
+            if (!repoManifests.Manifests.Any())
```
MichaelSimons, line 115: Is it possible to incorporate the expired check so that
1. the repo is deleted if all manifests are expired
1. only one operation is necessary

```suggestion

            IEnumerable<ManifestAttributes> expiredTestImages = repoManifests.Manifests
                .Where(manifest => IsExpired(manifest.LastUpdateTime, 7));
                .ToArray();
            if (expiredTestImages.Count() == repoManifests.Manifests.Count())
            {
                await DeleteRepositoryAsync(acrClient, deletedRepos, repository);
                return;
            }
            
            await DeleteManifestsAsync(acrClient, deletedImages, repository, expiredTestImages);
```

mthalman: What do you mean by "only one operation is necessary"?  One operation to do what?

MichaelSimons: I mean if the repository only contains expired manifests, all expired manifests and the repository itself can be deleted with one request.

mthalman: Great suggestion.

PR #463, opened 2020-04-06
Merge image info schema changes

mthalman:
Merges the image info schema changes from the feature branch.

Related to #438

### DIFF
--- a/eng/check-base-image-subscriptions.json
+++ b/eng/check-base-image-subscriptions.json
@@ -3,6 +3,19 @@
     "repoInfo": {
       "owner": "dotnet",
       "name": "dotnet-docker",
+      "branch": "master"
+    },
+    "manifestPath": "manifest.json",
+    "pipelineTrigger": {
+      "id": 373,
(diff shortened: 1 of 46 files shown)


eng/check-base-image-subscriptions.json
```diff
@@ -3,37 +3,13 @@
     "repoInfo": {
```
MichaelSimons, line 3: Shouldn't the changes to this file be reverted since these were specific to this branch?

Additionally we should cleanup theses [pipeline changes in master](https://github.com/dotnet/docker-tools/blob/master/eng/pipelines/check-base-image-updates.yml#L10) specific to this branch.

eng/pipelines/dotnet-buildtools-image-builder-pr.yml
```diff
@@ -2,6 +2,7 @@ pr:
   branches:
     include:
     - master
+    - feature/image-info-schema
```
MichaelSimons, line 5: Should these also be reverted?  Maybe they could be changed to `feature/*`?

PR #468, opened 2020-04-08
Adds option to specify which version components are significant

mthalman:
The current logic for consuming the productVersion attribute from the manifest is broken because it only makes use of the major/minor components of the value.  That's fine for .NET Core but doesn't work for .NET Framework for framework versions like 4.7.2 where the "2" is significant in the generation of the matrix.  This results in incorrectly defined build/test legs.

Fixing this by adding an option to specify how many components of the product version are significant.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixOptions.cs
@@ -42,7 +42,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 "Name of custom build leg grouping to use.");
             CustomBuildLegGrouping = customBuildLegGrouping;
 
-            int productVersionComponents = 0;
+            int productVersionComponents = 2;
             syntax.DefineOption(
                 "productVersionComponents",
                 ref productVersionComponents,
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixOptions.cs
```diff
@@ -40,6 +41,13 @@ public override void DefineOptions(ArgumentSyntax syntax)
                 ref customBuildLegGrouping,
                 "Name of custom build leg grouping to use.");
             CustomBuildLegGrouping = customBuildLegGrouping;
+
+            int productVersionComponents = 0;
```
MichaelSimons, line 45: Given this is an option, 0 doesn't seem like a good default as it will cause an exception.  My suggestions are:

1. Change this to a parameter
2. Set the default to something arbitrary like 2

mthalman: It will only cause an exception if a productVersion is set.

MichaelSimons: My first option doesn't make sense.  It still seem like a poor choice that a default option value will cause an exception.  The documentation makes no reference to this being required.  If we can provide a sensible default I feel we should and document the behavior.  I feel that would be consistent with the existing options across ImageBuilder.  Our builds' usage of the commands should be explicit in the options/values.

PR #469, opened 2020-04-08
Update ImageBuilder tags

mthalman:


mthalman:
Merging since build passed.

PR #470, opened 2020-04-08
Update common build.ps1 scipt's handling of path arg

MichaelSimons:
The way the `Path` filter was being appended to the ImageBuilder args made it difficult and confusing to use.  The `--path` option wasn't being prepended so the user needed to include it in the arg value. It was in a way behaving no different than the `OptionalImageBuilderArgs` except that it overrode the `Version` arg.  I changed the behavior so that `--path` gets prepended to the arg value.

### DIFF
--- a/eng/common/build.ps1
+++ b/eng/common/build.ps1
@@ -16,7 +16,7 @@ param(
     # Type of architecture to filter by
     [string]$Architecture,
 
-    # Additional custom path filters (overrides Version)
+    # Additional custom path filter (overrides Version)
     [string]$Path,
 
     # Path to manifest file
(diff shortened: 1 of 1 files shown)


eng/common/build.ps1
```diff
@@ -58,7 +58,7 @@ try {
     }
 
     if ($Path) {
-        $args += " $Path"
+        $args += " --path $Path"
```
mthalman, line 61: This doesn't let you specify multiple path args though.

MichaelSimons: hmmm, I see.  The name is confusing to me.  It implies a 1-1 mapping to the ImageBuilder path option.  Does that make sense to you?

MichaelSimons: I updated it to support multiples - let me know what you think.

mthalman: The original name was `PathFilters` but was renamed to `Path` in #429.  I missed that in the review.

PR #477, opened 2020-04-10
Fix graph generation for base image change detection

mthalman:
The logic in `GetStaleImagesCommand` produced an incorrect set of Dockerfile paths to be rebuild in the case where image info data didn't exist for a Dockerfile which other Dockerfiles were dependent upon.  For example, it there was no image data for a runtime-deps Dockerfile, then the command would only output that the runtime-deps Dockerfile needed to be rebuilt.  It didn't walk the graph and also output that runtime and aspnet also needed to be rebuilt.

There already existed logic for handling the graph walking when the image info _did_ exist but that wasn't executed for the scenario where it didn't exist.  Updated the logic to fix this.

Other changes:
* Added a test to catch this case and also fixed the logic in another test because it wasn't correctly constructing a graph of Dockerfiles.
* Removed some logging logic in `PlatformInfo.GetOSDisplayName` because that gets executed a lot when loading image info files now (not caused by these changes) which leads to a lot of unnecessary log output.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
@@ -129,8 +129,6 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
         {
             string displayName;
             string os = Model.OsVersion;
-            Logger.WriteMessage($"os: {os}");
-            Logger.WriteMessage($"osType: {Model.OS}");
 
             if (Model.OS == OS.Windows)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -129,8 +129,6 @@ public string GetOSDisplayName()
         {
             string displayName;
             string os = Model.OsVersion;
-            Logger.WriteMessage($"os: {os}");
```
MichaelSimons, line 132: I noticed this the other day, thanks for cleaning this up.

PR #478, opened 2020-04-10
Clean out-of-date image info content on publish

### LINKED ISSUE #436: PublishImageInfoCommand should remove out of date image info

mthalman:
We want to ensure the published image info files contain data only for currently supported images. The manifest represents what is currently supported. This updates the publish logic to remove any content from the target image info file that isn't represented in the manifest.

This is mostly just restoring the code from https://github.com/dotnet/docker-tools/pull/459 but fixes the bug that was contained in it and updates the test to account for the bug.  The bug was that it was ending up not finding any matching content between the manifest and target image info file because repo names weren't being compared correctly.  This meant that the source image info would simply overwrite the contents of the target image info.

Fixes #436

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
@@ -109,7 +109,10 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             for (int repoIndex = imageArtifactDetails.Repos.Count - 1; repoIndex >= 0; repoIndex--)
             {
                 RepoData repoData = imageArtifactDetails.Repos[repoIndex];
-                RepoInfo manifestRepo = Manifest.AllRepos.FirstOrDefault(manifestRepo => manifestRepo.Name == repoData.Repo);
+                
+                // Since the registry name is not represented in the image info, make sure to compare the repo name with the
+                // manifest's repo model name which isn't registry-qualified.
+                RepoInfo manifestRepo = Manifest.AllRepos.FirstOrDefault(manifestRepo => manifestRepo.Model.Name == repoData.Repo);
 
                 // If there doesn't exist a matching repo in the manifest, remove it from the image info
                 if (manifestRepo is null)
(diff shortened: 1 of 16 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
```diff
@@ -109,7 +109,10 @@ private void RemoveOutOfDateContent(ImageArtifactDetails imageArtifactDetails)
             for (int repoIndex = imageArtifactDetails.Repos.Count - 1; repoIndex >= 0; repoIndex--)
             {
                 RepoData repoData = imageArtifactDetails.Repos[repoIndex];
-                RepoInfo manifestRepo = Manifest.AllRepos.FirstOrDefault(manifestRepo => manifestRepo.Name == repoData.Repo);
+                
+                // Since the registry name is not represented in the image info, make sure to compare the repo name with the
+                // manifest's repo model name which isn't registry-qualified.
+                RepoInfo manifestRepo = Manifest.AllRepos.FirstOrDefault(manifestRepo => manifestRepo.Model.Name == repoData.Repo);
```
MichaelSimons, line 115: Does this logic need to account for potential repo-overrides that may be specified via the options?  I'm not even sure that is used anymore.  Regardless my concern is the view model is where this override logic resides (now and in the future).  Would it be better to introduce a QualifiedName in the view model for the name with the registry and then the Name property could safely be used here?

mthalman: Yes, that sounds good.  It'd be best to not have to access the model directly.  I'll also get rid of repo-override as I do not see it used anywhere.

PR #482, opened 2020-04-15
Support manifest-specific image infos in ImageBuilder

mthalman:
The changes in https://github.com/dotnet/docker-tools/pull/478 provide the logic to remove any content in an image file that isn't represented in the manifest.  The problem is that sample images are contained in a separate manifest but have their image info contained in the same file as the product.  This would cause the clean-up logic for publishing sample images to blow away the product image info content because it doesn't exist in the sample's manifest file (and vice versa with publishing product images).

This can be solved by having separate image info files that each correspond to a single manifest.  There should be a 1:1 correspondence to avoid issues.  These changes update Image Builder to support this.  The previous implementation was never great because knowledge of the image info file name was spread across both pipeline-related files and Image Builder code.  That implementation makes things even worse now when considering the need to vary an image info file's name by manifest.

The new design removes all derivation of image info file names from Image Builder.  It's now stored in the subscriptions file so that it is explicitly paired with the appropriate manifest file.  The publish job still needs to have knowledge of the path as well.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
@@ -18,12 +18,16 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         public string Repo { get; set; }
         public string Username { get; set; }
 
-        public GitOptions(string defaultOwner = "", string defaultRepo = "", string defaultBranch = "", string defaultPath = "")
+        public GitOptions()
         {
-            this.Owner = defaultOwner ?? throw new ArgumentNullException(nameof(defaultOwner));
-            this.Repo = defaultRepo ?? throw new ArgumentNullException(nameof(defaultRepo));
-            this.Branch = defaultBranch ?? throw new ArgumentNullException(nameof(defaultBranch));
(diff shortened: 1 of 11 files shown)


mthalman:
You can see how this changes the pipeline-related files in #483.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
```diff
@@ -18,7 +18,7 @@ public class GitOptions
         public string Repo { get; set; }
         public string Username { get; set; }
 
-        public GitOptions(string defaultOwner, string defaultRepo, string defaultBranch, string defaultPath)
+        public GitOptions(string defaultOwner = "", string defaultRepo = "", string defaultBranch = "", string defaultPath = "")
```
MichaelSimons, line 21: What are you trying to accomplish with the parameters being defaulted to empty in addition to the arg null validation?

The code still allows the properties to be set directly to null.  The caller specify a subset of values.  I am wondering what the difference would be to have a parameterless constructor or default the parameters to null and remove the arg validation.

Regardless of the approach, it seems like the descriptions of the options don't display well when a default value isn't supplied.  e.g. `GitHub branch to write to (defaults to )`  You could solve that easily by quoting the default value.

Issue #474, opened 2020-04-09
Consider generating image info file in PR build


mthalman:
Currently, PR builds do not generate an image info file since it's not desired to have one published to an external repo for such builds.  But this exposes a hole in the workflow because it's not executing logic that would get executed in the official builds.  There have been numerous cases where issues related to image info have occurred in official builds that could have been caught if the PR build had generated one.

I propose that PR builds follow the same workflow for generating an image info file except for the task of actually publishing it to an external repo.

MichaelSimons:
The broader issue here is that we should be running as much of the publish leg as possible without publishing the images.  This helps surface issues in PR validation.

PR #495, opened 2020-04-27
Fix commands to support dry-run option better

mthalman:
The `publishImageInfo` and `publicMcrDocs` commands did not support the `dry-run` option very well because they required GitHub credentials in order to read some files.  This is problematic as part of the work to support #474 because no credentials will be available in such a scenario.  So in order to still exercise these commands in that scenario, the commands need to be updated.

I've refactored the commands so that they now do all of their processing logic that generates the file content to be pushed before accessing GitHub.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
@@ -45,15 +45,13 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 GetUpdatedReadmes(productRepo)
                 .Concat(GetUpdatedTagsMetadata(productRepo));
 
-            if (Options.IsDryRun)
+            foreach (GitObject gitObject in gitObjects)
             {
-                foreach (GitObject gitObject in gitObjects)
-                {
-                    this.loggerService.WriteMessage(
-                        $"Updated file '{gitObject.Path}' with contents:{Environment.NewLine}{gitObject.Content}{Environment.NewLine}");
-                }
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
```diff
@@ -21,61 +21,87 @@ public class PublishMcrDocsCommand : ManifestCommand<PublishMcrDocsOptions>
...
 
             string productRepo = GetProductRepo();
 
-            using IGitHubClient gitHubClient = this.gitHubClientFactory.GetClient(Options.GitOptions.ToGitHubAuth(), Options.IsDryRun);
+            IEnumerable<GitObject> gitObjects =
+                GetUpdatedReadmes(productRepo)
+                .Concat(GetUpdatedTagsMetadata(productRepo));
 
-            await GitHelper.ExecuteGitOperationsWithRetryAsync(async () =>
+            if (Options.IsDryRun)
```
MichaelSimons, line 48: I have a tendency to want to avoid dry run specific logic.  The primary reason is to avoid regressions because of specialized code paths that are not always exercised.  In cases like this, why not emit this output all of the time?

MichaelSimons: This comment also applies to https://github.com/dotnet/docker-tools/pull/495/files#diff-2305a611cb553ee6ccd1b29731b97be9R50

mthalman: Fixed

Issue #497, opened 2020-04-27
NullRefEx thrown when running the build command with the dry run option


MichaelSimons:
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.DotNet.ImageBuilder.ExecuteHelper.Execute(ProcessStartInfo info, Func`2 executor, Boolean isDryRun, String errorMessage, String executeMessageOverride) in /image-builder/src/ExecuteHelper.cs:line 101
   at Microsoft.DotNet.ImageBuilder.ExecuteHelper.ExecuteWithRetry(ProcessStartInfo info, Action`1 processStartedCallback, Boolean isDryRun, String errorMessage, String executeMessageOverride) in /image-builder/src/ExecuteHelper.cs:line 56
   at Microsoft.DotNet.ImageBuilder.ExecuteHelper.ExecuteWithRetry(String fileName, String args, Boolean isDryRun, String errorMessage, String executeMessageOverride) in /image-builder/src/ExecuteHelper.cs:line 41
   at Microsoft.DotNet.ImageBuilder.DockerHelper.PullImage(String image, Boolean isDryRun) in /image-builder/src/DockerHelper.cs:line 97
   at Microsoft.DotNet.ImageBuilder.DockerService.PullImage(String image, Boolean isDryRun) in /image-builder/src/DockerService.cs:line 25
   at Microsoft.DotNet.ImageBuilder.DockerServiceExtensions.PullBaseImages(IDockerService dockerService, ManifestInfo manifest, Boolean isDryRun) in /image-builder/src/DockerServiceExtensions.cs:line 21
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.PullBaseImages() in /image-builder/src/Commands/BuildCommand.cs:line 314
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.ExecuteAsync() in /image-builder/src/Commands/BuildCommand.cs:line 43
   at Microsoft.DotNet.ImageBuilder.ImageBuilder.Main(String[] args) in /image-builder/src/ImageBuilder.cs:line 66
```

This looks to be a regression caused by https://github.com/dotnet/docker-tools/pull/430

PR #500, opened 2020-04-28
Add retry logic for ACR cleanup

mthalman:
When running the `cleanAcrImages` command, it can result in there being too many requests to the ACR REST endpoint, resulting in HTTP status 429 errors.  This occurs because there are so many requests occurring in parallel.  To mitigate this, I've added retry logic to attempt retries if this error is encountered.  I've refactored things to make use of the Polly library since it is well-suited for retry logic.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/AcrClient.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/AcrClient.cs
@@ -165,7 +165,12 @@ namespace Microsoft.DotNet.ImageBuilder
         {
             HttpResponseMessage response = await Policy
                 .HandleResult<HttpResponseMessage>(response => response.StatusCode == HttpStatusCode.TooManyRequests)
-                .WaitAndRetryAsync(RetryHelper.MaxRetries, RetryHelper.SleepDurationProvider,
+                .WaitAndRetryAsync(
+                    RetryHelper.MaxRetries,
+                    RetryHelper.ExponentialSleepDurationProvider
+                        .AddJitter(TimeSpan.FromSeconds(5))
+                        .AddOffset(TimeSpan.FromSeconds(5))
+                        .ToFunc(),
                     RetryHelper.GetOnRetryDelegate<HttpResponseMessage>(RetryHelper.MaxRetries, loggerService))
                 .ExecuteAsync(() => httpClient.SendAsync(createMessage()));
 
--- a/src/Microsoft.DotNet.ImageBuilder/src/AcrClient.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/AcrClient.cs
@@ -15,6 +15,7 @@ using Microsoft.DotNet.ImageBuilder.Models.Acr;
 using Microsoft.IdentityModel.Clients.ActiveDirectory;
 using Newtonsoft.Json;
 using Polly;
+using Polly.Contrib.WaitAndRetry;
 
 namespace Microsoft.DotNet.ImageBuilder
 {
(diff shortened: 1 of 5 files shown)


MichaelSimons:
Fixes #497

src/Microsoft.DotNet.ImageBuilder/src/AcrClient.cs
```diff
@@ -154,6 +154,26 @@ private async Task GetPagedResponseAsync<T>(string url, Action<T> onGetResults)
...
+            HttpResponseMessage response = await SendRequestAsync(
+                () => new HttpRequestMessage(HttpMethod.Get, url));
+            return JsonConvert.DeserializeObject<T>(await response.Content.ReadAsStringAsync());
+        }
+
+        private async Task<HttpResponseMessage> SendRequestAsync(Func<HttpRequestMessage> createMessage)
+        {
+            HttpResponseMessage response = await Policy
+                .HandleResult<HttpResponseMessage>(response => response.StatusCode == HttpStatusCode.TooManyRequests)
+                .WaitAndRetryAsync(RetryHelper.MaxRetries, RetryHelper.SleepDurationProvider,
```
MichaelSimons, line 168: Is this the right kind of wait policy for 429s?  Using an exponential policy seems strange at first thought.  If there is a lot of requests, waiting a short time interval the first time feel doesn't seem optimal.  How does this policy work in practice with ACR?  How many waits are needed?

mthalman: The vast majority of retries just retry once.  In some builds, I've seen a handful of requests get retried twice, but never more than that.

mthalman: I did find this: https://github.com/App-vNext/Polly/wiki/Retry#retryafter-when-the-response-specifies-how-long-to-wait.  I'll have to investigate to see whether this REST endpoint returns that header.

src/Microsoft.DotNet.ImageBuilder/src/RetryHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+using Polly;
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    public static class RetryHelper
+    {
+        private const int waitFactor = 5;
+        public const int MaxRetries = 5;
+
+        public static readonly Func<int, TimeSpan> SleepDurationProvider =
```
MichaelSimons, line 15: It may help the understanding of the consuming code as well as support other implementation if the need arises in the future if the name indicated the type of policy - e.g. ExponentialSleepDurationProvider

src/Microsoft.DotNet.ImageBuilder/src/AcrClient.cs
mthalman: Unfortunately, the service doesn't return that header.  So I've updated the code to add an offset of 5 seconds and also added jittering to the generated values to help avoid chunks of requests from all being delayed and re-sent all together again.

src/Microsoft.DotNet.ImageBuilder/src/RetryHelper.cs
```diff
@@ -7,14 +7,35 @@
...
         private const int waitFactor = 5;
         public const int MaxRetries = 5;
 
-        public static readonly Func<int, TimeSpan> SleepDurationProvider =
+        private static readonly Random jitterer = new Random();
+
+        public static SleepDurationProvider ExponentialSleepDurationProvider =
             retryAttempt => TimeSpan.FromSeconds(Math.Pow(waitFactor, retryAttempt - 1));
 
+        public static Func<int, TimeSpan> ExponentialSleepDurationProviderFunc =
```
MichaelSimons, line 22: I noticed Polly provides an [exponential](https://github.com/Polly-Contrib/Polly.Contrib.WaitAndRetry/blob/master/src/Polly.Contrib.WaitAndRetry/Backoff.Exponential.cs) and [jitter](https://github.com/Polly-Contrib/Polly.Contrib.WaitAndRetry/blob/master/src/Polly.Contrib.WaitAndRetry/Backoff.DecorrelatedJitterV2.cs) implementation.  Are they something that could be utilized here instead of maintaining a custom implementation?

mthalman: Excellent

PR #501, opened 2020-04-28
Refactor build job names to remove containing stage name

MichaelSimons:
This eliminates the redundancy in the various build UIs like `Build Build_Linux_arm64v8 3.1-alpine3.11` which shows up in GH PRs.  The shortened name helps reduce the amount of clipping especially in the Azure DevOps pipelines.


MichaelSimons:
Hold on reviewing I just spotted another scenario I need to verify.

PR #505, opened 2020-04-29
Changed linux ImageBuilder image to be based on Alpine vs Debian

MichaelSimons:
This reduces the size of the image from 740 MB to 461 MB (disk size).


MichaelSimons:
@mthalman - I pushed one more change I meant to make earlier.  We should be using the official released version of the manifest tool.  Aside from being the official release vs an RC candidate, there were some performance enhancements made which we can benefit from. Please take another look.

PR #507, opened 2020-04-30
Update image-builder used in builds

MichaelSimons:


mthalman:
@MichaelSimons - Is this needed right away?  There's some changes I need to make to Image Builder that I should be able to get in today.  Can we wait until those are ready before rolling this out?

MichaelSimons:
Nope - not needed.  I will close this PR and let you roll out the changes when you are ready.

PR #510, opened 2020-04-30
Run publish stage in PR builds

### LINKED ISSUE #474: Consider generating image info file in PR build

mthalman:
Updates the pipeline workflow to execute the Post_Build and Publish stages as part of PR builds.  This provides more coverage of the pipeline when validating PRs and should help to expose issues that have otherwise gotten through to internal builds.  

Fixes #474

### DIFF
--- a/eng/common/templates/jobs/publish.yml
+++ b/eng/common/templates/jobs/publish.yml
@@ -5,8 +5,8 @@ jobs:
   pool: ${{ parameters.pool }}
   variables:
     imageBuilder.commonCmdArgs: >
-      --manifest $(manifest)
-      --registry-override $(acr.server)
+      --manifest '$(manifest)'
+      --registry-override '$(acr.server)'
       $(imageBuilder.queueArgs)
     # publicSourceBranch is not necessarily the working branch, it is the branch referenced in the readme Dockerfile source links
     ${{ if contains(variables['Build.SourceBranchName'], 'nightly') }}:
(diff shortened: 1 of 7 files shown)


eng/common/templates/jobs/publish.yml
```diff
@@ -50,9 +56,13 @@ jobs:
       --password $(BotAccount-dotnet-docker-acr-bot-password)
       --os-type '*'
       --architecture '*'
+      $(dryRunArg)
       $(imageBuilder.pathArgs)
       $(imageBuilder.commonCmdArgs)
     displayName: Publish Manifest
+  - publish: $(Build.ArtifactStagingDirectory)/image-info.json
```
MichaelSimons, line 63: Is this done twice now?  Here and https://github.com/dotnet/docker-tools/pull/510/files#diff-c9b89cac83ca08d11b0ea0ef68a5b9acR97.  If so why?

eng/common/templates/variables/common.yml
```diff
@@ -15,6 +15,45 @@ variables:
   value: 2
 - name: imageInfoVariant
   value: ""
+
+  # Default values for variables defined in variable groups
+- name: acr.resourceGroup
```
MichaelSimons, line 20: Would it work to simply quote the references to these variables?  Would that eliminate the need to define these?

For example:
```
      $(runImageBuilderCmd) ingestKustoImageInfo
      $(artifactsPath)/image-info.json
      '$(kusto.cluster)'
      '$(kusto.database)'
      '$(kusto.table)'
      '$(kusto.servicePrincipalName)'
      '$(app-DotnetDockerTelemetryIngestion-client-secret)'
      '$(kusto.servicePrincipalTenant)'
      --manifest $(manifest)
      --os-type '*'
      --architecture '*'
      $(dryRunArg)
```

eng/common/templates/jobs/publish.yml
mthalman: Is that link correct?  It takes me into build-test-publish-repo.yml.

MichaelSimons: That doesn't appear to be the right link.  I am referring to [line 97](https://github.com/dotnet/docker-tools/blob/361d2b0b77174127eed399dceddca61fc9f77574/eng/common/templates/jobs/publish.yml#L97) of this file.

mthalman: Oh, got it.  I think this was the result of a mistake when merging from the tip.  I'll remove the second one.  The reason I moved it up to this point in the pipeline was so that we can access the final result of the image info file even if one of the later tasks fail.  I had run into a scenario where that would have been useful.

MichaelSimons: ah - moving it makes sense.  The duplication is what confused me.

eng/common/templates/variables/common.yml
mthalman: Yes, thanks for the suggestion

PR #512, opened 2020-05-01
Check base image updates: use single job for all archs

### LINKED ISSUE #465: No need for ARM machines in check-base-image-updates pipeline
The [check-base-image-updates](https://github.com/dotnet/docker-tools/blob/master/eng/pipelines/check-base-image-updates.yml) pipeline has build legs which use ARM machines to check for updates to any ARM base images.  But we don't actually need to execute this on an ARM machine.  We can use an AMD64 machine instead just to pull the ARM images and get what we need; these execute much faster.

mthalman:
The pipeline which checks whether base images have been updated is using ARM machines to pull the ARM images.  But this isn't needed because the AMD64 is capable of pulling ARM images that need to be inspected.

Update the pipeline to just use a single Linux and single Windows machine to do its processing.

Fixes #465

### DIFF
--- a/eng/pipelines/check-base-image-updates.yml
+++ b/eng/pipelines/check-base-image-updates.yml
@@ -13,49 +13,15 @@ variables:
 - template: templates/variables/common.yml
 
 jobs:
-- job: Get_Stale_Images_Linux_AMD
+- job: Get_Stale_Images_Linux
   pool: Hosted Ubuntu 1604
   steps:
   - template: templates/steps/get-stale-images.yml
     parameters:
       osType: linux
-      architecture: amd64
-      dockerClientOS: linux
-      staleImagePathsVariableName: linux-amd-stale-image-paths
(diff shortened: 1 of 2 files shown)


mthalman:
> This requires the base image to be architecture specific e.g. not a manifest tag. This works good for our official .NET images. It won't work for the .NET Core samples though will it?

Excellent point.  In that case, I'm having second thoughts on trying to change the current behavior.  IIRC, the samples may have been the whole reason for moving to the architecture-specific builds.  In order to try to work around this, there would need to be a way to input which subscriptions should be checked when calling the `getStaleImages` command.  I'm not sure the extra complexity in that implementation is worth it.

MichaelSimons:
>I'm having second thoughts on trying to change the current behavior.

I am as well.  I don't see the ROI for investing in infra to make this work.

mthalman:
Abandoning these changes based on previous comments.

PR #516, opened 2020-05-05
Fix publicSourceBranch variable for PR builds

### LINKED ISSUE #496: Publish build logic for determining publicSourceBranch does not handle nightly based branches
The following [publish logic](https://github.com/dotnet/docker-tools/blob/master/eng/common/templates/jobs/publish.yml#L11) does not handle feature branches off of nightly.

```
    # publicSourceBranch is not necessarily the working branch, it is the branch referenced in the readme Dockerfile source links
    ${{ if contains(variables['Build.SourceBranchName'], 'nightly') }}:
      publicSourceBranch: nightly
    ${{ if not(contains(variables['Build.SourceBranchName'], 'nightly')) }}:
      publicSourceBranch: master
```

This will cause test builds to fails.  It would be helpful if this logic were more robust.  A couple options come to mind.

1. Reverse the logic so that any branches other than master are considered to be based on nightly.  This assumes it is more common to create feature branches off of nightly.
2. Surface this parameter as something that can be overridden at queue time.

mthalman:
Now that PR builds run the publish stage in the build pipeline, the `publicSourceBranch` variable needs to be set correctly in order to target the correct image info file.  The `Build.SourceBranchName` variable is always set to `merge` for PR builds so that's not useful.  For PRs, the `System.PullRequest.TargetBranch` variable should be used.  In addition, even if the PR is targeting the nightly branch, builds for samples should still use master as the public source branch.

Fixes #496

### DIFF
--- a/eng/common/templates/variables/common.yml
+++ b/eng/common/templates/variables/common.yml
@@ -15,8 +15,6 @@ variables:
   value: 2
 - name: imageInfoVariant
   value: ""
-- name: publishRepoPrefix
-  value: "test/"
 
 - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
   - group: DotNet-Docker-Common
--- a/eng/common/templates/jobs/publish.yml
+++ b/eng/common/templates/jobs/publish.yml
@@ -8,19 +8,14 @@ jobs:
       --manifest '$(manifest)'
       --registry-override '$(acr.server)'
       $(imageBuilder.queueArgs)
-    # publicSourceBranch is not necessarily the working branch, it is the branch referenced in the readme Dockerfile source links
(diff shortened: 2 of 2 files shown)


eng/common/templates/variables/common.yml
```diff
@@ -15,8 +15,6 @@ variables:
   value: 2
 - name: imageInfoVariant
   value: ""
-- name: publishRepoPrefix
```
mthalman, line 18: This is removed because it ends up overriding the value that's explicitly defined by the pipeline in the AzDO variables UI.  By removing this, it means we'll need to explicitly define this value in all the PR pipelines.

PR #517, opened 2020-05-06
Clean image-builder images in ACR

### LINKED ISSUE #504: Clean out old ImageBuilder images
The `cleanAcrImages` command should be updated to remove old versions of ImageBuilder images.  These images are intended for internal purposes only and there's no need for us to retain old versions.

mthalman:
Includes the `dotnet-buildtools/image-builder` repo in the list of images that get cleaned up in ACR.

#Fixes #504

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
@@ -42,7 +42,11 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             List<string> deletedImages = new List<string>();
 
             IEnumerable<Task> cleanupTasks = catalog.RepositoryNames
-                .Where(repoName => IsTestRepo(repoName) || IsStagingRepo(repoName) || IsPublicNightlyRepo(repoName))
+                .Where(repoName =>
+                    IsTestRepo(repoName) ||
+                    IsStagingRepo(repoName) ||
+                    IsPublicNightlyRepo(repoName) ||
+                    IsImageBuilderRepo(repoName))
                 .Select(repoName => acrClient.GetRepositoryAsync(repoName))
                 .Select(getRepoTask => ProcessRepoAsync(acrClient, getRepoTask, deletedRepos, deletedImages))
                 .ToArray();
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -180,6 +189,13 @@ private async Task ProcessStagingRepoAsync(IAcrClient acrClient, List<string> de
             }
         }
 
+        private async Task ProcessImageBuilderRepoAsync(IAcrClient acrClient, List<string> deletedImages, Repository repository)
+        {
+            RepositoryManifests repoManifests = await acrClient.GetRepositoryManifestsAsync(repository.Name);
+            ManifestAttributes[] expiredImages = GetExpiredManifests(repoManifests, 7).ToArray();
```
MichaelSimons, line 195: Does this have the potential to delete all tags?  What if no tags are published for over a week?  Is a better policy here to leave the n most recent builds rather than a time based policy?

mthalman: Yeah, I like that approach.  Fixed

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -180,6 +189,22 @@ private async Task ProcessStagingRepoAsync(IAcrClient acrClient, List<string> de
             }
         }
 
+        private async Task ProcessImageBuilderRepoAsync(IAcrClient acrClient, List<string> deletedImages, Repository repository)
+        {
+            RepositoryManifests repoManifests = await acrClient.GetRepositoryManifestsAsync(repository.Name);
+
+            // Keep the last n number of images, delete everything else older than x days.
```
MichaelSimons, line 196: I like this approach.  My only question is how it would work with feature branches (e.g. image info schema changes - https://github.com/dotnet/docker-tools/pull/463)?  I thought we built images off of the feature branch and used it in a dotnet-docker feature branch until the entire set of changes were complete and flushed out.

mthalman: I can't think of a good solution for this.  There some solutions I can think of but they are either process-heavy that require manual maintenance or complicated engineering logic; none are appealing.

mthalman: Perhaps the better way to handle this is to update the code so that the command can be optionally called to clean up the Image Builder images (it wouldn't clean them up as part of the normal process).  This could then be executed periodically in a manual and targeted way

MichaelSimons: That sounds like a fine proposal to me.

mthalman:
Closing this PR because the new set of changes that are being proposed are going to be significantly different and not take the same approach as described in the PR.

PR #520, opened 2020-05-09
Fix publishMcrDocs to handle different product repos

mthalman:
The `publishMcrDocs` command computes one product family for the entire manifest when publishing.  Now that we have `dotnet/core-nightly` and `dotnet/nightly`, this doesn't work.  It ends up putting all the readmes into a single folder: `teams/dotnet/core-nightly`.  Fixed the logic to compute the product family from each repo name.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
@@ -109,24 +109,31 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         {
             List<GitObject> readmes = new List<GitObject>();
 
-            List<string> readmePaths = new List<string>();
-
             if (!string.IsNullOrEmpty(Manifest.Model.ReadmePath))
             {
-                readmePaths.Add(Manifest.Model.ReadmePath);
+                IEnumerable<string> productRepoNames = Manifest.FilteredRepos
+                    .Select(repo => GetProductRepoName(repo))
+                    .Distinct();
+                foreach (string productRepo in productRepoNames)
+                {
+                    readmes.Add(GetReadMeGitObject(productRepo, Manifest.Model.ReadmePath));
+                }
             }
 
             foreach (RepoInfo repo in Manifest.FilteredRepos)
             {
-                readmePaths.Add(repo.Model.ReadmePath);
-                string fullPath = Path.Combine(Manifest.Directory, repo.Model.ReadmePath);
-
-                string updatedReadMe = File.ReadAllText(fullPath);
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
```diff
@@ -102,47 +100,44 @@ public override async Task ExecuteAsync()
...
 
-        private GitObject[] GetUpdatedReadmes(string productRepo)
+        private GitObject[] GetUpdatedReadmes()
         {
             List<GitObject> readmes = new List<GitObject>();
 
-            List<string> readmePaths = Manifest.FilteredRepos
-                .Select(repo => repo.Model.ReadmePath)
-                .ToList();
+            List<string> readmePaths = new List<string>();
```
MichaelSimons, line 112: Is this needed anymore?

mthalman: The list?  Items are added to it down below.

mthalman: Oh wait, I see what you mean.

mthalman: Fixed this.  I realized that I wasn't handling the primary readme.md file which I've fixed.  My assumption is that the readme.md file needs to be distributed across all the product family repos in the manifest.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs
```diff
@@ -102,47 +100,51 @@ public override async Task ExecuteAsync()
...
-        private GitObject[] GetUpdatedReadmes(string productRepo)
+        private GitObject[] GetUpdatedReadmes()
         {
             List<GitObject> readmes = new List<GitObject>();
 
-            List<string> readmePaths = Manifest.FilteredRepos
-                .Select(repo => repo.Model.ReadmePath)
-                .ToList();
-
             if (!string.IsNullOrEmpty(Manifest.Model.ReadmePath))
```
MichaelSimons, line 112: These changes made me realize the main readme path in the manifest is a little strange now that there are multiple product families in the same manifest.  The code here is introducing a convention of sorts rather than being purely metadata driven.  It works just fine but I wonder if we should either log an issue to clean this up once 3.1 goes EOL or if we should change the ReadmePath to support a collection of readmes.  Either way don't let this hold your checkin.

mthalman: Logged https://github.com/dotnet/dotnet-docker/issues/1902

PR #522, opened 2020-05-13
Remove support for Windows ARM platform

mthalman:
Fixes https://github.com/dotnet/dotnet-docker/issues/1859


eng/pipelines/validate-remote-docker-config.yml
```diff
@@ -33,13 +33,8 @@ jobs:
   - script: $(remoteDockerRunCmd) info
     displayName: Validate Remote Docker Daemon Reachability
   - script: |
-      if [[ "$(DOCKER_CERT_PATH)" == *"humw"* ]] || [[ "$(DOCKER_CERT_PATH)" == *"rpiw"* ]]; then
```
MichaelSimons, line 36: Since these variables are now fixed, they could be declared as normal variables.  The customPinArgs variable could be removed all together and the value could be in-lined.

PR #537, opened 2020-06-03
Add uniqueness to artifact names to fix retry attempts

### LINKED ISSUE #519: Retry of publish job can fail due to duplicate artifact
If the publish job fails, it can be useful to retry the failed job if it's appropriate.  However, a problem exists if the previous failed job had completed the "Publish Image Info File Artifact " task: https://github.com/dotnet/docker-tools/blob/dd3859babc98d5a371708752e0e3b3646a8cf7d6/eng/common/templates/jobs/publish.yml#L70

By completing that task, the pipeline will now contain that artifact name and execution of this task in subsequent retries will fail with an `ArtifactExistsException`.  This is problematic because there are important tasks that need to get executed after this one but they'll never get executed.

Possible solutions:
* Is there a setting that can overwrite the existing pipeline artifact?
* Is there a variable that is set which indicates the retry count that can be used to produce a unique name for the artifact with each retry?
* Can we ignore the error?

mthalman:
A retry of a job that has already published the image info artifact will result in a failure because an artifact with the same name already exists.  Pipeline artifacts cannot be overwritten or deleted.  So the only option is to use a unique name to prevent a conflict.  This is done by appending the `System.JobAttempt` variable to the artifact name.  In the post-build stage, we collect all the image info artifacts and combine them into one image info file.  There needs to be a way to ignore the image info artifacts that were published by failed jobs.  This is done by a script which deletes all the artifacts from those failed jobs.

Fixes #519

### DIFF
--- a/eng/common/templates/jobs/post-build.yml
+++ b/eng/common/templates/jobs/post-build.yml
@@ -7,6 +7,27 @@ jobs:
   - template: ../steps/download-build-artifact.yml
     parameters:
       targetPath: $(Build.ArtifactStagingDirectory)
+  - pwsh: |
+      # Deletes the artifacts from all the unsuccessful jobs
+      Get-ChildItem $(Build.ArtifactStagingDirectory) -Directory |
+          ForEach-Object {
+              [pscustomobject]@{
+                  # Parse the artifact filename to separate the base of the filename from the job attempt number 
+                  BaseName = $_.Name.Substring(0, $_.Name.LastIndexOf('-'));
+                  JobAttempt = $_.Name.Substring($_.Name.LastIndexOf('-') + 1)
+                  FullName = $_.FullName
+              }
+          } |
+          Group-Object BaseName |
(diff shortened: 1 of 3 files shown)


eng/common/templates/jobs/post-build.yml
```diff
@@ -7,6 +7,27 @@ jobs:
...
+                  FullName = $_.FullName
+              }
+          } |
+          Group-Object BaseName |
+          # Delete all but the last artifact from each base filename
+          ForEach-Object {
+              $_.Group |
+                  Sort-Object JobAttempt |
+                  Select-Object -ExpandProperty FullName -SkipLast 1 |
+                  #Remove-Item -Recurse -Force
```
MichaelSimons, line 27: Was commented out for debugging?

mthalman: Yes, fixed

PR #541, opened 2020-06-05
Update ACR cleanup to be driven from cmd options

mthalman:
In order to better allow for more ad-hoc deletion of images in ACR, these changes update the `cleanAcrImages` command to specify what and how the deletions should be made by using the command parameters instead of having repo-specific logic embedded in the command.

Related to #504

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.CommandLine;
 
 namespace Microsoft.DotNet.ImageBuilder.Commands
@@ -10,6 +11,9 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     {
         protected override string CommandHelp => "Removes unnecessary images from an ACR";
 
+        public string RepoName { get; set; }
+        public CleanAcrImagesAction Action { get; set; }
+        public int DaysOld { get; set; }
         public string Username { get; set; }
         public string Password { get; set; }
         public string Tenant { get; set; }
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/ValidateImageSizeOptions.cs
```diff
@@ -21,8 +21,10 @@ public override void DefineOptions(ArgumentSyntax syntax)
         {
             base.DefineOptions(syntax);
 
-            string mode = ImageSizeValidationMode.All.ToString().ToLowerInvariant();
-            syntax.DefineOption("mode", ref mode, "Mode of validation. Options: all (default), size, integrity");
+            ImageSizeValidationMode defaultValidationMode = ImageSizeValidationMode.All;
```
MichaelSimons, line 24: I noticed there is a bit on inconsistency in the patterns used to handle enum options.  The patterns used by the GenerateBuildMatrixOptions is a bit nicer because the option is strongly typed.

            MatrixType matrixType = MatrixType.PlatformDependencyGraph;
            syntax.DefineOption(
                "type",
                ref matrixType,
                value => (MatrixType)Enum.Parse(typeof(MatrixType), value, true),
                $"Type of matrix to generate. {EnumHelper.GetHelpTextOptions(matrixType)}");
            MatrixType = matrixType;

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
```diff
@@ -45,5 +53,38 @@ public override void DefineParameters(ArgumentSyntax syntax)
...
+            base.DefineOptions(syntax);
+
+            const CleanAcrImagesAction defaultAction = CleanAcrImagesAction.PruneDangling;
+            string action = defaultAction.ToString().ToCamelCase();
+            syntax.DefineOption("action", ref action,
+                $"Type of delete action. {EnumHelper.GetHelpTextOptions(defaultAction)}");
+            Action = (CleanAcrImagesAction)Enum.Parse(typeof(CleanAcrImagesAction), action, ignoreCase: true);
+
+            int daysOld = 30;
+            syntax.DefineOption("days", ref action, "Number of days old to be considered for deletion");
```
MichaelSimons, line 68: The wording of the description was a bit awkward to me on first read.  What do you think about using "age" nomenclature instead?

```suggestion
            syntax.DefineOption("age", ref age, "Minumum age (days) of images to apply delete action to");
```

Note: the original code has the wrong `ref` variable passed - `action`

PR #546, opened 2020-06-09
Support for multiple os-version options

### LINKED ISSUE #545: Regression from removal of osVersion parameter
The changes made in #540 caused a regression for the building of sample images.  The symptom of this is that Windows version-specific build jobs end up building sample images for all of the Windows version, instead of just the version that the build job was intended to build.

Sample images are unique compared to the official product images because a single Dockerfile can be used to build multiple versions.  This means that the `imageBuilderPaths` property produced by the matrix generation is not unique between jobs.  The assumption that this would be unique across jobs was the reason that the changes in #540 were made.  So it turns out that `osVersion` is necessary in order to produce a unique set of images across jobs.

The changes from #540 need to be reverted and replaced with another implementation that allows `osVersion` to remain, or at least allow it to remain for certain scenarios.

mthalman:
* Adds back the `os-version` parameter when running the `build` command.
* Fixes the value of the osVersion field in the build matrix to not be the normalized value.
* Adds support for multiple `os-version` options.  This supports cases where two different OS versions need to be built in the same job (e.g. Nano Server 1909 and Windows Server Core 1909).

Once a new version of Image Builder is available with these changes, another change will be made to consume that new version and update the pipeline to use the multiple `os-version` options.

Fixes #545

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
@@ -60,6 +60,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 matrix.Legs.Add(leg);
 
                 AddImageBuilderPathsVariable(dockerfilePaths, leg);
+                AddOsVersionsVariable(subgraph, leg);
                 AddCommonVariables(platformGrouping, leg);
             }
         }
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestFilter.cs
```diff
@@ -54,9 +54,9 @@ public IEnumerable<Platform> GetPlatforms(Image image)
                     Regex.IsMatch(platform.Dockerfile, pathsRegexPattern, RegexOptions.IgnoreCase));
             }
 
-            if (IncludeOsVersion != null)
+            if (IncludeOsVersions?.Any() ?? false)
             {
-                string includeOsVersionPattern = GetFilterRegexPattern(IncludeOsVersion);
+                string includeOsVersionPattern = GetFilterRegexPattern(IncludeOsVersions.ToArray());
```
MichaelSimons, line 59:
```suggestion
                string includeOsVersionsPattern = GetFilterRegexPattern(IncludeOsVersions.ToArray());
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -60,6 +60,7 @@ public override Task ExecuteAsync()
                 matrix.Legs.Add(leg);
 
                 AddImageBuilderPathsVariable(dockerfilePaths, leg);
+                AddOsVersionsVariable(subgraph, leg);
```
MichaelSimons, line 63: It feels like this should be part of the AddCommonVariables method since it is a common variable across the supported matrix types.

Issue #486, opened 2020-04-20
Use aarch64 version of AzDO agent


mthalman:
The current infrastructure for supporting Linux ARM builds has been to use a Linux AMD64 agent machine that sends remote Docker commands to an aarch64 machine.  This was necessary because the AzDO agent software didn't support aarch64.  But now that a release candidate is available for this support (https://github.com/microsoft/azure-pipelines-agent/issues/1911), we can update the infrastructure so that the Linux AMD64 agent machine goes away and the aarch64 machine directly runs the agent software.

- [x] Publish an ARM version of [Image Builder](https://github.com/dotnet/docker-tools/blob/master/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux) with a multi-arch tag
- [x] Publish an ARM version of the [Test Runner](https://github.com/dotnet/docker-tools/blob/master/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux.runner) image with a multi-arch tag
- [x] Update common pipeline logic to consume multi-arch tags for Image Builder and Test Runner
- [x] Update common pipeline logic to conditionally support ARM agents
- [x] Install agent software on public ARM machines
- [x] Disable use of remote Docker server in common pipeline logic in nightly branch
- [x] Test for a period of time to ensure builds are working
- [ ] Install agent software on internal ARM machines
- [ ] Disable use of remote Docker server in common pipeline logic in master branch
- [ ] De-allocate AMD build agents used w/ARM machines

MichaelSimons:
We should do this.  We should be scouting and providing feedback.  We should role this out to CI first to ensure stability.

mthalman:
I've sent an email to DDFUN to get this process started.

The proposed process is to start by rolling this out to a single ARM machine in the public CI pool.  That will determine whether it's even possible to have something working.  If that works, we'll roll it out to the rest of the ARM machines in the public pool.  Once we've determined that things are stable there, we'll roll it out to the internal build machines.

MichaelSimons:
@mthalman, Any update on the roll out?

mthalman:
It's been on hold for a few weeks due to other higher priority work.  I've verified that the ARM version of the agent software does in fact work; I was able to run builds directly on the ARM machine.  The next step is to start producing ARM versions of Image Builder since it'll now need to be executed directly on ARM hardware.

mthalman:
[...]

PR #553, opened 2020-06-12
Publish Linux ARM version of Image Builder

mthalman:
Updates the pipeline that builds Image Builder to use the common infrastructure for building images.  This provides a means to easily build/publish a Linux ARM64 version of Image Builder as well as publish a multi-arch tag.

Using the common infrastructure means that there needs to be a manifest to describe the metadata of the Image Builder images.  Since we want to generate unique tags for each build of the images, a `UniqueId` variable used to for the shared tag.  This is used instead of the built-in `System.TimeStamp` variable because it needs to be the same value across all the build jobs.  The value of the variable is based off the `sourceBuildId` variable of the pipeline.

The Linux Dockerfile has parameterized with `ARG` values so that it can be shared by both AMD64 and ARM64 builds.

Once an official ARM version of Image Builder has been published, the infrastructure will be updated in a subsequent PR to reference the multi-arch tag.

Related to #486


eng/pipelines/templates/variables/image-builder.yml
```diff
@@ -0,0 +1,10 @@
+variables:
+- template: common.yml
+- name: manifest
+  value: src/Microsoft.DotNet.ImageBuilder/manifest.json
+- name: publishReadme
+  value: false
+- name: manifestVariables
+  value: --var UniqueId=$(sourceBuildId)
+- name: imageInfoVariant
+  value: "-imagebuilder"
```
MichaelSimons, line 10: missing EOF

src/Microsoft.DotNet.ImageBuilder/manifest.json
```diff
@@ -0,0 +1,51 @@
...
+      "name": "dotnet-buildtools/image-builder",
+      "images": [
+        {
+          "sharedTags": {
+            "$(UniqueId)": {}
+          },
+          "platforms": [
+            {
+              "buildArgs": {
+                "ALPINE_TAG_SUFFIX": ""
```
MichaelSimons, line 14: Why is just one of three builds args being explicitly defined here?  It feels a bit inconsistent - either utilize the defaults or be explicit.

eng/common/templates/jobs/build-images.yml
```diff
@@ -36,8 +36,9 @@ jobs:
       --architecture $(architecture)
       --retry
       --source-repo $(publicGitRepoUri)
-      $(imageBuilderBuildArgs)
       $(imageBuilderImageInfoArg)
+      $(manifestVariables)
```
MichaelSimons, line 40: Seeing the number of steps this needed to be added to makes me think that the `init` steps should dynamically define a `commonImageBuilderArgs`.  It would make it easier to make these types of changes in the future and possibly help simplify the ImageBuilder command steps.  I'm not necessarily advocating for a change in this PR but something to consider.

src/Microsoft.DotNet.ImageBuilder/manifest.json
mthalman: It's needed because of how Image Builder works.  When loading the manifest, it reads the `FROM` lines in the Dockerfiles and attempts to replace any variables.  If the build arg isn't defined here, Image Builder throws an exception that the variable doesn't exist.  I've added the other variables for consistency.

PR #555, opened 2020-06-15
Update Clean ACR image pipelines

### LINKED ISSUE #504: Clean out old ImageBuilder images

mthalman:
Updates the pipelines that cleanup ACR images to consume the updated command API from https://github.com/dotnet/docker-tools/pull/541.  This requires defining which images are to be cleaned up through the command parameters instead of defined within the command logic itself.  In order to provide a means to cleanup images in an ad hoc fashion (such as Image Builder images, #504), an additional pipeline is added for executing custom cleanup operations where the parameters can be defined at queue time.

Fixes #504

### DIFF
--- a/eng/pipelines/cleanup-acr-images-custom.yml
+++ b/eng/pipelines/cleanup-acr-images-custom.yml
@@ -1,0 +1,18 @@
+trigger: none
+pr: none
+
+variables:
+- template: ../common/templates/variables/common.yml
+
+jobs:
+- job: Build
+  pool: Hosted Ubuntu 1604
+  steps:
+  - template: ../common/templates/steps/init-docker-linux.yml
+  - template: templates/steps/clean-acr-images.yml
+    parameters:
+      repo: $(repo)
+      action: $(action)
+      age: $(age)
+      customArgs: $(customArgs)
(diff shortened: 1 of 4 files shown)


eng/pipelines/cleanup-acr-images-custom.yml
```diff
@@ -0,0 +1,18 @@
...
+- job: Build
+  pool: Hosted Ubuntu 1604
+  steps:
+  - template: ../common/templates/steps/init-docker-linux.yml
+  - template: templates/steps/clean-acr-images.yml
+    parameters:
+      repo: $(repo)
+      action: $(action)
+      age: $(age)
+      customArgs: $(customArgs)
```
MichaelSimons, line 17: Thoughts on defaulting this to `--dry-run` in the modeled build definition?  It feels like a good practice to run a dry-run first to help ensure the specified parameters are correct when running on a one-off basis.

mthalman: Yep, that's my intention.  👍

PR #557, opened 2020-06-15
Skip image info initialization when not needed

mthalman:
The recent change in https://github.com/dotnet/docker-tools/pull/553/files#diff-f50a5e9dc445f037cb667a4130a220c6 has caused issues in some pipelines which use the init-docker templates but don't run the generate build matrix template.  When run in Windows, an undefined variable reference causes an error.  The undefined variable in this case is `legName` which is set as part of the matrix generation.  This task should only be run for those pipelines that are making use of a matrix.  Updated the templates to provide a parameter that allows pipelines to disable this task.

### DIFF
--- a/eng/common/templates/steps/init-docker-linux.yml
+++ b/eng/common/templates/steps/init-docker-linux.yml
@@ -3,15 +3,10 @@ parameters:
   setupImageBuilder: true
   setupTestRunner: false
   cleanupDocker: true
-  useImageInfo: true
 
 steps:
 - script: echo "##vso[task.setvariable variable=artifactsPath]/artifacts"
   displayName: Define Artifacts Path Variable
-
-- ${{ if eq(parameters.useImageInfo, 'true') }}:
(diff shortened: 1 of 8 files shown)


eng/common/templates/steps/init-docker-linux.yml
```diff
@@ -3,13 +3,15 @@ parameters:
...
   cleanupDocker: true
+  useImageInfo: true
 
 steps:
 - script: echo "##vso[task.setvariable variable=artifactsPath]/artifacts"
   displayName: Define Artifacts Path Variable
 
-- script: echo '##vso[task.setvariable variable=imageBuilderImageInfoArg]--image-info-output-path $(artifactsPath)/$(legName)-image-info.json'
-  displayName: Set Image Info Arg for Image Builder
+- ${{ if eq(parameters.useImageInfo, 'true') }}:
```
MichaelSimons, line 12: Have you considered moving this outside of init-docker?  It looks like this variable is only used for the build job.  The logic looks to be the same for both windows and linux.  I think the overall code would be simpler.

mthalman: Yep, much simpler that way.  Fixed

PR #565, opened 2020-06-18
Update pipelines to use ubuntu-latest VM image

### LINKED ISSUE #538: Investigate using Ubuntu 18.04 for Linux builds
Currently the Linux build legs are configured to use [Ubuntu 16.04](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1604-README.md) as the VM image.  I've read some articles stating that there are scenarios where performance is significantly better using Ubuntu 18.04 compared to 16.04.  We should do some testing using the [18.04](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md) VM image and see how it compares.

mthalman:
Fixes #538

### DIFF
--- a/eng/common/templates/variables/common.yml
+++ b/eng/common/templates/variables/common.yml
@@ -17,7 +17,7 @@ variables:
   value: ""
 - name: manifestVariables
   value: ""
-- name: defaultAmd64PoolImage
+- name: defaultLinuxAmd64PoolImage
   value: ubuntu-latest
 
 - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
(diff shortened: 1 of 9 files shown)


eng/common/templates/variables/common.yml
```diff
@@ -17,6 +17,8 @@ variables:
   value: ""
 - name: manifestVariables
   value: ""
+- name: defaultAmd64PoolImage
```
MichaelSimons, line 20: I feel like the name should call out the platform.  This helps the readability of the various pipelines for getting an understanding if running on linux vs windows.

```suggestion
- name: defaultLinuxAmd64PoolImage
```

PR #567, opened 2020-06-24
Add test script for Image Builder

mthalman:
The PR build for Image Builder is failing because the build stage is attempting to run a test script but none exists.  Adding a test script to run the Image Builder tests.  This ends up building the Image Builder project again outside the context of a Docker build but it's very low overhead.  Since the tests can be run independently, the Dockerfiles have been updated to no longer run tests.

Also moved the `test` folder from the root of the repo to the `eng` folder.


src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -12,16 +12,13 @@ ARG RID_ARCH=x64
...
 COPY src/Microsoft.DotNet.ImageBuilder.csproj ./src/
-COPY tests/Microsoft.DotNet.ImageBuilder.Tests.csproj ./tests/
-RUN dotnet restore Microsoft.DotNet.ImageBuilder.sln
+RUN dotnet restore ./src/Microsoft.DotNet.ImageBuilder.csproj
 
 # copy everything else and build
 COPY . ./
-RUN dotnet build Microsoft.DotNet.ImageBuilder.sln
-RUN dotnet test tests/Microsoft.DotNet.ImageBuilder.Tests.csproj
+RUN dotnet build ./src/Microsoft.DotNet.ImageBuilder.csproj
```
MichaelSimons, line 21: Explicitly building doesn't seem necessary anymore and should be removed for simplicity.  This comment applies to the Windows Dockerfile as well.

mthalman: Fixed

Issue #547, opened 2020-06-10
Consume MCR Status Flow API


mthalman:
MCR now has a REST API that can be used to check the status of image and readme ingestion.  By querying this API, this can allow the publishing stage of the pipeline to be able to confirm whether MCR has successfully ingested the images and readmes.  If they were not successfully ingested, the build should fail.  This will greatly improve the accuracy of when and whether a publish has completed.

PR #568, opened 2020-06-24
Access token and locking refactoring

mthalman:
Updates the Image Builder code in support of https://github.com/dotnet/docker-tools/issues/547.  Consuming the MCR status API will require a long-running operation that could result in an expired access token.  These changes start the process of defining a common pattern for handling expired access tokens, starting with the `AcrClient` code.  This also resulted in defining some common locking logic that is used across the project.


src/Microsoft.DotNet.ImageBuilder/src/AuthHelper.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public static async Task<string> GetAadAccessTokenAsync(string resource, string tenant, string username, string password)
+        {
+            AuthenticationContext authContext = new AuthenticationContext($"https://login.microsoftonline.com/{tenant}");
+            AuthenticationResult result = await authContext.AcquireTokenAsync(
+                resource, new ClientCredential(username, password));
+            return result.AccessToken;
+        }
+    }
+}
```
MichaelSimons, line 20: Missing EOF

PR #574, opened 2020-07-09
Wait for MCR image/doc ingestion in pipeline

mthalman:
Updates the publish stage of the pipeline to wait for MCR image and doc ingestion after they've been published.  This is an opt-in model that will be enabled by setting the `waitForIngestionEnabled` variable to `true`.  Also included is a job template that can be used by each repo to wait for ingestion as a standalone pipeline.

Related to #547

### DIFF
--- a/eng/common/templates/jobs/wait-for-ingestion.yml
+++ b/eng/common/templates/jobs/wait-for-ingestion.yml
@@ -1,5 +1,5 @@
 jobs:
-- job: Build
+- job: WaitForIngestion
   pool:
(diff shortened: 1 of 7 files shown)


eng/common/templates/jobs/wait-for-ingestion.yml
```diff
@@ -0,0 +1,27 @@
+jobs:
+- job: Build
```
MichaelSimons, line 2: The name of the job is strange to me.  Why not something like "WaitForIngestions"?

PR #575, opened 2020-07-09
Add 'import' concept to manifests

MichaelSimons:
This PR adds an `import' concept to the manifest.  This allows the manifest.json to be split across multiple files.  This is being added to support the [Dockerfile templates feature](https://github.com/dotnet/dotnet-docker/issues/1933).

The initial implementation only supports variables.  This is a scoping decision based on what is currently required.  There is nothing that prevents future expansion to support other model types.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
@@ -125,14 +125,14 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
                 foreach (PlatformInfo platform in platforms)
                 {
-                    pathsToRebuild.AddRange(await GetPathsToRebuildAsync(allPlatforms, platform, repoData));
+                    pathsToRebuild.AddRange(GetPathsToRebuild(allPlatforms, platform, repoData));
                 }
             }
 
             return pathsToRebuild.Distinct().ToList();
         }
 
-        private async Task<List<string>> GetPathsToRebuildAsync(
+        private List<string> GetPathsToRebuild(
             IEnumerable<PlatformInfo> allPlatforms, PlatformInfo platform, RepoData repoData)
         {
             bool foundImageInfo = false;
--- a/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
@@ -55,9 +55,8 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
 
         private static ManifestInfo Create(string manifestPath, ManifestFilter manifestFilter, IManifestOptionsInfo options)
         {
-            string manifestJson = File.ReadAllText(manifestPath);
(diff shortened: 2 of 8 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs
```diff
@@ -125,14 +125,14 @@ private async Task<IEnumerable<string>> GetPathsToRebuildAsync(Subscription subs
...
-                    pathsToRebuild.AddRange(await GetPathsToRebuildAsync(allPlatforms, platform, repoData));
+                    pathsToRebuild.AddRange(GetPathsToRebuild(allPlatforms, platform, repoData));
                 }
             }
 
             return pathsToRebuild.Distinct().ToList();
         }
 
-        private async Task<List<string>> GetPathsToRebuildAsync(
+        private List<string> GetPathsToRebuild(
```
MichaelSimons, line 135: This change is unrelated.  It was made in order to remove a build warning.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
```diff
@@ -168,5 +167,35 @@ public RepoInfo GetRepoByModelName(string name)
...
+                {
+                    ModelExtensions.ValidateFileReference(includePath, manifestDirectory);
+                    manifestJson = File.ReadAllText(Path.Combine(manifestDirectory, includePath));
+                    Manifest includeModel = JsonConvert.DeserializeObject<Manifest>(manifestJson);
+                    foreach (KeyValuePair<string, string> kvp in includeModel.Variables)
+                    {
+                        if (model.Variables.ContainsKey(kvp.Key))
+                        {
+                            throw new InvalidOperationException(
+                                $"The manifest contains multiple '{kvp.Key}' variables.  Each variable name must be unique.");
```
mthalman, line 190: Include in the exception message a reference to the `includePath`.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -40,6 +40,11 @@ public static string GetDisplayName(this Architecture architecture, string varia
 
         public static void Validate(this Manifest manifest, string manifestDirectory)
         {
+            if (manifest.Repos == null)
```
mthalman, line 43: Shouldn't this technically also check that at least one repo exists in the set?

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
MichaelSimons: I opted not to do this because I didn't want to track where the other "duplicate" variable was coming from.  Only listing one of the json files seemed of limited value.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
MichaelSimons: Good point, fixed.

MichaelSimons:
FYI - I had to make one small teak to support a manifest that doesn't have any variables but defined an import.

PR #584, opened 2020-07-15
Handle undocumented platform tag generation

mthalman:
If a platform defined in the manifest has no documented tags of its own but is defined within an image that has shared tags, tag generation will fail here: https://github.com/dotnet/docker-tools/blob/92ef21737f934add62745a505a3d795e88313325/src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs#L62

This happens because the `ImageDocumentationInfo` objects that were generated include the platform that has no documented platform tags and it never gets removed from the set because there isn't a readme template that references one of its tags.

This is fixed by only generating an `ImageDocumentationInfo` object if a platform has any documented _platform_ tags.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs
@@ -44,7 +44,7 @@ namespace Microsoft.DotNet.ImageBuilder
             _imageDocInfos = _repo.FilteredImages
                 .SelectMany(image =>
                     image.AllPlatforms.SelectMany(platform => ImageDocumentationInfo.Create(image, platform)))
-                .Where(info => info.DocumentedTags.Any())
+                .Where(info => info.DocumentedPlatformTags.Any())
                 .ToList();
 
             StringBuilder yaml = new StringBuilder();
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs
```diff
@@ -126,15 +126,19 @@ private string GetVariableValue(string variableType, string variableName)
...
+            public IEnumerable<TagInfo> DocumentedTags { get; }
+            public string FormattedDocumentedTags { get; }
             public PlatformInfo Platform { get; }
 
             private ImageDocumentationInfo(ImageInfo image, PlatformInfo platform, string documentationGroup)
             {
                 Platform = platform;
-                DocumentedTags = GetDocumentedTags(Platform.Tags, documentationGroup)
-                    .Concat(GetDocumentedTags(image.SharedTags, documentationGroup))
+                DocumentedSharedTags = GetDocumentedTags(image.SharedTags, documentationGroup).ToArray();
```
MichaelSimons, line 138: Did you consider an approach where you skip the instantiation of the ImageDocumentationInfo in the case where there are not documented platform tags?  It feels more efficient and centralize the logic more for determining which images are documented or not.

mthalman: The logic has been updated where this isn't relevant anymore.

mthalman:
> A thought I had would be to change the IsDocumented property from a bool to an DocumentationType Enum. Documented(default),Undocumented,PlatformDocumented. The last one would be used on SharedTags for this situation.

I like this suggestion.  Updated the code to go with this approach.

Issue #592, opened 2020-07-23
Add validation in build command to prevent inadvertent pulls


mthalman:
There have been cases where bugs in the infrastructure have resulted in the implicit pulling of images during a `docker build` command.

Example from https://github.com/dotnet/dotnet-docker/issues/2090:
```
-- EXECUTING: docker build -t mcr.microsoft.com/dotnet/core-nightly/aspnet:3.1.6-alpine3.12 -t mcr.microsoft.com/dotnet/core-nightly/aspnet:3.1-alpine3.12 -t mcr.microsoft.com/dotnet/core-nightly/aspnet:3.1.6-alpine -t mcr.microsoft.com/dotnet/core-nightly/aspnet:3.1-alpine -f src/aspnet/3.1/alpine3.12/amd64/Dockerfile --build-arg REPO=mcr.microsoft.com/dotnet/core-nightly/runtime src/aspnet/3.1/alpine3.12/amd64
Sending build context to Docker daemon   2.56kB

Step 1/3 : ARG REPO=mcr.microsoft.com/dotnet/core/runtime
Step 2/3 : FROM $REPO:3.1-alpine3.12
3.1-alpine3.12: Pulling from dotnet/core-nightly/runtime
df20fa9351a1: Already exists
ed937d6b1c19: Pulling fs layer
ab29de8d4074: Pulling fs layer
ed937d6b1c19: Verifying Checksum
ed937d6b1c19: Download complete
ed937d6b1c19: Pull complete
ab29de8d4074: Verifying Checksum
ab29de8d4074: Download complete
ab29de8d4074: Pull complete
Digest: sha256:42e51005c7ed9d42a193744d116b6e81bb68faef1e3afd1a39530fdd60aa8db1
Status: Downloaded newer image for mcr.microsoft.com/dotnet/core-nightly/runtime:3.1-alpine3.12
 ---> c255a55d06e7
```

A pull should never occur implicitly when we run a `docker build` command since our intent is to have all dependent images pre-pulled before building.  The impact of this is that the image gets built based on an unintended source.  Even more problematic is that there's nothing that alerts us of this problem.  The image gets pulled and as long as the build and tests pass, we don't notice.  But this kind of behavior is not correct and can lead to very wrong results.

By adding logic to Image Builder's `build` command, we can parse the standard output and look for any occurrences of `Pulling from` in the output and throw an exception in that case.  This will ensure we are aware of the issue right away.

[...]

PR #593, opened 2020-07-24
Validate build doesn't cause an inadvertent pull

mthalman:
Related to #592

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -161,13 +161,21 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                                 .Select(tag => tag.FullyQualifiedName)
                                 .ToList();
 
-                            this.dockerService.BuildImage(
+                            string buildOutput = this.dockerService.BuildImage(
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -161,14 +161,22 @@ private void BuildImages()
...
-                            this.dockerService.BuildImage(
+                            string buildOutput = this.dockerService.BuildImage(
                                 dockerfilePath,
                                 platform.BuildContextPath,
                                 allTags,
                                 platform.BuildArgs,
                                 Options.IsRetryEnabled,
                                 Options.IsDryRun);
 
+                            if (buildOutput.Contains("Pulling from"))
```
MichaelSimons, line 172: Is this the correct behavior if the `skip-pulling` option is specified?  I utilize this option frequently during my local dev process when iterating repeatedly.  As it behaves now, the documentation is a little misleading as it implies implicit pulling still would happen - `Skip explicitly pulling the base images of the Dockerfiles`.

mthalman: Good point.  Updated to check for that option.

PR #594, opened 2020-07-24
Fix platformVersionedOs matrix generation to not be based on product version

mthalman:
Matrix generation for the platformVersionedOs matrix type currently groups together platforms that have the same product version and OS version.  This is problematic now that the monitor image has been introduced to the dotnet-docker repo because it has a product version that matches .NET but is versioned independently, so they're not related and should not be grouped together.

This is fixed by removing this kind of grouping dependency based on product and OS version.  Instead, dependencies are based on the actual Dockerfile `FROM` references.  Those Dockerfiles that reference others are grouped.  In addition to the `FROM` references, there is a need to make use of custom build leg dependencies in order to describe images that need to be grouped as part of the platformVersionedOs matrix generation but do not have "physical" references between images in the Dockerfiles.  Examples of this include the 2.1 SDK images which are based on the base OS images, not on the corresponding ASP.NET images.  By defining a custom build leg dependency in the manifest, the algorithm will take such a dependency into account when constructing the grouping.

There was also a need to add more detail to the custom build leg dependencies because there needs to be different handling for the scenarios described above versus the scenario that first introduced the concept of custom build leg dependencies.  The original need for these was to support scenarios where another image from a different matrix leg was needed into order to test an image (e.g. 2.1 Buster ARM SDK defined as a custom build leg dependency in order to test Alpine ARM runtime).  In such a scenario, the image gets included in two legs.  But in the scenario for this PR, the image should only be included in one leg.  For that reason, a dependency type has been included in the manifest model.

Related to dotnet/dotnet-docker#2090


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixOptions.cs
```diff
@@ -14,7 +13,7 @@ public class GenerateBuildMatrixOptions : ManifestOptions, IFilterableOptions
 
         public ManifestFilterOptions FilterOptions { get; } = new ManifestFilterOptions();
         public MatrixType MatrixType { get; set; }
-        public string CustomBuildLegGrouping { get; set; }
+        public string CustomBuildLegGroup { get; set; }
```
MichaelSimons, line 16: I was expecting this would need to be a collection.  'dotnet-docker' will need multiple groupings.

1. A group that specifies the existing product grouping for aspnet, sdk, and runtime for the older versions where image dependencies don't exist.
2. The existing PR groupings for the arm SDK scenario

The PR builds would need to run with both groupings and official builds would only run with the first.

src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest/CustomBuildLegGroup.cs
```diff
@@ -14,18 +14,21 @@ namespace Microsoft.DotNet.ImageBuilder.Models.Manifest
...
+            " custom label that can then be used by tooling to lookup the group when necessary."
+            )]
         [JsonProperty(Required = Required.Always)]
         public string Name { get; set; }
 
-        [Description("The set of image tags that this image is dependent upon for this scenario.")]
+        [Description("The set of dependencies the image has for this scenario.")]
         [JsonProperty(Required = Required.Always)]
-        public string[] Dependencies { get; set; }
+        public CustomBuildLegDependency[] Dependencies { get; set; }
```
MichaelSimons, line 32: Defining the `CustomBuildLegDependency` provides a little more flexibility in the schema but it does come at the cost of verbosity for the way it will be used in practice for our repos.  My understanding is that each grouping will be tied to a specific type therefore the type could be moved up to the grouping.  I would prefer keeping the manifest less verbose

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixOptions.cs
mthalman: Fixed

src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest/CustomBuildLegGroup.cs
mthalman: Fixed

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixOptions.cs
```diff
@@ -14,7 +15,7 @@ public class GenerateBuildMatrixOptions : ManifestOptions, IFilterableOptions
 
         public ManifestFilterOptions FilterOptions { get; } = new ManifestFilterOptions();
         public MatrixType MatrixType { get; set; }
-        public string CustomBuildLegGrouping { get; set; }
+        public IEnumerable<string> CustomBuildLegGroups { get; set; } = Enumerable.Empty<string>();
```
MichaelSimons, line 18: What is the value in defaulting the property here when a default is specified in the DefineOptions method?

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -65,23 +65,24 @@ public override Task ExecuteAsync()
         }
 
         private IEnumerable<IEnumerable<PlatformInfo>> GetCustomLegGroupingDependencies(
-            IEnumerable<PlatformInfo> subgraph, IEnumerable<IEnumerable<PlatformInfo>> subgraphs)
+            IEnumerable<PlatformInfo> subgraph, IEnumerable<IEnumerable<PlatformInfo>> subgraphs, CustomBuildLegDependencyType? dependencyType = null)
```
MichaelSimons, line 68: I'm not seeing where is the dependencyType optional parameter used?

src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest/CustomBuildLegDependencyType.cs
```diff
@@ -0,0 +1,34 @@
...
+            "have its own leg generated."
+            )]
+        Integral,
+
+        [Description(
+            "Indicates the dependency is considered to be a supplemental companion to the depending image." +
+            "This means the dependent image will have its own dependency graph considered for build leg " +
+            "generation. An example of this is when a custom build leg dependency is defined to " +
+            "include an SDK image supported on a particular architecture in order to test a runtime OS " +
+            "that doesn't its own SDK on that architecture (Buster ARM SDK to test Alpine ARM runtime); " +
```
MichaelSimons, line 28:
```suggestion
            "that doesn't have its own SDK on that architecture (Buster ARM SDK to test Alpine ARM runtime); " +
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
mthalman: It's not. Removed.

PR #600, opened 2020-08-13
Store digests of internal FROM images in image info

mthalman:
In order to be able to implement the build caching logic documented in #186, there needs to be a way to look up the digest of the base image for each of the published images.  Currently this is tracked in the image info file but only for base images that are external (those which we do not build); we don't track the digests of base images that have been built internally from our own Dockerfiles.  These changes add logic to be able to track these digests.

It's not as simple as getting the digest after having pushed the images because the images are pushed to a staging location and so the digest would be based off of that registry/repo name.  Instead, we need to manually construct what the eventual registry/repo name is and append it to the SHA value to get the digest value that we actually want to store.

For consistency, I've updated the logic for all digest values stored in the image info file to be based on this scheme instead of only storing the SHA value.

Related to #186

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -84,56 +84,70 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 {
                     if (Options.IsPushEnabled)
                     {
-                        // The digest of an image that is pushed to ACR is guaranteed to be the same when transferred to MCR
-                        // It is output in the form of <repo>@<sha> but we only want the sha because the repo portion is tied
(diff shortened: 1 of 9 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -66,16 +66,29 @@ private void PublishImageInfo()
                 throw new InvalidOperationException("Source repo URL must be provided when outputting to an image info file.");
             }
```
MichaelSimons, line 68: This PublishImageInfo method is getting pretty big.  IMO the readability would be improved if it was refactored to pull out some of the logic into helper methods.

mthalman: Split it out into a few extra methods.

PR #604, opened 2020-08-14
Handle null digest value

mthalman:
The changes from https://github.com/dotnet/docker-tools/pull/600 cause an issue when running dry-runs in the publish stage of PR builds.  The `ingestKustoImageInfo` command will encounter a NRE when attempting to extract the SHA from a digest value of the generated image info.  The digest value will be null in such a dry-run build because no image was pushed.  Fixed the code to gracefully handle a null digest value.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
@@ -129,16 +129,8 @@ namespace Microsoft.DotNet.ImageBuilder
                 "inspect", ".Created", "Failed to retrieve created date", image, isDryRun);
         }
 
-        public static string GetDigestSha(string digest)
-        {
-            if (digest is null)
-            {
-                return null;
-            }
-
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
```diff
@@ -129,7 +129,15 @@ public static string GetCreatedDate(string image, bool isDryRun)
                 "inspect", ".Created", "Failed to retrieve created date", image, isDryRun);
         }
 
-        public static string GetDigestSha(string digest) => digest.Substring(digest.IndexOf("@") + 1);
```
MichaelSimons, line 132: Have you considered using the null conditional operator?  It feels like is would simplify the code.

`public static string GetDigestSha(string digest) => digest?.Substring(digest.IndexOf("@") + 1);`

mthalman: Fixed

PR #612, opened 2020-08-24
Fix IsLocal tag issues in BuildCommand

mthalman:
Fixes two issues that were identified in how `BuildCommand` handles tags that are marked as `IsLocal`:

* Local tags were being pushed to the staging location but not actually being made public.  This was due to an issue in the `PushImages` method that was enumerating built tags but not filtering out those marked as local: https://github.com/dotnet/docker-tools/blob/8c270da5124f32ce1a407f977817e31c7bba113e/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs#L386

* Another issue was that an exception would be thrown when trying to find the digest of a tag that is marked as local: https://github.com/dotnet/docker-tools/blob/8c270da5124f32ce1a407f977817e31c7bba113e/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs#L120
  This occurred because when constructing the lookup data, it was actually filtering out tags marked as local.

These issues have been fixed by exposing the TagInfo objects rather than just the tag strings so that the appropriate metadata is available to determine whether the tag is local or not.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Models/Image/PlatformData.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Models/Image/PlatformData.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Models/Image/PlatformData.cs
```diff
@@ -32,10 +43,14 @@ public class PlatformData : IComparable<PlatformData>
...
         [JsonIgnore]
-        public IEnumerable<string> FullyQualifiedSimpleTags { get; set; }
+        public ImageInfo ImageInfo { get; set; }
+
+        [JsonIgnore]
+        public PlatformInfo PlatformInfo { get; set; }
 
         [JsonIgnore]
-        public IEnumerable<string> AllTags { get; set; }
+        public IEnumerable<TagInfo> AllTags =>
```
MichaelSimons, line 52: Would it make sense to encapsulate these JsonIgnore properties within a view model?  Would this help?

mthalman: At this point I don't think that's necessary.  It would just introduce yet another class that we'd have to name and deal with.  By encapsulating the setting of these properties via the constructor and image info load logic, I think this alleviates the issues so that the properties are always relevant.

PR #616, opened 2020-08-25
Update ImageBuilder tag

mthalman:


mthalman:
Closing this since more than a simple tag change is required to make use of the new caching logic.

PR #618, opened 2020-08-26
Enable build cache from pipeline

mthalman:
Makes use of the new build cache logic from https://github.com/dotnet/docker-tools/pull/610 from the pipeline templates.  

A few notes on how this is implemented:
An image info file from the dotnet/versions repo needs to be passed to the Image Builder to make use of build caching.  The way in which that file is retrieved is by configuring the build job to checkout multiple Git repos: the primary one for the build and the versions repo.  How this is done differs on whether it is a public or internal build.

Public builds only have access to the versions repo in GitHub while internal builds only have access to the internal AzDO versions repo.  Those are two different repo types and require different ways of accessing them as described in the [official docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/multi-repo-checkout?view=azure-devops).  The reference to the internal repo is encapsulated within the build job template.  But in order to reference the public repo, it requires a repositories reference in the pipeline YAML.  The build job template defines a well-known name for this reference, "VersionsRepo", but it's up to each public pipeline that wants to make use of caching to define this referenced repository in the following way:

```
resources:
  repositories:
  - repository: VersionsRepo
    type: github
    endpoint: dotnet
    name: dotnet/versions
```

Checking out multiple repositories on an agent produces a change to the folder structure.  With a single repo, the repo root is located at directory `s`.  But with multiple repos, they get nested within that directory.  For example: `s/dotnet-dotnet-docker` and `s/dotnet-versions`.  This requires updating some of the usage of paths throughout the pipeline YAML files to account for this difference.

Related to #186


eng/common/templates/steps/set-image-info-path-var.yml
```diff
@@ -0,0 +1,19 @@
...
+        $publicSourceBranch = "nightly"
+    }
+    else {
+        $publicSourceBranch = "master"
+    }
+
+    $imageInfoVersionsPath = "build-info/docker/image-info.$buildRepoName-$publicSourceBranch$env:IMAGEINFOVARIANT.json"
+
+    echo "##vso[task.setvariable variable=imageInfoVersionsPath]$imageInfoVersionsPath"
+  displayName: Set Image Info Path Var
```
MichaelSimons, line 19: Missing EOF

eng/common/templates/jobs/build-images.yml
```diff
@@ -23,23 +23,57 @@ jobs:
...
+      echo "##vso[task.setvariable variable=buildRepoName]$buildRepoName"
+      echo "##vso[task.setvariable variable=engCommonPath]$engCommonPath"
+      echo "##vso[task.setvariable variable=engPath]$engPath"
+      echo "##vso[task.setvariable variable=testScriptPath]$testScriptPath"
+    displayName: Set Common Paths
   - template: ${{ format('../steps/init-docker-{0}.yml', parameters.dockerClientOS) }}
   - ${{ parameters.customInitSteps }}
-  - script: echo "##vso[task.setvariable variable=imageBuilderImageInfoArg]--image-info-output-path $(artifactsPath)/$(legName)-image-info.json"
-    displayName: Set Image Info Arg for Image Builder
+  - powershell: |
```
MichaelSimons, line 55: Thoughts on consolidating this logic which sets the $imageBuilderBuildArgs with the conditionalized logic in the variables?  It would be nice to have all the logic in one place.

mthalman: Good suggestion.  Fixed

mthalman:
I've updated the template to allow pipelines to explicitly opt-out of caching via a parameter.  Caching can still be disabled at build time using a `noCache` build variable.

eng/pipelines/dotnet-docker-tools-eng-validation-pr.yml
```diff
@@ -12,3 +12,5 @@ variables:
 
 stages:
 - template: ../common/templates/stages/build-test-publish-repo.yml
+  parameters:
+    noCache: true
```
MichaelSimons, line 16: missing EOF

PR #620, opened 2020-08-27
Handle multiple digests for a given tag

mthalman:
The current logic which retrieves an image's digest is flawed because it doesn't handle cases where there are multiple digests associated with the tag being queried.  This happens when the local machine has pulled multiple tags of differing media types.  For example, if you pull `mcr.microsoft.com/windows/nanoserver:2004` (a manifest list) and `mcr.microsoft.com/windows/nanoserver:2004-amd64` (a regular manifest), you'll end up with something like the following result:

```
$ docker inspect mcr.microsoft.com/windows/nanoserver:2004 -f "{{ index .RepoDigests }}"
[mcr.microsoft.com/windows/nanoserver@sha256:b81252be487c6873e4c72c08bae06f8005b17db4e1047e0666c823acdc4bb181 mcr.microsoft.com/windows/nanoserver@sha256:d924594d5f49b352e6eaadd6227281080a7b616817967c7d211154812365f965]
```

The logic for retrieving the digest is broken with this scenario because it only retrieves the first digest in this array.  But the value of the first digest in the array depends on which tag was last pulled (and actually downloaded, not a cache hit).

These changes fix the logic to make use of the manifest tool to get the digest of the queried tag rather than the `RepoDigests` value from `docker inspect`.  This value is then validated against the list of pulled digests to ensure they're not out-of-date.

Somewhat related to this, I've updated `GetStaleImagesCommand` to use the manifest tool to get the image digest rather than pulling and executing `docker inspect`.  Pulling the image is wasted compute since it will be executing the manifest tool anyway.


src/Microsoft.DotNet.ImageBuilder/src/ManifestToolServiceExtensions.cs
```diff
@@ -2,24 +2,48 @@
...
         {
-            JArray tagManifest = manifestToolService.Inspect(tag, isDryRun);
-            return tagManifest?
-                .OfType<JObject>()
-                .First(manifestType =>
+            IEnumerable<JObject> tagManifests = manifestToolService.Inspect(tag, isDryRun).OfType<JObject>();
+            return GetDigestOfMediaType(tag, tagManifests, ManifestToolService.ManifestListMediaType, throwIfNull: true);
+        }
+
+        public static ManifestDigest GetAnyManifestDigest(this IManifestToolService manifestToolService, string tag, bool isDryRun)
```
MichaelSimons, line 20: This name is confusing to me especially in the context of a manifest list.  `Any` in the name made me think it could return a digest of one of the concrete tags.  I think removing `Any` from the name would make it more clear.  Another alternative that comes to mind is to have a single GetManifestDigest that takes an option enum on the media type - manifest, manifestlist, any(Default).

src/Microsoft.DotNet.ImageBuilder/src/ManifestDigest.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.ImageBuilder
```
MichaelSimons, line 5: Is this type necessary?  I didn't spot any places which referenced the MediaType.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -455,7 +455,25 @@ private void PullBaseImages()
...
+                {
+                    foreach (string fromImage in baseImages)
+                    {
+                        dockerService.PullImage(fromImage, Options.IsDryRun);
+
+                        // Ensure the digest of the pulled image is retrieved right away after pulling so it's available in
+                        // the DockerServiceCache for later use.  The longer we wait to get the digest after pulling, the
+                        // greater change the tag could be updated resulting in a different digest returned than what was
+                        // originally pulled.
+                        dockerService.GetImageDigest(fromImage, Options.IsDryRun);
```
MichaelSimons, line 470: I don't think we want to pull the images in parallel but we could certainly retrieve the image digests in parallel.  This seems potentially useful because of the network calls.  Additionally I think this may be retrieving more digests than necessary.  Don't we only need the digests of the final stage FROM images?

mthalman: Fixed

src/Microsoft.DotNet.ImageBuilder/src/ManifestDigest.cs
mthalman: This was leftover from an earlier implementation.  Removed it.

src/Microsoft.DotNet.ImageBuilder/src/ManifestToolServiceExtensions.cs
mthalman: Refactored to have a single method with enum.

src/Microsoft.DotNet.ImageBuilder/src/ManifestToolServiceExtensions.cs
```diff
@@ -11,25 +11,33 @@ namespace Microsoft.DotNet.ImageBuilder
...
+            }
+            
+            if (mediaType.HasFlag(ManifestMediaType.Manifest))
+            {
+                return GetDigestOfMediaType(
+                    tag, tagManifests, ManifestToolService.ManifestMediaType, throwIfNull: true);
             }
 
-            return digest;
+            throw new ArgumentException($"Unsupported media type: '{mediaType}'.", nameof(mediaType));
```
MichaelSimons, line 37: If I call `GetManifestDigestSha(ManifestMediaType.ManifestList, "non-list tag", false)`, it looks like this will cause an ArgumentException to be thrown.  I was expecting `null` to be returned in this case.

mthalman: Fixed the logic to not throw an unsupported media type in that scenario.  But it will still throw an exception in this case.  If there is no match for the requested media type, it throws.  I believe this should be the behavior we want because it's a programming error IMO if that case should occur.

PR #623, opened 2020-08-28
Add GenerateImageGraph command

MichaelSimons:
This command generates a DOT (graph description language) file illustrating the image and layer hierarchy.  This command is not intended for any production scenarios it is intended for local desktop usages to generate visualizations of the images within a manifest.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateImageGraphCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateImageGraphCommand.cs
@@ -1,0 +1,156 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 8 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateImageGraphCommand.cs
```diff
@@ -0,0 +1,156 @@
...
+                myEdge.Attributes.Style = DotStyle.Dashed;
+                graph.Edges.Add(myEdge);
+            }
+        }
+
+        private decimal ConvertToMB(long bytes) => bytes / 1048576m;
+
+        private static string GetImageDisplayName(string tag, bool trimParentRepos)
+        {
+            string displayName = tag.TrimStart("mcr.microsoft.com/");
```
mthalman, line 127: Get the registry from Manifest.Registry instead?

MichaelSimons: Not sure why I hard coded this...Fixed

PR #627, opened 2020-08-31
Fix digest parsing

mthalman:
Fixes an issue with parsing the digests returned from the `docker inspect` command.  A locally built image will have an empty array returned for its `RepoDigests` field.  When trimming the array characters, it's left with an empty string.  Calling `Split` on an empty string produces an array with a single empty string rather than an empty array.  Fixed the code to return an empty enumerable in that case.

The namespaces were reordered according to Visual Studio's sorting.  Without fixing the sorting, it couldn't automatically place new namespaces in the correct order.


mthalman:
The last set of changes I made was to fix an issue that wasn't caught by the PR build because the issue was a result of two separate PRs that were merged in but the combination of the two was never tested.

Issue #613, opened 2020-08-25
Mitigate Docker manifest request rate limiting


mthalman:
Docker has [announced](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/) new rate limits for manifest requests.  The rate limits take effect Nov. 1, 2020.

> Free plan – anonymous users: 100 pulls per 6 hours 
> Free plan – authenticated users: 200 pulls per 6 hours
> Pro plan – unlimited
> Team plan – unlimited

Currently, our infrastructure makes these requests anonymously so it would be limited to 100 requests per 6 hrs.  However, even the 200 request limit is likely to problematic for some of our scenarios.

There needs to be a plan for mitigating this rate limit.

mthalman:
Current POR:

* We should get a paid account to be used by our bot.  This will eliminate any possibility of encountering the rate limit.
* We should mirror the images we depend on from Docker Hub into our ACR.  We shouldn't be reliant on the availability and operation of a third party service in order to publish our images.
* Mirroring should be done periodically by the base image monitoring system.
* Any time an official build is started, it should first trigger a mirroring update.  If mirroring fails, the build can still continue.

mthalman:
Tasks:

- [ ] ImageBuilder command to mirror images
- [ ] Scheduled pipeline to run command to mirror images
- [ ] Dynamically rewrite Dockerfiles at build time to target mirror location
- [ ] Use paid account for connecting to Docker Hub

MichaelSimons:
@dleeapho, What is the status of getting a Docker Hub account for us to use?

mthalman:
A Docker Hub account is now available so I'll start work on updating our pipelines to use that.

PR #629, opened 2020-09-01
Command to mirror DH images

mthalman:
Adds a command that will find all the external base image tags that are being depended upon and copy those images to our ACR in a mirrored repo location.  This refactors the previous `CopyImagesCommand` class since it has common implementation that can be shared by the two commands.

Related to #613


src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesOptions.cs
```diff
@@ -6,15 +6,14 @@
 
 namespace Microsoft.DotNet.ImageBuilder.Commands
 {
-    public class CopyAcrImagesOptions : ManifestOptions, IFilterableOptions
+    public class CopyAcrImagesOptions : CopyImagesOptions, IFilterableOptions
```
MichaelSimons, line 9: Why not make `CopyImageOptions` implement `IFilterableOptions`?  It seems like it would reduce the amount of duplicated code.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyDockerHubImagesOptions.cs
```diff
@@ -0,0 +1,22 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.CommandLine;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class CopyDockerHubImagesOptions : CopyImagesOptions, IFilterableOptions
```
MichaelSimons, line 9: The name feels a little vague.  Thoughts on CopyBaseImages, MirrorBaseImages, or MirrorDockerHubBaseImages?

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyDockerHubImagesCommand.cs
```diff
@@ -0,0 +1,33 @@
...
+            : base(azureManagementFactory, loggerService)
+        {
+        }
+
+        public override async Task ExecuteAsync()
+        {
+            this.LoggerService.WriteHeading("COPYING IMAGES");
+
+            IEnumerable<Task> importTasks = Manifest.GetExternalFromImages()
+                .Select(fromImage => ImportImageAsync($"{ Options.RepoPrefix}{fromImage}", fromImage, srcRegistryName: "docker.io"));
```
MichaelSimons, line 28:
```suggestion
                .Select(fromImage => ImportImageAsync($"{Options.RepoPrefix}{fromImage}", fromImage, srcRegistryName: "docker.io"));
```

MichaelSimons, line 28: > srcRegistryName: "docker.io"

Should this logic assume all base images are coming from DH?  In buildtools-prereqs, I could easily see images being created that are based on MCR images.  With how this command get's used, will it run on the samples?  It might be easier to let it run and no-op versus having to special case the samples.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesOptions.cs
mthalman: Fixed

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyDockerHubImagesOptions.cs
mthalman: How about CopyDockerHubBaseImages?

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyDockerHubImagesCommand.cs
mthalman: Added a check to filter out any that are from MCR.

MichaelSimons (approved):

PR #641, opened 2020-09-11
Add generateReadmes command to generate the readmes from templates

MichaelSimons:
Related to #https://github.com/dotnet/dotnet-docker/issues/1931

* Created a generateReadmes command that will generated the readmes from templates.
* Refactored the generateDockerfiles command to extract the generation logic into a base class that is shared with the new generateReadmes command.
* Removed the generateTagsReadme command and made it part of the new generateReadmes command.  As part of this, I simplified the usage as the old generateTagsReadme command only supported one repo at a time.  The results is generating the readmes for repos like dotnet-docker is much faster.
* Renamed the `ReadmePath` and `McrTagsMetadataTemplatePath` properties in the model to not have the `Path` suffix.  The model was inconsistent as some properties used the suffix while others didn't - e.g. `Dockerfile` and `DockerfileTemplate`.


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateArtifactsCommand.cs
```diff
@@ -0,0 +1,173 @@
...
+                }
+                else if (!Options.IsDryRun)
+                {
+                    await File.WriteAllTextAsync(artifactPath, generatedArtifact);
+                    Logger.WriteMessage($"Updated '{artifactPath}'");
+                }
+            }
+        }
+
+        private static string GetSnippet(string source, int index) => source.Substring(index, Math.Min(100, source.Length - index));
```
mthalman, line 110: Perhaps include leading text prior to the diff in addition to the trailing text?  That will help provide more context at a glance.
```suggestion
        private static string GetSnippet(string source, int index) => source.Substring(Math.Max(0, index - 100), Math.Min(100, source.Length - index));
```

src/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
```diff
@@ -13,11 +13,12 @@ public static class ReadmeHelper
 
         public static string UpdateTagsListing(string readme, string tagsListing)
         {
-            tagsListing = $"{TagsSectionHeader}{Environment.NewLine}{Environment.NewLine}{tagsListing}{Environment.NewLine}{Environment.NewLine}";
-
             // Normalize the line endings to match the readme.
             tagsListing = tagsListing.NormalizeLineEndings(readme);
 
+            string targetLineEnding = readme.Contains("\r\n") ? "\r\n" : "\n";
```
mthalman, line 19: Can you put this into a helper method and refactor `StringExtensions` to also use that method?

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateArtifactsCommand.cs
MichaelSimons: This can get tricky because to be useful, I believe you should provide an indication where the difference begins.  With my experience so far, including the diff snippet is "good enough" to quickly know what is happening.  I did however add the diff index to the message to provide some context.  

```
                    Logger.WriteError($"{artifactName} out of sync with template starting at index '{differIndex}'{Environment.NewLine}"
                        + $"Current:   '{GetSnippet(currentArtifact, differIndex)}'{Environment.NewLine}"
                        + $"Generated: '{GetSnippet(generatedArtifact, differIndex)}'");
```

If this proves to be problematic, I suggest we refactor at that time.

src/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
MichaelSimons: Fixed.

mthalman (approved):

Issue #632, opened 2020-09-02
Test trimming based on build cache results


mthalman:
This is part of a set of phases from the work outlined in https://github.com/dotnet/docker-tools/issues/186.

This phase of work is just about resolving the issue of testing images that do not need to be tested.  Phase 1 allowed for previously published images to be pulled and reused but the testing stage would continue to test all images.  This is inefficient because images had already been tested when they were originally published.

One of the goals in resolving the issue is to determine what needs to be tested when the test matrix is generated in order to avoid spinning up test jobs that may just result in no images being tested.  In order to do that, the matrix generation logic will require the image info that was output by the build.  That determines which images were built and, when compared against the dependency graph of the Dockerfiles, it can be determined which images are actually cached versions.

By providing to the `generateBuildMatrix` command a set of values that map Dockerfile path prefixes to test categories, the logic can output the set of test categories that should be passed to the test script.

For example, consider a call to the `generateBuildMatrix` command with the following subset of parameters:

```
--dockerfile-test-category 'runtime-deps=src/runtime-deps/*' --dockerfile-test-category 'runtime=src/runtime/*' --dockerfile-test-category 'aspnet=src/aspnet/*' --dockerfile-test-category 'sdk=src/sdk/*'
```
 And let's say that the only new image produced by the build was sdk.  The test matrix generation uses the image info to determine that only the sdk image was built by this build.  The matrix output that it generates the specifies a test-categories field that indicates which categories should be passed to and executed by the test script, in this case "sdk".

Example representation of a leg of the test matrix:

```
    5.0-focal:
      legName: linuxarm64v85.0-focal
      osType: linux
      architecture: arm64
      osVersions: --os-version focal
      dotnetVersion: 5.0
      osVariant: focal
      test-categories: sdk
```

If a test leg results in there being no test categories due to all of the images being the result of cache hits, that test leg is excluded entirely from the matrix.  This prevents the execution of test jobs that end up not needing to test anything.

At this point, the test logic executes as it always has since it already supports the ability to test subsets of images based on test categories.

PR #645, opened 2020-09-15
Support test job trimming from cached images

mthalman:
Updates the `generateBuildMatrix` command to output a `testCategories` field that defines which test categories should be run in a test job.  Which test categories to include is determined by whether the image was built or retrieved from the cache.  Images from the cache do not need to be tested.  If a test leg doesn't have any test categories, it is excluded from the matrix in order to prevent any test job from running.

Related to #632


mthalman:
Closing this PR in favor of a simpler solution.  This approach made use of a testCategories field that would get included in the legs of the generated matrix that would allow running tests on a partial set of image categories.  Even in the case where only an SDK is being built, the minimum number of tests to be run would include runtime, ASP.NET, and SDK scenarios.  Based on that, there's not much benefit to filtering which test categories are to be run.  We might as well run everything because there's not much additional overhead and it allows for a simpler implementation of the `generateBuildMatrix` command.

PR #652, opened 2020-09-17
Fix matrix generation to support multiple platforms referencing a common Dockerfile

mthalman:
When there are more than one platforms defined in the manifest that are in separate repos but reference the same Dockerfile, it leads to multiple build legs being generated that have the same name.  An example of this is with the consolidation of runtime-deps between 3.1 and 5.0 (dotnet/dotnet-docker#2122) where matrix legs like the following can end up being generated:

```
  linuxAmd64:
    src-runtime-deps-3.1-alpine3.12-graph:
      imageBuilderPaths: --path src/runtime-deps/3.1/alpine3.12/amd64 --path src/runtime/3.1/alpine3.12/amd64 --path src/aspnet/3.1/alpine3.12/amd64 --path src/sdk/3.1/alpine3.12/amd64 --path src/monitor/5.0/alpine/amd64
      legName: linuxamd64src-runtime-deps-3.1-alpine3.12-graph
      osType: linux
      architecture: amd64
      osVersions: --os-version alpine3.12
    src-runtime-deps-3.1-alpine3.12-graph:
      imageBuilderPaths: --path src/runtime-deps/3.1/alpine3.12/amd64 --path src/runtime/5.0/alpine3.12/amd64 --path src/aspnet/5.0/alpine3.12/amd64 --path src/sdk/5.0/alpine3.12/amd64
      legName: linuxamd64src-runtime-deps-3.1-alpine3.12-graph
      osType: linux
      architecture: amd64
      osVersions: --os-version alpine3.12
```

These changes fix this issue by combining legs that share a common Dockerfile into a single leg.


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -11,6 +11,7 @@
 using Microsoft.DotNet.ImageBuilder.Models.Image;
 using Microsoft.DotNet.ImageBuilder.Models.Manifest;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
+using Microsoft.VisualStudio.Services.Common;
```
MichaelSimons, line 14: I'm not spotting why is this needed.  Why type(s) are being referenced?

MichaelSimons (approved):

PR #654, opened 2020-09-18
Support for Windows Server Core in Dockerfile generation

mthalman:


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateDockerfilesCommand.cs
```diff
@@ -47,12 +47,28 @@ public override async Task ExecuteAsync()
...
             symbols["OS_ARCH_HYPHENATED"] = GetOsArchHyphenatedName(platform);
 
             return symbols;
         }
 
+        private static string GetOsVersionNumber(PlatformInfo platform)
+        {
+            const string PrefixGroup = "Prefix";
+            const string VersionGroup = "Version";
+            Match match = Regex.Match(platform.Model.OsVersion, @$"(-(?<{PrefixGroup}>\w*))?(?<{VersionGroup}>\d+.\d+)");
```
MichaelSimons, line 60: Are the two groups necessary here?  Can't they be combined?  `(-(?<{VersionGroup}>\w*\d+.\d+)`

mthalman: I couldn't think of a way to do it with one.  Your suggestion doesn't match on `alpine3.12` and if you change your suggested regex to make the hyphen optional (`-?`) then it'll match on the whole string of `alpine3.12` which is not what we want either.

MichaelSimons: I see now, I got mixed up on the parens.

MichaelSimons (approved):

PR #655, opened 2020-09-22
Build support for shared Dockerfiles

mthalman:
Related to https://github.com/dotnet/dotnet-docker/issues/2122

Fixes the issue described in this comment: https://github.com/dotnet/dotnet-docker/pull/2260#discussion_r492816689.

This is fixed by tracking info about the images that have been retrieved from the cache.  Regardless of whether an image has an entry in the image info file, it will get a cache hit if it shares the same Dockerfile path as an image that has been retrieved from the cache and tagged appropriately.

Note that this does not handle the scenario where there's a shared Dockerfile and neither of the platforms result in cache hits.  In that scenario, the first platform will end up having its image built as part of the normal cache miss processing.  The second platform will also get a cache miss even though technically its Dockerfile has already been built.  However, running a build won't produce a new image because the layers are already cached locally on the build machine.  This is a compromise so that we can rely on Docker for the caching behavior and keep our code simpler.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -92,19 +92,16 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
             foreach (PlatformData platform in GetBuiltPlatforms())
             {
-                PlatformInfo manifestPlatform = Manifest.GetFilteredPlatforms()
-                   .First(manifestPlatform => platform.Equals(manifestPlatform));
-
                 foreach (TagInfo tag in GetPushTags(platform.PlatformInfo.Tags))
                 {
                     if (Options.IsPushEnabled)
                     {
-                        SetPlatformDataDigest(platform, manifestPlatform, tag.FullyQualifiedName);
+                        SetPlatformDataDigest(platform, tag.FullyQualifiedName);
                         SetPlatformDataBaseDigest(platform, platformDataByTag);
                     }
 
                     SetPlatformDataCreatedDate(platform, tag.FullyQualifiedName);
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -231,19 +216,57 @@ private void BuildImages()
...
-                    imageArtifactDetails?.Repos.Add(repoData);
+                    _imageArtifactDetails?.Repos.Add(repoData);
                 }
             }
         }
 
+        private bool CheckForCachedImage(
+            ImageData? srcImageData, PlatformInfo platform, IEnumerable<string> allTags, PlatformData? platformData)
+        {
+            if (platformData != null && _cachedDockerfilePaths.TryGetValue(platformData.Dockerfile, out BuildCacheInfo? cacheInfo))
```
MichaelSimons, line 235: I think it is possible this cache may break one scenario.  The scenario involves having a parameterized Dockerfile which is built several times with different inputs.   This is not something we do today but is being considered in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/231.  To support this, the cache key would need to include the args the Dockerfile was built with.

mthalman: In theory, you could also run into this same situation even when Dockerfile sharing isn't happening by dynamically specifying manifest variables at build time that are configured as build args in the manifest.  That's not something we do today either and is even more of an edge case.  In that case, an option would be store the build args in the image info.  Regardless, they have separate solutions so I'll make a fix for the scenario you identified now.

mthalman: Fixed

MichaelSimons (approved):

PR #659, opened 2020-09-25
Fix issue with digest renaming in caching logic

mthalman:
There was an issue with the changes from #655 in the logic which attempts to rewrite the image name of a digest to be based on a different repo.  This logic targeted the repo from the tag which is actually the overriden (staging) registry and repo location, rather than the MCR location.  This causes the incorrect image name to be stored in the image info file which messes things up.

I've fixed this by using the raw registry and repo names of the target repo in order to build up the correct image name.


mthalman:
I'm merging this now to unblock the issue with builds in nightly.  Feel free to continue to reviewing this.

PR #661, opened 2020-09-28
Ensure new tags are published despite cache results

mthalman:
There's an issue with the caching logic that prevents publishing of new tags that are introduced if the image results in a cache hit during the build.  Just because an image was retrieved from the cache doesn't mean it still has a tag that needs to be published; the two are orthogonal.

The caching logic has been fixed to check whether the previously published tags differ from the set of tags that were built.  If they differ the image needs to be published.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -236,7 +236,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             {
                 OnCacheHit(repo, allTags, pullImage: false, cacheInfo.Digest);
                 platformData.BaseImageDigest = cacheInfo.BaseImageDigest;
-                platformData.ExcludeFromPublish = srcPlatformData != null &&
+                platformData.IsUnchanged = srcPlatformData != null &&
                     CachedPlatformHasAllTagsPublished(srcPlatformData);
                 return true;
             }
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -229,42 +229,48 @@ private void BuildImages()
...
-                // If this Dockerfile has been built and published before
-                if (srcPlatformData != null)
-                {
-                    isCachedImage = CheckForCachedImageFromImageInfo(repo, platform, srcPlatformData, allTags);
+                isCachedImage = CheckForCachedImageFromImageInfo(repo, platform, srcPlatformData, allTags);
 
-                    if (platformData != null)
+                if (platformData != null)
+                {
+                    platformData.ExcludeFromPublish = isCachedImage &&
```
MichaelSimons, line 253: I didn't have a favorable initial reaction to this name.  I think this is because I don't like the idea of the BuildCommand knowing about publish.  Perhaps renaming this to something is `HasChanges` or `HasChanged` would be sufficient to remove this apparent coupling.

mthalman: I went with `IsUnchanged` because it's important for the default value to be false.  Everything that gets published and, by extension, everything that's in the image info file is inherently "changed" so all content in the image info file has `IsUnchanged` set to false.

src/Microsoft.DotNet.ImageBuilder/src/Commands/TrimCachedPlatformsCommand.cs
```diff
@@ -49,7 +49,7 @@ private void RemoveCachedPlatforms(ImageArtifactDetails imageArtifactDetails)
                     for (int i = image.Platforms.Count - 1; i >= 0; i--)
                     {
                         PlatformData platform = image.Platforms[i];
-                        if (platform.IsCached)
+                        if (platform.ExcludeFromPublish)
```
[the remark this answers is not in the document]
mthalman: Updated to `TrimUnchangedPlatformsCommand`.

MichaelSimons (approved):

PR #662, opened 2020-09-29
Dockerfile generation support for amd64 tag suffix

mthalman:
For the changes being made for #2182, there needs to be a way to generate Dockerfiles from a template that makes use of the `ARCH_TAG_SUFFIX` variable which is always populated.  Currently, it defaults to an empty value if the architecture is AMD64.

This is fixed by including an option to set the default architecture that, if set, will be excluded as a tag suffix.  For 2.1/3.1 Dockerfile templates, this will be set to "AMD64".  For 5.0, this option will not be set.

Related to https://github.com/dotnet/dotnet-docker/issues/2182

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateDockerfilesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateDockerfilesCommand.cs
@@ -44,18 +44,11 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             symbols["ARCH_SHORT"] = platform.Model.Architecture.GetShortName();
             symbols["ARCH_NUPKG"] = platform.Model.Architecture.GetNupkgName();
             symbols["ARCH_VERSIONED"] = versionedArch;
+            symbols["ARCH_TAG_SUFFIX"] = $"-{versionedArch}";
             symbols["OS_VERSION"] = platform.Model.OsVersion;
             symbols["OS_VERSION_BASE"] = platform.BaseOsVersion;
             symbols["OS_VERSION_NUMBER"] = GetOsVersionNumber(platform);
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateDockerfilesCommand.cs
```diff
@@ -44,12 +44,19 @@ public override async Task ExecuteAsync()
             symbols["ARCH_SHORT"] = platform.Model.Architecture.GetShortName();
             symbols["ARCH_NUPKG"] = platform.Model.Architecture.GetNupkgName();
             symbols["ARCH_VERSIONED"] = versionedArch;
-            symbols["ARCH_TAG_SUFFIX"] = platform.Model.Architecture != Architecture.AMD64 ? $"-{versionedArch}" : string.Empty;
             symbols["OS_VERSION"] = platform.Model.OsVersion;
             symbols["OS_VERSION_BASE"] = platform.BaseOsVersion;
             symbols["OS_VERSION_NUMBER"] = GetOsVersionNumber(platform);
             symbols["OS_ARCH_HYPHENATED"] = GetOsArchHyphenatedName(platform);
 
+            string archTagSuffix = $"-{versionedArch}";
```
MichaelSimons, line 52: What do you think about always setting the `ARCH_TAG_SUFFIX` and then having the templates decide whether or not it should be used via conditionals.  The benefits are that it keeps the infrastructure simpler, makes the templates more self-contained, and `ARCH_TAG_SUFFIX` will always be used going forward making this new option the norm once 3.1 and 2.1 reach EOL.

mthalman: Good idea.  Fixed

MichaelSimons (approved):

Issue #664, opened 2020-09-30
Redesign manifest schema to more easily reuse platforms


mthalman:
The changes outlined in dotnet/dotnet-docker#2182 require having multiple distinct sets of manifest tags associated with a given platform.  For example, the Dockerfile at https://github.com/dotnet/dotnet-docker/blob/aa8e95b22897ed44f6ba359e2a1d74ed9caca03f/src/runtime/5.0/buster-slim/amd64/Dockerfile would have two separate sets of manifest tags associated with it:
* 5.0.0-rc.2, 5.0, latest
* 5.0.0-rc.2-buster-slim, 5.0-buster-slim

The schema currently doesn't support being able to define more than 1 set of manifest tags for a given platform without having to duplicate the platform definition.  This is because manifest tags are defined by an image and images contain platforms.

In order to avoid duplication of platforms across the manifest, it'd be better if the manifest schema could be redesigned to better support this scenario.

mthalman:
To compensate for not having this implemented, a strategy was taken to duplicate platforms within the manifest so that they could have multiple sets of shared tags associated with them.  Some changes to Image Builder were made to handle this (see below).  As part of the implementation of the schema change proposed in this issue, we should determine whether these other changes can be removed.

https://github.com/dotnet/docker-tools/pull/665
https://github.com/dotnet/docker-tools/pull/666

mthalman:
Another scenario for platform reuse -- and image reuse as well -- is the need to be able to tag a given image with tags that consist of multiple repos (e.g. syndication: https://github.com/dotnet/dotnet-docker/issues/2317).  

This would then argue for a design that decouples images from repos and platforms from images so that they can be mixed and matched to produce the desired state.  If we're going to do the work to redesign the manifest, we might as well go all the way to be able to handle these scenarios.

For completeness, here's a set of scenarios that it would make sense to support with a schema change that doesn't require duplication of elements:
* Associate multiple distinct sets of manifest tags with a given platform.
  * Example: an image could be assigned the following sets of manifest tags: `5.0.0-rc.2, 5.0, latest` and `5.0.0-rc.2-buster-slim, 5.0-buster-slim`.
* Tag an image across multiple repos.
  * Example: an image could be tagged with `mcr.microsoft.com/dotnet/core/runtime:3.1-buster-slim` and `mcr.microsoft.com/dotnet/runtime:3.1-buster-slim`.
* Publish an image that is not tagged.
* Associate a manifest tag with an untagged image.

[...]

PR #666, opened 2020-10-01
Manifest publishing support for duplicated platforms

mthalman:
With the changes being done for dotnet/dotnet-docker#2182, it will cause a platform to be defined multiple times in a manifest, contained in separate images. This requires changes to `PublishManfiestCommand` in order to correctly handle this scenario because it will otherwise throw an exception when it encounters a platform with no concrete platform tags.

This is solved by finding the matching platform that does contain a concrete tag and using that as the reference tag for the manifest creation.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
@@ -37,14 +37,17 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
             ExecuteWithUser(() =>
             {
-                IEnumerable<ImageInfo> multiArchImages = Manifest.GetFilteredImages()
-                    .Where(image => image.SharedTags.Any())
+                IEnumerable<(RepoInfo Repo, ImageInfo Image)> multiArchRepoImages = Manifest.FilteredRepos
+                    .SelectMany(repo =>
+                        repo.FilteredImages
+                            .Where(image => image.SharedTags.Any())
+                            .Select(image => (repo, image)))
                     .ToList();
 
                 DateTime createdDate = DateTime.Now.ToUniversalTime();
-                Parallel.ForEach(multiArchImages, image =>
+                Parallel.ForEach(multiArchRepoImages, repoImage =>
                 {
-                    string manifest = GenerateManifest(image);
+                    string manifest = GenerateManifest(repoImage.Repo, repoImage.Image);
 
                     string manifestFilename = $"manifest.{Guid.NewGuid()}.yml";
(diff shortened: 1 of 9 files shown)


mthalman:
Moving this to draft because I found some issues that need fixing.

mthalman:
I've also fixed an issue I found with the merging of image info content.  Duplicated platforms contained in separate images were being detected as equal in `PlatformData.Equals`.  This meant there content would be merged together instead of kept in separate images.

The solution is not great.  It now takes into account the convention a duplicated platform will not have simple tags defined for it.  When one platform has simple tags and another does not, they are now considered to not be equal regardless of whether they having a matching Dockerfile, platform, etc.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
```diff
@@ -107,7 +110,32 @@ private string GenerateManifest(ImageInfo image)
...
             {
-                manifestYml.AppendLine($"- image: {platform.Tags.First().FullyQualifiedName}");
+                string imageTag;
+                if (platform.Tags.Any())
+                {
+                    imageTag = platform.Tags.First().FullyQualifiedName;
+                }
+                else
+                {
+                    (ImageInfo Image, PlatformInfo Platform) matchingImagePlatform = repo.AllImages
```
MichaelSimons, line 120: It would be good to track these specializations required for working around the schema limitations.  A good place to track them is dotnet/docker-tools#664.  The one added yesterday around the MCR tags yaml should be tracked as well.

It would be nice to clean these up when the schema supports this scenario without duplicating platforms.

MichaelSimons (approved):
Seeing this issue triggered me to question if the IngestKustoImageInfoCommand is going to handle this situation.  I think it should but can you take a look if you haven't already to ensure the correct info is getting ingested?

mthalman:
> Seeing this issue triggered me to question if the IngestKustoImageInfoCommand is going to handle this situation. I think it should but can you take a look if you haven't already to ensure the correct info is getting ingested?

It would result in duplicate digest lines outputted.  Is that ok?

MichaelSimons:
> > Seeing this issue triggered me to question if the IngestKustoImageInfoCommand is going to handle this situation. I think it should but can you take a look if you haven't already to ensure the correct info is getting ingested?
> 
> It would result in duplicate digest lines outputted.  Is that ok?

Yes, duplicated digests will be alright.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
mthalman: I've tracked the changes in this comment: https://github.com/dotnet/docker-tools/issues/664#issuecomment-702783549

Issue #650, opened 2020-09-16
CommitUrl diffs when image is built from feature branch


mthalman:
The image caching feature relies on the CommitUrl field defined in the image info file in order to determine whether to retrieve a cached image.  There are problems with this value depending on how the images were built.  This can lead to diffs in the value resulting in incorrect cache misses.

An example of this was the recent 9B release for the .NET Fx images (https://github.com/microsoft/dotnet-framework-docker/issues/634).  In that release, the images were optimistically built from a feature branch while the PR was under review and validating.  Prior to publishing those images, the PR was merged.  The build to publish the images was then run from the master branch.  The problem with this workflow is that the image info gets populated with the commit SHAs from the feature branch.  Let's call this SHA X.  The image info that gets published contains SHA X.  When the PR gets merged to master, a new SHA gets calculated: SHA Y.  When a subsequent build runs that theoretically should take advantage of caching, it will get a cache miss because the SHA that is retrieved from the build is SHA Y which differs from the SHA X in the image info file.

One solution to this problem is to defer the populating of the CommitUrl field in the image info until the publish stage instead of in the build stage.  This works as long as two conditions are met:
1. We only publish from an official branch, not a feature branch.  This will be enforced by https://github.com/dotnet/docker-tools/issues/605.
2. There are no subsequent commits that change the Dockerfiles to be built between the time when the build stage runs and the publish stage runs.  In that case, the commit SHA is not representative of the version of the Dockerfile that was built.  This would be a bad practice in general but needs to be called out.

The alternative is to change our workflow to never build images in a feature branch for an official release.  The advantage of using a feature branch is that it expedites the delivery of images by optimistically getting the images pre-built while further validation of the changes is ongoing.

MichaelSimons:
Decision is to make PR, allow one PR validation leg to complete as a smoke test, merge the PR, and then trigger the official build off of master.  

With this plan, no changes are needed - closing.

PR #668, opened 2020-10-05
Disallow official builds from non-official branches

### LINKED ISSUE #605: Disallow builds from non-official branches by default
In order to help prevent the unintentional publishing of images that are not meant to be public, we should consider there being a validation check in the publish stage that will fail the build if it's occurring on a non-official branch and the publish destination is "public".

I think this should only be a check within the publish stage because it is still useful to sometimes prep images from a non-official branch.  But once time comes to publish, it should be from an official branch.

mthalman:
Adds a validation step that would get executed in all stages of the build.  This validates that the build is running on an official branch if it's an official build.  Official branches are defined in a common variable that each pipeline can override if necessary.

I also added a variable that would allow this validation to be overriden.  This variable is intended as an escape hatch for now and would need to be explicitly added to a build run.

Fixes #605

### DIFF
--- a/eng/common/templates/jobs/generate-matrix.yml
+++ b/eng/common/templates/jobs/generate-matrix.yml
@@ -8,6 +8,7 @@ jobs:
   pool:
     vmImage: $(defaultLinuxAmd64PoolImage)
   steps:
+  - template: ../steps/validate-branch.yml
   - template: ../steps/init-docker-linux.yml
   - script: >
       $(runImageBuilderCmd) generateBuildMatrix
--- a/eng/common/templates/variables/common.yml
+++ b/eng/common/templates/variables/common.yml
@@ -23,6 +23,9 @@ variables:
   value: "00:20:00"
 - name: mcrDocIngestionTimeout
   value: "00:05:00"
+- name: officialBranches
+  # list multiple branches as "'branch1', 'branch2', etc."
+  value: "'master'"
 
 - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
   - group: DotNet-Docker-Common
--- a/eng/common/templates/jobs/publish.yml
+++ b/eng/common/templates/jobs/publish.yml
@@ -10,6 +10,7 @@ jobs:
       $(manifestVariables)
       $(imageBuilder.queueArgs)
   steps:
+  - template: ../steps/validate-branch.yml
   - template: ../steps/init-docker-linux.yml
   - template: ../steps/download-build-artifact.yml
     parameters:
--- a/eng/common/templates/steps/validate-branch.yml
+++ b/eng/common/templates/steps/validate-branch.yml
@@ -1,0 +1,10 @@
+steps:
+- ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+  - script: |
+      if [[ "$OFFICIALBRANCHES" != *\'$BUILD_SOURCEBRANCHNAME\'* && \
+            "$PUBLISHREPOPREFIX" == "public/" && \
+            "$OVERRIDEOFFICIALBRANCHVALIDATION" != "true" ]]; then
+          echo "##vso[task.logissue type=error]Official builds must be done from an official branch: $OFFICIALBRANCHES"
+          exit 1
+      fi
+    displayName: Validate Branch


eng/common/templates/variables/common.yml
```diff
@@ -23,6 +23,9 @@ variables:
   value: "00:20:00"
 - name: mcrDocIngestionTimeout
   value: "00:05:00"
+- name: officialBranches
```
MichaelSimons, line 26: How do you envision specifying this in nightly?

mthalman: In the nightly pipeline YML , it would override this variable with `"'nightly'"`

MichaelSimons: Would it make sense to include that change now?

eng/common/templates/jobs/generate-matrix.yml
```diff
@@ -8,6 +8,7 @@ jobs:
   pool:
     vmImage: $(defaultLinuxAmd64PoolImage)
   steps:
+  - template: ../steps/validate-branch.yml
```
MichaelSimons, line 11: I have mixed feeling about putting this check here.  If your intent is to run just the build and test legs, there is nothing wrong and you shouldn't have to change the PUBLISHREPOPREFIX.  I kind of like the idea of only enforcing this in the publish leg.

mthalman: Why would we run only the build and test legs for an official build without actually publishing?  If the intent is to eventually publish, we need to run the build in the official branch as well (this enforces that).  This was part of the discussion from https://github.com/dotnet/docker-tools/issues/650.

eng/common/templates/variables/common.yml
mthalman: You mean just create a PR in dotnet-docker that sets this value?  I was intending to do that as part of the PR that integrate these changes.

MichaelSimons: I mean include the nightly pipeline change with this PR.  It is a pretty simple change.  Including additional changes as part of the integration means there changes that need to be integrated immediately back to master.

mthalman: How can I include it in this PR when the nightly pipeline YAML file isn't in this repo?  I'm confused.

eng/common/templates/jobs/generate-matrix.yml
MichaelSimons: I used to do this frequently when testing MSRC releases.  Make the changes that pulled from the MSRC locations and run the  build and test legs.

eng/common/templates/variables/common.yml
MichaelSimons: I'm sorry, I somehow thought you were making this PR against master in dotnet-docker.  Please ignore my comment 😢

eng/common/templates/jobs/generate-matrix.yml
mthalman: But that's a test build so I would say you should be setting the publishRepoPrefix to "test/".  I really don't want to go through an entire build/test process only to have it fail in publish because it didn't start from an official branch.

mthalman: Well, it's even worse than that because you may have done the publish from an official branch but not the build.  So then you just end up in the problem of #650.  We need to prevent that from happening, explicitly.

MichaelSimons: That part I didn't like was having to specify publish related properties when running a test build that wasn't going to publish.  

There are numerous ways that you can "break" the build one of which you just mention.  I think the goal is to have basic validation, it is not worthwhile for us to catch and prevent every possible scenario.  For advanced scenarios, the responsibility lies in the person queuing the build.

MichaelSimons (approved):

eng/common/templates/jobs/generate-matrix.yml
mthalman: For scenarios like this, it would make sense to the use the playground pipeline which is already configured with "test/" as the publishRepoPrefix.

PR #672, opened 2020-10-08
Set image info data for duplicated platforms

mthalman:
The image info output of `BuildCommand` is incorrect for platforms defined in the manifest that have been duplicated in order to declare them in separate images and do not define concrete tags.  Those duplicated platforms do not have their Digest and Created fields set.

This happens because the code which sets those values does so by querying Docker based on the platform's concrete tags.  Since the platform has no concrete tags, those platforms do not get the digest and created date info set.

To fix this, there's a post-processing step added to the logic which finds the matching platform for each of these platforms without concrete tags.  It finds the matching platform that does its data set and copies the necessary data from it.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -90,9 +90,14 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 }
             }
 
-            foreach (PlatformData platform in GetBuiltPlatforms())
+            IEnumerable<PlatformData> builtPlatforms = GetBuiltPlatforms();
+            List<PlatformData> platformsWithNoPushTags = new List<PlatformData>();
+
+            foreach (PlatformData platform in builtPlatforms)
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -103,9 +108,28 @@ private void PublishImageInfo()
...
 
+                if (!pushTags.Any())
+                {
+                    platformsWithNoPushTags.Add(platform);
+                }
+
                 platform.CommitUrl = _gitService.GetDockerfileCommitUrl(platform.PlatformInfo, Options.SourceRepoUrl);
             }
 
+            // Some platforms do not have concrete tags. In such cases, they must be duplicates of a platform in a different
```
MichaelSimons, line 119: Should this logic get tracked as something to cleanup as part of dotnet/docker-tools#664?

MichaelSimons (approved):

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
mthalman: Already is.  😄

PR #682, opened 2020-10-26
Support tag syndication

mthalman:
Updates the manifest schema to allow each tag to be syndicated to an additional repository.  That means the tag will be published both to the repository in which it's contained as well as the repository that it's configured to be syndicated to.  This supports dotnet/dotnet-docker#2317 by allowing, for example, the `2.1` tag to be published to both the `dotnet/core/runtime` and `dotnet/runtime` repositories.

This syndicated repository only applies to the publish logic.  Images are not tagged with these syndicated repositories at build time.  Specifically, it impacts the following commands:
* `copyAcrImages`
* `ingestKustoImageInfo`
* `publishManifest`
* `waitForMcrImageIngestion`

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
@@ -49,8 +49,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                         .SelectMany(platform => GetTagInfos(repo, platform))
                         .Select(tagInfo =>
                             ImportImageAsync(
-                                TrimRegistry(tagInfo.DestinationTag),
-                                TrimRegistry(tagInfo.SourceTag),
+                                DockerHelper.TrimRegistry(tagInfo.DestinationTag, Manifest.Registry),
+                                DockerHelper.TrimRegistry(tagInfo.SourceTag, Manifest.Registry),
                                 srcResourceId: resourceId)))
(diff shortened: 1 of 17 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
```diff
@@ -90,12 +104,16 @@ private IEnumerable<string> GetDestinationTagNames(RepoInfo repo, PlatformInfo p
...
+                tags.AddRange(platform.Tags
+                    .Select(tag => (GetSourceTag(tag.FullyQualifiedName), tag.FullyQualifiedName)));
             }
 
-            return destTagNames
-                .Select(tag => tag.TrimStart($"{Manifest.Registry}/"));
+            return tags;
         }
+
+        private string TrimRegistry(string tag) => tag.TrimStart($"{Manifest.Registry}/");
```
MichaelSimons, line 114: Consider putting this in a reusable helper class.  I know there is at least on other place that does the same - `GenerateImageGraphCommand`

MichaelSimons (approved):

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyAcrImagesCommand.cs
mthalman: With any luck, we'll eventually be using https://github.com/mthalman/DockerfileModel and get some nice image name parsing done for us.

PR #687, opened 2020-11-04
Update readme generation of full tags section

mthalman:
The logic to generate the `Full Tag Listing` section in the readme aggressively replaces all content between `# Full Tag Listing` and the next header.  Because of that it's not possible to include additional content to the `Full Tag Listing` section.  It will just be overridden by the replacement logic.  I've fixed this by only replacing the `# Full Tag Listing` header and any following whitespace.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
 using Sprache;
 
 namespace Microsoft.DotNet.ImageBuilder
@@ -18,9 +19,18 @@ namespace Microsoft.DotNet.ImageBuilder
 
             string targetLineEnding = readme.GetLineEndingFormat();
 
+            Parser<string> headerParser =
+                from headerPrefix in Parse.String(targetLineEnding + "# ").Text()
+                from label in Parse.AnyChar.Until(Parse.String(targetLineEnding)).Text()
+                select string.Concat(headerPrefix, label, targetLineEnding);
+
+            Parser<IEnumerable<char>> endOfGeneratedTagsParser =
+                Parse.String(EndOfGeneratedTagsMarker + targetLineEnding).XOr(headerParser);
+
             Parser<string> parser =
                 from leadingContent in Parse.AnyChar.Until(Parse.String(TagsSectionHeader + targetLineEnding)).Text()
-                from tagsContent in Parse.AnyChar.Until(Parse.String(EndOfGeneratedTagsMarker + targetLineEnding)).Text()
+                from tagsContent in Parse.AnyChar.Except(endOfGeneratedTagsParser).Many().Text()
+                from endOfGeneratedTags in endOfGeneratedTagsParser.Text()
                 from trailingContent in Parse.AnyChar.Many().Text()
                 select string.Concat(
                     leadingContent,
@@ -28,8 +38,7 @@ namespace Microsoft.DotNet.ImageBuilder
                     targetLineEnding,
                     targetLineEnding,
                     tagsListing,
-                    EndOfGeneratedTagsMarker,
-                    targetLineEnding,
+                    endOfGeneratedTags,
                     trailingContent);
 
             return parser.Parse(readme);
--- a/src/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
@@ -2,13 +2,14 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Text.RegularExpressions;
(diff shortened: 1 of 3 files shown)


MichaelSimons:
Our infrastructure is only used to generate the readmes that are checked into GH not DH.  MCR does this replacement when they syndicate the readmes to DH.  How do they handle this, I assume they do the same thing?

mthalman:
> Our infrastructure is only used to generate the readmes that are checked into GH not DH. MCR does this replacement when they syndicate the readmes to DH. How do they handle this, I assume they do the same thing?

I updated the mcrdocs content to test it out and it looks like the content is preserved.  So there should be no issue with these changes.

src/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
```diff
@@ -19,7 +19,7 @@ public static string UpdateTagsListing(string readme, string tagsListing)
             tagsListing = $"{TagsSectionHeader}{targetLineEnding}{targetLineEnding}{tagsListing}{targetLineEnding}";
 
             // Regex to find the entire tags listing section including the header.
-            Regex regex = new Regex($"^{TagsSectionHeader}\\s*(^(?!# ).*\\s)*", RegexOptions.Multiline);
```
MichaelSimons, line 22: I'm still not following how this will work.  When we publish the readmes to MCR, we start with the fully populated tags listing section (e.g. what is checked into GH).  This Regex still needs to strip out everything that's not "custom" content doesn't it?

mthalman: Ah, I was testing it with `GenerateReadmesCommand` where it was working.  But it wouldn't work for `PublishMcrDocsCommand`.

mthalman: What do you thinking of using a comment to denote the ending of the generated content?

The readme template could look something like this:

```markdown
# Full Tag Listing

<!--End of tags-->
For tags contained in the old dotnet/core{{if (PARENT_REPO = "nightly"):-nightly}}/{{SHORT_REPO}} repository, you can retrieve a list of those tags at https://mcr.microsoft.com/v2/dotnet/core{{if (PARENT_REPO = "nightly"):-nightly}}/{{SHORT_REPO}}/tags/list.
```

The generated content would look like this at the end:

```markdown
2.1.23-bionic-arm32v7, 2.1-bionic-arm32v7 | [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/runtime-deps/2.1/bionic/arm32v7/Dockerfile) | Ubuntu 18.04

You can retrieve a list of all available tags for dotnet/nightly/runtime-deps at https://mcr.microsoft.com/v2/dotnet/nightly/runtime-deps/tags/list.

<!-- End of tags-->
For tags contained in the old dotnet/core-nightly/runtime-deps repository, you can retrieve a list of those tags at https://mcr.microsoft.com/v2/dotnet/core-nightly/runtime-deps/tags/list.
```

MichaelSimons: That sounds like a good suggestion to me.  I think the `End of tags` label is a little confusing for those who may see it.  Maybe including "generated" in the content would help - `End of generated tags`

mthalman: Updated this to look for the comment marker.  Also switched it to use the Sprache library instead of regex for readability purposes.

src/Microsoft.DotNet.ImageBuilder/src/ReadmeHelper.cs
```diff
@@ -2,25 +2,37 @@
...
 
             string targetLineEnding = readme.GetLineEndingFormat();
-            tagsListing = $"{TagsSectionHeader}{targetLineEnding}{targetLineEnding}{tagsListing}{targetLineEnding}";
 
-            // Regex to find the entire tags listing section including the header.
-            Regex regex = new Regex($"^{TagsSectionHeader}\\s*(^(?!# ).*\\s)*", RegexOptions.Multiline);
-            return regex.Replace(readme, tagsListing);
+            Parser<string> parser =
+                from leadingContent in Parse.AnyChar.Until(Parse.String(TagsSectionHeader + targetLineEnding)).Text()
+                from tagsContent in Parse.AnyChar.Until(Parse.String(EndOfGeneratedTagsMarker + targetLineEnding)).Text()
```
MichaelSimons, line 23: I was originally thinking the `EndOfGeneratedTagsMarker` was going to be optional and would only be necessary in scenarios where custom content is needed.  The primary reason I was thinking this was that it would make this change a non-breaking change and is not required in all repos.

mthalman: So the logic would be to include up to this marker OR the next header is reached?

MichaelSimons: Yes, that's what I was thinking.

mthalman: Fixed

MichaelSimons (approved):

PR #697, opened 2020-11-16
Indicate Window Server type in display name

### LINKED ISSUE #676: osVersion stored in the versions information does not differentiate between nanoserver and windowservercore based images
This is required because the versions information is the basis for the image information used in the image metrics reporting for .NET

[Nano Server example](https://github.com/dotnet/versions/blob/master/build-info/docker/image-info.dotnet-dotnet-docker-nightly.json#L1875)
[Windows Server Core example](https://github.com/dotnet/versions/blob/master/build-info/docker/image-info.dotnet-dotnet-docker-nightly.json#L1951)

mthalman:
* Include the specific type of Windows Server in the display name of the OS.  
* Changes the `PlatformData.OsVersion` property to represent an identifier rather than display name.  This is important because this value is stored and used for comparison to check for things like cached images; we should be free to alter the display name without affecting this behavior.
* Makes some minor updates to the sorting of image info data in order to provide consistent output.  Currently, results can be ordered differently depending on the source of the data (Linux vs Windows).  

Fixes #676


src/Microsoft.DotNet.ImageBuilder/tests/IngestKustoImageInfoCommandTests.cs
```diff
@@ -245,14 +243,14 @@ public async Task SyndicatedTag()
...
             string expectedData =
-@"""jkl"",""amd64"",""Linux"",""Ubuntu 19.04"",""1.0.5"",""1.0/sdk/os/Dockerfile"",""repo1"",""2020-04-20 21:56:58""
-""jkl"",""amd64"",""Linux"",""Ubuntu 19.04"",""1.0.5"",""1.0/sdk/os/Dockerfile"",""repo2"",""2020-04-20 21:56:58""
-""t1"",""amd64"",""Linux"",""Ubuntu 19.04"",""1.0.5"",""1.0/sdk/os/Dockerfile"",""repo1"",""2020-04-20 21:56:58""
-""t1"",""amd64"",""Linux"",""Ubuntu 19.04"",""1.0.5"",""1.0/sdk/os/Dockerfile"",""repo2"",""2020-04-20 21:56:58""
-""t2"",""amd64"",""Linux"",""Ubuntu 19.04"",""1.0.5"",""1.0/sdk/os/Dockerfile"",""repo1"",""2020-04-20 21:56:58""
-""t2a"",""amd64"",""Linux"",""Ubuntu 19.04"",""1.0.5"",""1.0/sdk/os/Dockerfile"",""repo2"",""2020-04-20 21:56:58""
-""t2b"",""amd64"",""Linux"",""Ubuntu 19.04"",""1.0.5"",""1.0/sdk/os/Dockerfile"",""repo2"",""2020-04-20 21:56:58""
-""t3"",""amd64"",""Linux"",""Ubuntu 19.04"",""1.0.5"",""1.0/sdk/os/Dockerfile"",""repo1"",""2020-04-20 21:56:58""";
+@"""jkl"",""amd64"",""Linux"",""focal"",""1.0.5"",""1.0/sdk/os/Dockerfile"",""repo1"",""2020-04-20 21:56:58""
```
MichaelSimons, line 246: The kusto data should continue to contain the friendly display names.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -138,23 +138,30 @@ public bool IsInternalFromImage(string fromImage)
...
             string os = BaseOsVersion;
 
             if (Model.OS == OS.Windows)
             {
-                if (os.Contains("2016"))
+                string version = os.Split('-')[1];
+                if (os.StartsWith("nanoserver"))
                 {
-                    displayName = "Windows Server 2016";
+                    displayName = $"Windows Nano Server, version {version}";
```
MichaelSimons, line 149: The brand name is just `Nano Server`.  
```suggestion
                    displayName = $"Nano Server, version {version}";
```

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -201,10 +208,11 @@ public string GetOSDisplayName()
 
                     displayName = os.FirstCharToUpper();
                 }
-                else
-                {
-                    throw new InvalidOperationException($"The OS version '{os}' is not supported.");
-                }
+            }
+
+            if (displayName is null)
```
MichaelSimons, line 213: Is the change do initialize displayName to null and moving this fallback check necessary?  The previous pattern is beneficial because it provides compile time checking that all Windows paths set the display name.

src/Microsoft.DotNet.ImageBuilder/tests/IngestKustoImageInfoCommandTests.cs
mthalman: Fixed

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
mthalman: Fixed

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
mthalman: The alternative is to repeat the exception check.  I've made that change.  Let me know if that's what you prefer.  I'm open either way.

MichaelSimons (approved):

PR #710, opened 2020-11-27
Fix runtime-deps sharing

mthalman:
Publishing of images that share runtime-deps between different product versions is currently broken.  As an example, for the sharing of 3.1 and 5.0 runtime-deps for Focal, it results in the 3.1 images being published but not the 5.0 images.  This happens because the comparison logic of the `PlatformData` class doesn't take into account the product version of the image that it's contained within.  This leads to two different platforms (3.1 vs 5.0) as being considered equal when they are not.  Because of that `CopyAcrImagesCommand` ends up publishing runtime-deps 3.1 tags twice rather than the 3.1 tags followed by the 5.0 tags.

This has been broken since the release of 5.0.  It wasn't known at that time because of a specific scenario where this issue doesn't occur.  For the 5.0 release, the build cache trimmed out 3.1 changes but left 5.0 to be published because of the tag update.  So 5.0 runtime-deps published fine in that case.  It's the scenario where both 3.1 and 5.0 runtime-deps tags are being published that leads to the issue.


MichaelSimons (commented):
LGTM

PR #717, opened 2020-12-09
Remove pipeline support for Windows Server 1903

mthalman:
For check-base-image-updates.yml, I decided to replace 1903 with 2019 LTSC instead of 20H2 because 2019 LTSC will be around a lot longer and not require updating for a long time.  It doesn't matter what version the agent pool is anymore because we no longer pull images, only use the manifest tool.


MichaelSimons (approved):
Friendly reminder to deallocate the build machines once all dependencies are removed.

PR #722, opened 2021-01-04
Upgrade ImageBuilder to .NET 5

### LINKED ISSUE #691: Upgrade Image Builder to .NET 5
Image Builder is currently targeting .NET Core 3.1: https://github.com/dotnet/docker-tools/blob/daf226c1dcc2c1b7a33708d4673c19cac6febaab/src/Microsoft.DotNet.ImageBuilder/src/Microsoft.DotNet.ImageBuilder.csproj#L6

Now that .NET 5 has been released, it should really be updated to that version.

mthalman:
Image Builder tests currently fail because the infra is configured to install .NET 5.0 on the machine running the tests: https://github.com/dotnet/docker-tools/blob/fb70fa59252fa72773f248afce37d4fbb4d7cb8a/eng/common/Install-DotNetSdk.ps1#L40

But the projects are still targeting .NET Core 3.1, so this ends up with a test failure because the required runtime framework version is not available.  

Fixes #691

### DIFF
--- a/eng/file-pusher/Dockerfile
+++ b/eng/file-pusher/Dockerfile
@@ -1,7 +1,7 @@
 # This Dockerfile is intended to be built at the root of the repo.
 
 # build image
-FROM mcr.microsoft.com/dotnet/core/sdk:3.1-alpine AS build-env
+FROM mcr.microsoft.com/dotnet/sdk:5.0-alpine AS build-env
 
 WORKDIR /file-pusher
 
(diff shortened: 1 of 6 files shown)


MichaelSimons (approved):

mthalman:
Also updated File Pusher to use 5.0.

MichaelSimons (approved):

PR #723, opened 2021-01-04
Set TLS version when downloading .NET install script

mthalman:
Explicitly setting the TLS version to handle running this on machines that have a minimum TLS version configured.  See https://github.com/microsoft/dotnet-framework-docker/pull/488


MichaelSimons (approved):
Friendly reminder to backport to docker-tools (I'm sure you have this covered 😸).

PR #730, opened 2021-01-15
 Convert to newer version of System.CommandLine

### LINKED ISSUE #154: Replace dependency on System.CommandLine
ImageBuilder has a dependency on System.CommandLine for handling the parsing of the command line args.  This library is unsupported at this time.  We should find an alternative library to use.

mthalman:
The major design choice here was to introduce a separate class hierarchy, symbol builders, for the use of initialize the symbol objects that are used by the`System.CommandLine` API.  Using the old API, the option classes themselves would define these but that doesn't work with the new API because `System.CommandLine`actually instantiates the option classes through its binding and so the symbols need to be defined before instantiating the option classes.

Fixes #154


src/Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryOptions.cs
```diff
@@ -3,42 +3,60 @@
...
+            return base.GetCliOptions().Concat(
+                new Option[]
+                {
+                    (passwordOption = CreateOption<string?>("password", nameof(DockerRegistryOptions.Password),
+                        "Password for the Docker Registry the images are pushed to",
+                        resultArg =>
+                        {
+                            string password = resultArg.GetTokenValue();
+                            string? username = resultArg.FindResultFor(
+                                usernameOption ?? throw new InvalidOperationException("username option not set yet"))?.GetTokenValue();
```
MichaelSimons, line 35: Can you explain how this works?  I am not following how this exception is used?

MichaelSimons (commented):
How does the cli help work with these changes?  I tried this and don't get the command listing with `--help` and don't see command help with `<cmd> --help`.

Overall the changes look nice.  It would be nice if the SymbolsBuilders weren't necessary.  Having yet another type that has to redefine the Options hierarchy is not ideal.  I find the introduction of the term "Symbols" to be confusing. I also found it confusing that not all Options classes have a SymbolsBuilder. Would it make sense to create a SymbolsBuilder for all Option types? What do you think about naming the SymbolsBuilder to be in terms of Options?

mthalman:
CLI help is not working. It's somehow related to using the CommandBuilder. It may be a bug in the library. But I wanted to get this in without being delayed by the help. I intend to log an issue to address this later.

Symbols is the terminology of 'System.CommandLine' so I'm basing it from that. Options and arguments are both types of symbols in that API. I'm open to naming it OptionsBuilder if you're ok with that. Otherwise, we can leave it as is and follow up with work to rename the Options classes instead.

The Options classes that don't have corresponding builder classes are the composable ones, not the "root" command options. The need for a builder class didn't exist like for the command option classes because of things like the lack of an inheritance hierarchy. However, I'm open to defining builder classes for them so they are consistent.

src/Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryOptions.cs
mthalman: The `usernameOption` variable should always be set at this point but since it was initialized to null, the compiler deems it as still possibly having a null value.  In other words, this is making the compiler happy because of the use of NRT.  It should really be thought of as an Assert.  It's providing compile-time validation that a null variable value is handled by throwing in this case.

MichaelSimons:
> CLI help is not working. It's somehow related to using the CommandBuilder. It may be a bug in the library. But I wanted to get this in without being delayed by the help. I intend to log an issue to address this later.

I'm fine with tracking it with a separate issue.  I do consider this an important issue as the dev flow is hindered without this.  I personally rely on the help as I don't always remember the option names.

> Symbols is the terminology of 'System.CommandLine' so I'm basing it from that. Options and arguments are both types of symbols in that API. I'm open to naming it OptionsBuilder if you're ok with that. Otherwise, we can leave it as is and follow up with work to rename the Options classes instead.

My preference is to have some symmetry in the naming.  I realize there is value in basing the terminology on the library being used but I question the ROI in doing this work.  I tend towards renaming `SymbolsBuilder` to `OptionsBuilder` and declare it good enough.

>The Options classes that don't have corresponding builder classes are the composable ones, not the "root" command options. The need for a builder class didn't exist like for the command option classes because of things like the lack of an inheritance hierarchy. However, I'm open to defining builder classes for them so they are consistent.

I see value in using a consistent pattern so I vote to defining the builder classes in all cases.

src/Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryOptions.cs
MichaelSimons: That makes sense now.  The exception message is a little confusing because it made me think it was thrown as a result of the consumer not specifying the option.  `username option` is really referring to the local variable not the command line option.

mthalman:
> I see value in using a consistent pattern so I vote to defining the builder classes in all cases.

I've defined builder classes for everything now.  For consistency, I've also added a `GetCliArguments` method to `ManifestFilterOptionsBuilder` even though it doesn't define any args.  I figured it was better for the code to just assume that it could have args in the future if we configure it that way.

MichaelSimons (approved):

PR #737, opened 2021-01-21
Update ImageBuilder tag

mthalman:


MichaelSimons (commented):
LGTM.  

FYI, I ran a test build in nightly to validate the Version Info is correctly tracking the layers.  Everything looked good.

PR #739, opened 2021-01-21
Sub-options need setters

mthalman:
All the options classes that compose sub-options need to have setters defined for those properties.  Without the setter, the binding logic is unable to set them and the properties are left to their default values instead of the ones provided via command line.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
@@ -14,7 +14,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         public string RepoName { get; set; }
         public CleanAcrImagesAction Action { get; set; }
         public int Age { get; set; }
-        public ServicePrincipalOptions ServicePrincipal { get; } = new ServicePrincipalOptions();
+        public ServicePrincipalOptions ServicePrincipal { get; set; } = new ServicePrincipalOptions();
         public string Subscription { get; set; }
         public string ResourceGroup { get; set; }
         public string RegistryName { get; set; }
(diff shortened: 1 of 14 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
```diff
@@ -14,7 +14,7 @@ public class CleanAcrImagesOptions : Options
         public string RepoName { get; set; }
         public CleanAcrImagesAction Action { get; set; }
         public int Age { get; set; }
-        public ServicePrincipalOptions ServicePrincipal { get; } = new ServicePrincipalOptions();
+        public ServicePrincipalOptions ServicePrincipal { get; set; } = new ServicePrincipalOptions();
```
MichaelSimons, line 17: Is it necessary for the default initializer if the binder sets the value?

mthalman: It's useful for unit tests to have things initialized.  Otherwise, it'll require a lot of updates to test to ensure all these properties have been set.

MichaelSimons: It's rather unfortunate to have to initialize these for the production code path.  Is this worth circling back on and cleaning up after the builds are unblocked?

MichaelSimons (approved):

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
mthalman: There is a use for it outside of tests as well.  If no service principal options were specified in the command line args, the binding from `System.CommandLine` wouldn't apply and the setter would not be called.  So the initializer keeps it as a non-null value, keeping it semantically equivalent to the code that exist before the package upgrade.  I guess I don't see a problem with having an initializer.

PR #743, opened 2021-01-22
Add IngestKustoImageInfoCommand support for layers

MichaelSimons:
Related to #729

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/IngestKustoImageInfoCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/IngestKustoImageInfoCommand.cs
@@ -99,6 +99,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
                         for (int i = 0; i < platform.Layers.Count; i++)
                         {
+                            // TODO: Track layer size (currently set to 0) https://github.com/dotnet/docker-tools/issues/745
                             layerInfo.AppendLine(FormatLayerCsv(
                                 platform.Layers[i], 0, platform.Layers.Count - i, sha, platform, image, repo.Repo, timestamp));
                         }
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/IngestKustoImageInfoCommand.cs
```diff
@@ -80,33 +79,65 @@ private string GetImageInfoCsv()
...
+                                    imageInfo.AppendLine(
+                                       FormatImageCsv(destinationTag, platform, image, tag.SyndicatedRepo, timestamp));
                                 }
                             }
                         }
+
+                        for (int i = 0; i < platform.Layers.Count; i++)
+                        {
+                            layerInfo.AppendLine(FormatLayerCsv(
+                                platform.Layers[i], 0, platform.Layers.Count - i, sha, platform, image, repo.Repo, timestamp));
```
mthalman, line 103: Please add a comment stating that using `0` for the size is just a placeholder until that data becomes available.

src/Microsoft.DotNet.ImageBuilder/src/Commands/IngestKustoImageInfoCommand.cs
```diff
@@ -80,33 +79,65 @@ private string GetImageInfoCsv()
...
+            string timestamp) =>
+                $"\"{layerDigest}\",\"{size}\",\"{ordinal}\",{FormatImageCsv(imageDigest, platform, image, repo, timestamp)}";
+
+        private async Task IngestInfoAsync(string info, string table)
+        {
+            using MemoryStream stream = new();
+            using StreamWriter writer = new(stream);
+            writer.Write(info);
+            writer.Flush();
+            stream.Seek(0, SeekOrigin.Begin);
```
mthalman, line 134: Couldn't all these lines go inside the `if` block below?

MichaelSimons: I prefer to have the _dry run_ code path do as much as possible without actually doing the actual thing.

src/Microsoft.DotNet.ImageBuilder/src/Commands/IngestKustoImageInfoCommand.cs
MichaelSimons: Added note.

mthalman (approved):

MichaelSimons:
Waiting to merge this until the production Kusto layer table is created.

PR #751, opened 2021-02-04
Image Builder: use Docker creds when pulling from Docker Hub

mthalman:
Currently, Image Builder pulls images from Docker Hub using anonymous access.  This is problematic due to the rate limiting change (#613).  These changes update Image Builder commands to take Docker Hub creds and log in with them to ensure that any pulls are not subject to the rate limiting.

What I wanted to accomplish with these changes is a way to ensure that we don't accidentally introduce code which executes pulls (or manifest inspect requests) anonymously.  So I've consolidated things into the `DockerService` class, including the login/logout logic, which validates that any attempt to make a request to Docker Hub is done within the context of logged-in DH credentials.

For those commands that require access to Docker Hub, I've centrally located the logic to run the login/logout in the base `Command` class which wraps the entire execution of the command within the context of logged-in credentials.  This required refactoring the `ExecuteAsync` method so that it can be appropriately wrapped with this logic.  I have a commit which contains just this refactoring.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -22,7 +22,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     [Export(typeof(ICommand))]
     public class BuildCommand : DockerRegistryCommand<BuildOptions, BuildOptionsBuilder>
     {
-        private readonly IDockerService _dockerService;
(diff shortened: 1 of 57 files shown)


mthalman:
After offline discussion with @MichaelSimons, I'm going to be closing this PR because these changes should become unnecessary once support for tag mirroring (described in https://github.com/dotnet/docker-tools/issues/613) is completed.  So the plan is to abandon these changes and proceed with implementing the mirroring logic.

PR #756, opened 2021-02-05
Provide source credentials for copying base images

mthalman:
* Renames `CopyDockerHubBaseImagesCommand` to `CopyBaseImagesCommand`.
* Provides source credentials options on a per-registry basis that the command will pass along to the ACR Import operation.
* Dynamically determines source registry from the image tag.

Related to #613

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesOptions.cs
@@ -5,32 +5,25 @@
 using System.Collections.Generic;
 using System.CommandLine;
 using System.Linq;
-using Microsoft.Azure.Management.ContainerRegistry.Fluent.Models;
-using static Microsoft.DotNet.ImageBuilder.Commands.CliHelper;
 
 #nullable enable
 namespace Microsoft.DotNet.ImageBuilder.Commands
 {
     public class CopyBaseImagesOptions : CopyImagesOptions
     {
-        public IDictionary<string, ImportSourceCredentials> SourceCredentials { get; set; } =
-            new Dictionary<string, ImportSourceCredentials>();
+        public RegistryCredentialsOptions CredentialsOptions { get; set; } = new RegistryCredentialsOptions();
     }
 
     public class CopyBaseImagesOptionsBuilder : CopyImagesOptionsBuilder
     {
+        private readonly RegistryCredentialsOptionsBuilder _registryCredentialsOptionsBuilder =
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.ImageBuilder/src/DockerHelper.cs
```diff
@@ -190,6 +190,17 @@ public static Docker.Manifest InspectManifest(string image, bool isDryRun)
             return JsonConvert.DeserializeObject<Docker.Manifest>(manifest);
         }
 
+        public static string GetRegistry(string imageName)
+        {
+            string firstSegment = imageName.Substring(0, imageName.IndexOf("/"));
+            if (firstSegment.Contains("."))
```
MichaelSimons, line 196: According to this, a '.' or ':' are used to identify a registry - https://stackoverflow.com/a/37867949/3182557

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public class CopyBaseImagesOptionsBuilder : CopyImagesOptionsBuilder
+    {
+        public override IEnumerable<Option> GetCliOptions() =>
+            base.GetCliOptions()
+                .Concat(new Option[]
+                {
+                    CreateDictionaryOption("source-creds", nameof(CopyBaseImagesOptions.SourceCredentials),
+                        "Named credentials that map to a source registry ((<registry>=<username>;<password>)",
```
MichaelSimons, line 27: On one hand I like this short hand but on the other it introduces some inconsistency in the CLI specifically with the creds for the destination registry.  Would it be confusing to move this option to the DockerRegistryOptions as a single registry creds option?  This represents the creds for any registry/registries a command operates on.  Maybe the registry could be considered optional.

"Named credentials for a registry ([<registry>=]<username>;<password>)"

MichaelSimons, line 27:
```suggestion
                        "Named credentials that map to a source registry (<registry>=<username>;<password>)",
```

mthalman: I've refactored things out into a composable `RegistryCredentialsOptions` class.  In order to keep things simple, I've made the registry portion of it required in the syntax.

MichaelSimons (approved):
I like the single credentials options.

PR #760, opened 2021-02-12
Copy base images using subscriptions

mthalman:
Updates the `copyBaseImages` command to optionally consume a subscriptions file instead of a local manifest.  This is needed in the scenario where this command will be run as part of checking for any base image updates, which occurs from the dotnet/docker-tools repo. In that case, the subscription is used to determine all the manifests from all the product repos that are participating in the base image check logic and copies those base images to the mirroring location.  

Related to #613

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesCommand.cs
@@ -46,11 +46,10 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         {
             LoggerService.WriteHeading("COPYING IMAGES");
 
-            IEnumerable<Task> copyImageTasks;
+            IEnumerable<ManifestInfo> manifests;
             if (Options.SubscriptionOptions.SubscriptionsPath is null)
             {
-                copyImageTasks = GetFromImages(Manifest)
-                    .Select(fromImage => CopyImageAsync(fromImage, GetBaseRegistryName(Manifest.Registry)));
+                manifests = new ManifestInfo[] { Manifest };
             }
             else
             {
@@ -59,15 +58,17 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                     throw new InvalidOperationException($"{Options.RegistryOverride} must be set.");
                 }
 
-                IEnumerable<(Subscription Subscription, ManifestInfo Manifest)> subscriptionManifests =
-                    await SubscriptionHelper.GetSubscriptionManifestsAsync(
+                manifests =
+                    (await SubscriptionHelper.GetSubscriptionManifestsAsync(
                         Options.SubscriptionOptions.SubscriptionsPath, Options.FilterOptions, _httpClient,
-                        options => options.RegistryOverride = Options.RegistryOverride);
-                copyImageTasks = subscriptionManifests
-                    .SelectMany(subscriptionManifest => GetFromImages(subscriptionManifest.Manifest))
-                    .Distinct()
-                    .Select(fromImage => CopyImageAsync(fromImage, GetBaseRegistryName(Options.RegistryOverride)));
+                        options => options.RegistryOverride = Options.RegistryOverride))
+                    .Select(subscriptionManifest => subscriptionManifest.Manifest);
             }
+
+            IEnumerable<Task> copyImageTasks = manifests
+                .SelectMany(manifest => GetFromImages(manifest))
+                .Distinct()
+                .Select(fromImage => CopyImageAsync(fromImage, GetBaseRegistryName(Manifest.Registry)));
 
             await Task.WhenAll(copyImageTasks);
         }
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesCommand.cs
@@ -2,12 +2,15 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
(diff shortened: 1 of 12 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/SubscriptionOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+        public string? SubscriptionsPath { get; set; }
+    }
+
+    public class SubscriptionOptionsBuilder
+    {
+        public IEnumerable<Option> GetCliOptions() =>
+            new Option[]
+            {
+                CreateOption<string?>("subscriptions-path", nameof(SubscriptionOptions.SubscriptionsPath),
+                    $"Path to the subscriptions file")
```
MichaelSimons, line 23: Curious why the default was removed.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesCommand.cs
```diff
@@ -2,50 +2,91 @@
...
+                    throw new InvalidOperationException($"{Options.RegistryOverride} must be set.");
+                }
+
+                copyImageTasks =
+                    (await SubscriptionHelper.GetSubscriptionManifestsAsync(
+                        Options.SubscriptionOptions.SubscriptionsPath, Options.FilterOptions, _httpClient,
+                        options => options.RegistryOverride = Options.RegistryOverride))
+                    .SelectMany(subscriptionManifest => GetFromImages(subscriptionManifest.Manifest))
+                    .Distinct()
+                    .Select(fromImage => CopyImageAsync(fromImage, GetBaseRegistryName(Options.RegistryOverride)));
```
MichaelSimons, line 67: Consider refactoring so that the there is a single code path the operates on the manifests.  I think this would make the code a little easier to follow on first read.

1. get manifests via if else
2. manifests.SelectMany(GetFromImages).Select(CopyImageAsync)

MichaelSimons (approved):

src/Microsoft.DotNet.ImageBuilder/src/Commands/SubscriptionOptions.cs
mthalman: The goal is to allow this command to be used in the context of a build job.  In that case, the logic is determined by the manifest, not the subscription file.  The logic in `CopyBaseImagesCommand` uses this option value to determine whether it should use a subscription file or the manifest.  So we wouldn't wait it to have a default value in the case when a manifest is intended to be used.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesCommand.cs
mthalman: I made some changes.  Is that what you were thinking?

MichaelSimons: Partially.  

```
            string destinationRegistry;
            IEnumerable<ManifestInfo> manifests;
            if (Options.SubscriptionOptions.SubscriptionsPath is null)
            {
                destinationRegistry = GetBaseRegistryName(Manifest.Registry);
                manifests = new[] { Manifest };
            }
            else
            {
                if (Options.RegistryOverride is null)
                {
                    throw new InvalidOperationException($"{Options.RegistryOverride} must be set.");
                }

                destinationRegistry = GetBaseRegistryName(Options.RegistryOverride);
                manifests =
                    (await SubscriptionHelper.GetSubscriptionManifestsAsync(
                        Options.SubscriptionOptions.SubscriptionsPath, Options.FilterOptions, _httpClient,
                        options => options.RegistryOverride = Options.RegistryOverride))
                    .Select(t => t.Manifest);
            }

            IEnumerable<Task> copyImageTasks = manifests
                .SelectMany(manifest => GetFromImages(manifest))
                .Distinct()
                .Select(fromImage => CopyImageAsync(fromImage, destinationRegistry));
            await Task.WhenAll(copyImageTasks);

File: src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesCommand.cs [hunk 3]
```diff
@@ -2,50 +2,93 @@
...
             LoggerService.WriteHeading("COPYING IMAGES");
 
-            IEnumerable<Task> importTasks = Manifest.GetExternalFromImages()
-                .Where(fromImage => !fromImage.StartsWith(Manifest.Registry))
-                .Select(fromImage =>
+            IEnumerable<Task> copyImageTasks;
+            if (Options.SubscriptionOptions.SubscriptionsPath is null)
+            {
+                copyImageTasks = GetFromImages(Manifest)
+                    .Select(fromImage => CopyImageAsync(fromImage, GetBaseRegistryName(Manifest.Registry)));
```
MichaelSimons, line 53: Should this be using the RegistryOverride for the destination?

mthalman: It's using `ManifestInfo` so its `Registry` property reflects the resolved value that takes into account `RegistryOverride`.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesCommand.cs
mthalman: Ok, I see what you're saying now.  Fixed.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyBaseImagesCommand.cs
MichaelSimons: Makes sense.

MichaelSimons (approved):

PR #764, opened 2021-02-18
Support pulling from mirror location during build

mthalman:
This updates the `build` command so that it can handle pulling base images from a specified mirror location instead of directly from an externally-owned registry like Docker Hub.

This defines an additional option, `source-repo-prefix`, to the `build` command. When that option is set, it pulls any external base images by constructing a tag that represents the mirror location.

For example, consider the following scenario:

* Dockerfile has `FROM amd64/alpine:3.13`
* `source-repo-prefix` is set to `mirror/`
* `registry-override` is set to `dotnetdocker.azurecr.io`

When pulling base images the `build` command will pull from `dotnetdocker.azurecr.io/mirror/amd64/alpine:3.13` instead of from `amd64/alpine:3.13`.  It then tags the pulled image as `amd64/alpine:3.13` so that the Dockerfile can be built directly without modification.

Related to #613

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -624,12 +624,14 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         /// </remarks>
         private string GetFromImageTag(string fromImage, string? registry)
         {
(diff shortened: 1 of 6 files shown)


mthalman:
Depending on the approval of https://github.com/dotnet/docker-tools/pull/765, this will need to be modified to take those changes into account.

MichaelSimons (approved):

PR #774, opened 2021-03-05
Refactor image-info layer retrieval logic

### LINKED ISSUE #773: Layer image-info information is incorrect
There are two issues with the layer information stored in the image-info.

1. The wrong digests are being stored.  The digests that are stored are obtained from `docker inspect`.  We should be storing the manifest digest obtained from the registry.
2. The layer information is getting merged into the image-info rather than replacing the current data - example https://github.com/dotnet/versions/blob/master/build-info/docker/image-info.dotnet-docker-tools-main-imagebuilder.json#L29.

MichaelSimons:
Fixed #773 

I have validated this e2e with a test build - https://dev.azure.com/dnceng/internal/_build/results?buildId=1025102&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=78db2542-c627-4140-8a7a-d06178fff4e4 (internal link)

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
@@ -53,7 +53,7 @@ namespace Microsoft.DotNet.ImageBuilder
             return digest;
         }
 
-        public IEnumerable<string> GetImageLayers(string image, bool isDryRun) => DockerHelper.GetImageLayers(image, isDryRun);
+        public IEnumerable<string> GetImageLayers(string image, bool isDryRun) => _manifestToolService.GetImageLayers(image, isDryRun);
 
         public void PullImage(string image, bool isDryRun) => DockerHelper.PullImage(image, isDryRun);
 
--- a/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
@@ -53,7 +53,7 @@ namespace Microsoft.DotNet.ImageBuilder
             return digest;
         }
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
```diff
@@ -53,7 +53,7 @@ public string GetImageDigest(string image, bool isDryRun)
             return digest;
         }
 
-        public IEnumerable<string> GetImageLayers(string image, bool isDryRun) => DockerHelper.GetImageLayers(image, isDryRun);
+        public IEnumerable<string> GetImageLayers(string image, bool isDryRun) => _manifestToolService.GetImageLayers(image, isDryRun);
```
mthalman, line 56: Wouldn't it be better to have the caller invoked `ManifestToolService` directly rather than through `DockerService`?  I imagine there also being an implementation of `GetImageLayers` just as you had it before that relied on `docker inspect`. But in this case the caller explicitly wants the semantics/implementation provided via the manifest command/tool.  So to me, it would make more sense to have the caller explicitly reference `ManifestToolService`.

MichaelSimons: I want the caching infrastructure provided by the `DockerService`.  In ways, I think the ManifestToolService could be combined with the DockerService.  How about just renaming the method to `GetImageManifestLayers`?

mthalman: > How about just renaming the method to GetImageManifestLayers?

That sounds good to me.

MichaelSimons: Fixed

mthalman (approved):

PR #777, opened 2021-03-11
Common component governance detection job

mthalman:
The CG component detection task that gets [executed in the publish job](https://github.com/dotnet/docker-tools/blob/main/eng/common/templates/jobs/publish.yml#L105-L106) isn't actually doing what we think it's doing.  It was thought that this could just scan the source code and detect components.  But it actually requires the source to be built, at least it does to detect NuGet packages that are referenced from projects.  So it's not actually reporting any NuGet package references from our repos.

Another problem is that we build all of our jobs within a container.  And the component detection task is run on the agent where it scans the agent's source directory.  Since the project is built in a container, it's outside the context of the agent's file system and the component detection will not find anything in that case either.

To solve this problem, I've defined a common job template that can be used across the repos.  It will dynamically build all projects in the repo directly on the build agent and then run the component detection task.  This will allow reporting of _all_ components that are referenced within the repos.

In some cases, there are projects that we don't want to build such as test projects.  In that case, the pipeline can set the common `cgBuildGrepArgs` to set argument values to the `grep` command that can filter out project file paths as necessary.

### DIFF
--- a/eng/common/templates/steps/test-images-linux-client.yml
+++ b/eng/common/templates/steps/test-images-linux-client.yml
@@ -81,5 +81,3 @@ steps:
   continueOnError: true
 - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
   - template: cleanup-docker-linux.yml
-- task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
-  displayName: Component Detection
--- a/eng/common/templates/steps/test-images-linux-client.yml
+++ b/eng/common/templates/steps/test-images-linux-client.yml
@@ -81,3 +81,5 @@ steps:
   continueOnError: true
 - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
   - template: cleanup-docker-linux.yml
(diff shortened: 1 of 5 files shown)


eng/common/templates/steps/test-images-linux-client.yml
```diff
@@ -81,3 +81,5 @@ steps:
   continueOnError: true
 - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
   - template: cleanup-docker-linux.yml
+- task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
```
MichaelSimons, line 84: How long does this task take?  Is there a way to only run this once instead of for each test leg?

mthalman: That wasn't supposed to be there.  Left over from some testing I was doing.  I've removed it.

The task doesn't take long to run.  For example, it ran in 11 sec to scan all three projects in this repo.

MichaelSimons (approved):

Issue #394, opened 2020-01-28
Implement a monitoring solution for image publishing


MichaelSimons:
The only option that exists today for monitoring the infrastructure used during image production, is to subscribe to the AzDO pipeline notifications.  This is not a user friendly solution as you need to inspect the builds to see things like the queue parameters, etc.  This has to be done for numerous different builds.  It is difficult and time consuming to do this type of monitoring and as a result, I postulate we are not doing the best of job.  A recent occurrence which we missed via monitoring is https://github.com/dotnet/docker-tools/pull/389.

To solve this problem, it would be nice to have a monitoring solution that would give a high level overview of what is happening.  Examples include:

* Base image change notification and the images that need to get rebuilt as a result.
* Image publishing notification
* ACR cleanup notification - how much space was freed and how much space is being used.
* etc.

I don't think an elaborate system is needed.  It should be simple and cheap to build (e.g email a DL).  It can be refactored and expanded as necessary.

MichaelSimons:
We feel this is an important issue to address as we should have more visibility into what is happening within the builds.  Assigning to @mthalman to investigate sending a notification/email from within the build.

mthalman:
AzDO has no native solution for sending emails and dnceng isn't aware of any existing pipelines in the org that do this.  Our best bet might be to make use of [SendGrid](https://docs.microsoft.com/en-us/azure/sendgrid-dotnet-how-to-send-email) in Azure.  They have a free service tier that should be sufficient.

mthalman:
[Triage]
The current solution we have for notifying of publishing and rebuilds gives us visibility into the state of the image publishing infrastructure. For image cleanup, the clean up pipeline also reports on what has been cleaned up. At this point, investing in anything beyond this isn't worth it.

PR #790, opened 2021-05-25
Add build notifications for AutoBuilder

mthalman:
This implements a basic notification service within Image Builder that allows notifications to be posted as issues in a GitHub repo.  The use of the service will be expanded in the future but for now I've configured it so that a notification will be sent whenever AutoBuilder queues a build.

Related to https://github.com/dotnet/docker-tools/issues/394

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/NotificationLabels.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/NotificationLabels.cs
@@ -1,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public static class NotificationLabels
+    {
+        public const string AutoBuilder = "autobuilder";
+        public const string Failure = "failure";
+    }
+}
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/NotificationLabels.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/NotificationLabels.cs
@@ -5,7 +5,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 {
     public static class NotificationLabels
     {
-        public const string AutoBuilder = "autobuilder";
-        public const string Failure = "failure";
+        private const string NotifyPrefix = "notify-";
+        public const string AutoBuilder = NotifyPrefix + "autobuilder";
+        public const string Failure = NotifyPrefix + "failure";
     }
 }
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
@@ -199,9 +199,19 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             string header = $"AutoBuilder - {category}";
             notificationMarkdown.Insert(0, $"# {header}{Environment.NewLine}{Environment.NewLine}");
 
-            await _notificationService.PostAsync($"{header} - {subscription}", notificationMarkdown.ToString(),
-                new string[] { NotificationLabels.AutoBuilder }.AppendIf(NotificationLabels.Failure, () => exception is not null),
-                $"https://github.com/{Options.GitOptions.Owner}/{Options.GitOptions.Repo}", Options.GitOptions.AuthToken);
+            if (Options.GitOptions.AuthToken == string.Empty ||
+                Options.GitOptions.Owner == string.Empty ||
+                Options.GitOptions.Repo == string.Empty)
+            {
+                _loggerService.WriteMessage(
+                    "Skipping posting of notification because GitHub auth token, owner, and repo options were not provided.");
+            }
+            else
+            {
+                await _notificationService.PostAsync($"{header} - {subscription}", notificationMarkdown.ToString(),
+                    new string[] { NotificationLabels.AutoBuilder }.AppendIf(NotificationLabels.Failure, () => exception is not null),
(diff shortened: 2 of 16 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
```diff
@@ -95,38 +100,96 @@ private async Task QueueBuildForStaleImages(Subscription subscription, IEnumerab
                 return;
             }
 
-            (Uri baseUrl, VssCredentials credentials) = Options.AzdoOptions.GetConnectionDetails();
+            StringBuilder notificationMarkdown = new();
```
MichaelSimons, line 103: To me, it's not ideal that the notification logic is so ingrained with the core logic of the command.  I think it would be beneficial if this could be separated.  I think this would help the readability of the logic.  While I realize separating this entirely is not possible, the more it is separated the easier it may be to enable notifications in more places (e.g. the try catch finally pattern).  I also wonder if there isn't an opportunity to share/integrate to logging and notification services in some way.  For example when calling one, could a flag be passed so that the output is captured by both?

MichaelSimons (commented):
Can you provide a sample of an issue created from this?  It would be helpful to see what the "output" looks like.

src/Microsoft.DotNet.ImageBuilder/src/NotificationService.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.DotNet.ImageBuilder
+{
+    [Export(typeof(INotificationService))]
+    public class NotificationService : INotificationService
+    {
+        public async Task<Uri> PostAsync(string title, string description, IEnumerable<string> labels, string repoUrl, string gitHubAccessToken)
+        {
+            IssueManager issueManager = new(gitHubAccessToken);
+            int issueId = await issueManager.CreateNewIssueAsync(repoUrl, title, description, labels: labels);
```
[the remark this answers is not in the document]
mthalman: It automatically gets created.

src/Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
mthalman: I've factored out the logging and notification logic into a separate method.  I don't see a great way to integrate logging with the notifications mainly because of the formatting necessary for notifications (Markdown) is unnecessary and ugly when used in the context of log output.

src/Microsoft.DotNet.ImageBuilder/src/Commands/NotificationLabels.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public static class NotificationLabels
+    {
+        public const string AutoBuilder = "autobuilder";
```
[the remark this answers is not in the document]
mthalman: I've added a `notify` prefix to the labels to indicate the source of these issues are from the notification system.

mthalman:
> Can you provide a sample of an issue created from this? It would be helpful to see what the "output" looks like.

Here are examples:

* https://github.com/dotnet/dotnet-docker-internal/issues/8
* https://github.com/dotnet/dotnet-docker-internal/issues/7
* https://github.com/dotnet/dotnet-docker-internal/issues/9

src/Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
```diff
@@ -95,38 +100,119 @@ private async Task QueueBuildForStaleImages(Subscription subscription, IEnumerab
...
             }
+            else
+            {
+                throw new NotSupportedException("Unknown state");
+            }
+
+            string header = $"AutoBuilder - {category}";
+            notificationMarkdown.Insert(0, $"# {header}{Environment.NewLine}{Environment.NewLine}");
+
+            await _notificationService.PostAsync($"{header} - {subscription}", notificationMarkdown.ToString(),
```
MichaelSimons, line 202: Is there an opt out mechanism with the notification service?  I'm not seeing one.  It feels like having an option (e.g. by not specifying the git options) would be worthwhile.  I could envision needing to disable it temporarily or maybe folks like IIS would not want it (at least initially).

src/Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildOptions.cs
```diff
@@ -13,13 +13,15 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     public class QueueBuildOptions : Options
     {
         public string SubscriptionsPath { get; set; } = string.Empty;
-        public AzdoOptions AzdoOptions { get; set; } = new AzdoOptions();
+        public AzdoOptions AzdoOptions { get; set; } = new();
         public IEnumerable<string> AllSubscriptionImagePaths { get; set; } = Enumerable.Empty<string>();
+        public GitOptions GitOptions { get; set; } = new();
```
MichaelSimons, line 18: What's the UX like for understanding what these GitOptions are used for (e.g. they are for the git repo used to post notifications for)?

mthalman: I've redesigned things to provide a better UX in help.  Previously, there were options being registered that were not even being used by the command.  That's fixed now to only include the necessary options.  And custom descriptions specific to notifications are now provided for these options as well.

src/Microsoft.DotNet.ImageBuilder/src/Commands/QueueBuildCommand.cs
mthalman: Fixed

MichaelSimons (approved):

Issue #798, opened 2021-06-14
New Windows Release - Windows Server 2022


mthalman:
# New Windows Release

Windows version: Windows Server 2022

## Tasks

1. - [x] Well before the Windows release date, contact DDFun to schedule the provisioning of an Azure scale set for the new Windows version: https://portal.microsofticm.com/imp/v3/incidents/details/253843530/home
2. - [x] If this is an LTS release of Windows, update [ImageBuilder](https://github.com/dotnet/docker-tools/blob/main/src/Microsoft.DotNet.ImageBuilder/src/McrTagsMetadataGenerator.cs) code to generate the correct README display name from the version specified in the manifest.
3. - [x] Include additional build and test jobs in the [common pipeline](https://github.com/dotnet/docker-tools/blob/main/eng/common/templates/stages/build-test-publish-repo.yml) to support the new Windows version.
4. - [x] Add support for producing an image of Image Builder for the new version of Windows by updating Image Builder's manifest.json file.

PR #799, opened 2021-06-14
Add display name for Windows Server 2022

mthalman:
Related to https://github.com/dotnet/docker-tools/issues/798


src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -158,6 +158,10 @@ public string GetOSDisplayName()
                     {
                         displayName = "Windows Server Core 2019";
                     }
+                    else if (version == "ltsc2022")
```
MichaelSimons, line 161: Would it make sense to refactor this to handle all ltsc versions with a single piece of logic?

```
if (version.startswith("ltsc")
{
    displayName = $"Windows Server Core {version.trimstart(ltsc))}";
}
else
{
    displayName = $"Windows Server Core, version {version}";
}
```

MichaelSimons (approved):

PR #801, opened 2021-06-21
Update version of Image Builder

mthalman:


mthalman:
Need to make another update to the display name logic.

PR #820, opened 2021-07-09
Automatically create PR to reference new Image Builder tag

### LINKED ISSUE #226: Auto-create PR to update reference to ImageBuilder's image tag
Scenario:
1. I make a change to the Image Builder code and merge the changes into the repo.
2. docker-tools-official build runs and builds the Docker image for Image Builder which generates a unique tag.
3. I want the pipeline code in docker-tools to reference this new tag in the common docker-images.yml file.  I have to wait for the build in step 2 to finish, grab the tag, and create a new PR that updates the file.

Step 3 should automatically be done by the build in step 2.  After building the image, it should automatically create a PR that updates that file so that it can be verified and merged by a human.

mthalman:
Defines a pipeline which updates the referenced Image Builder tag variable with a new tag value.  This pipeline gets triggered after the completion of the Image Builder pipeline.

Fixes #226

### DIFF
--- a/eng/common/templates/jobs/publish.yml
+++ b/eng/common/templates/jobs/publish.yml
@@ -21,9 +21,9 @@ jobs:
       artifactName: image-info
   - template: ../steps/set-image-info-path-var.yml
   - template: ../steps/set-dry-run.yml
-  - script: echo $(sourceBuildId) > $(Build.ArtifactStagingDirectory)/source-build-id.json
+  - script: echo $(sourceBuildId) > $(Build.ArtifactStagingDirectory)/source-build-id.txt
     displayName: Write Source Build ID to File
-  - publish: $(Build.ArtifactStagingDirectory)/source-build-id.json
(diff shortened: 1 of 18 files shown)


MichaelSimons (approved):
LGTM - Nice improvement.

eng/common/templates/jobs/publish.yml
```diff
@@ -21,6 +21,11 @@ jobs:
       artifactName: image-info
   - template: ../steps/set-image-info-path-var.yml
   - template: ../steps/set-dry-run.yml
+  - script: echo $(sourceBuildId) > $(Build.ArtifactStagingDirectory)/source-build-id.json
```
MichaelSimons, line 24: Does this file just contain the build id?  Is it a valid json file?  It feels wrong to use the json suffix if it isn't.

mthalman: Not sure what I was thinking there.  I guess I just love JSON.  Fixed.

PR #824, opened 2021-07-22
Scripts to check package updates

mthalman:
Adding a set of shell scripts to handle gathering the data within a container to determine whether any of it's packages require an upgrade. This will be used in order to support https://github.com/dotnet/dotnet-docker/issues/1455.

The first script that will be used in the workflow is `get-installed-packages.sh`.  This will be called within a container to determine which packages are installed in it.  By executing that script between a base image, such as Ubuntu Focal, and a derived image, such as `runtime-deps:focal`, it is possible to determine the closure of packages that were installed on top of the base image. In other words, it let's us determine all the packages that were caused to be installed by our Dockerfile (the full closure of packages, dependencies and all).

That set of package names will then be fed into the next script, `get-upgradable-packages.sh`. This is a POSIX-based wrapper script around a Bash-based script to ensure operability in a variety of Linux distros that may not have Bash installed by default.  It then determines which scripts are upgradable according to the rules being defined for the container's particular package manager. In the case of `apt`, used by Debian and Ubuntu, we'll determine whether package upgrades are relevant by checking whether they come from the security package repository. For `apk` (Alpine) and `tdnf` (CBL-Mariner), there is no such repository distinction. So it'll check for whether any package upgrade exists at all.  Those list of packages that have upgrades are intersected with the list of package names that were passed to the script. That gives us the list of packages that are upgradable and were caused to be installed by our Dockerfile. It then outputs that into a comma-delimited file with the format `<package-name>,<current-version>,<upgrade-version`.

These scripts will be executed by Image Builder as part of its `GetStaleImages` command in order to provide additional data to determine whether an image should be rebuilt.


eng/get-upgradable-packages.bash.sh
```diff
@@ -0,0 +1,174 @@
...
+    apt update
+
+    grep security /etc/apt/sources.list > /tmp/security.list
+
+    # Find all upgradable packages from security feeds and store the output in an array
+    mapfile -t aptPackages < <(apt upgrade -oDir::Etc::Sourcelist=/tmp/security.list -s 2>/dev/null | grep Inst | sort)
+
+    # Regex to parse the output from apt to get the package name, current version, and upgrade version
+    regex="Inst\s(\S+)\s\[(\S+)]\s\((\S+)\s"
+    upgradablePackages=()
```
MichaelSimons, line 22: What's this for?

eng/get-upgradable-packages.bash.sh
```diff
@@ -0,0 +1,174 @@
...
+    echo "Updating package cache..."
+    apk update
+
+    # Find all installed package names and store the output in an array
+    mapfile -t apkPackageNames < <(apk info | sort)
+
+    # Find all upgradable packages
+    apkPackages=$(apk version | tail -n +2 | sort)
+
+    upgradablePackages=()
```
MichaelSimons, line 51: What's this for?

eng/get-upgradable-packages.bash.sh
```diff
@@ -0,0 +1,174 @@
...
+    echo
+    echo "Upgradable packages:"
+    for pkg in "${aptPackages[@]}"
+    do
+        if [[ $pkg =~ $regex ]]; then
+            pkgName=${BASH_REMATCH[1]}
+            currentVersion=${BASH_REMATCH[2]}
+            upgradeVersion=${BASH_REMATCH[3]}
+            
+            local versionData=$(echo "$currentVersion,$upgradeVersion")
```
MichaelSimons, line 33: These three lines are shared across all package manager types.  Consider encapsulating in a shared "AddUpgradablePackage" function

eng/get-upgradable-packages.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env sh
+
+# This is the main POSIX-compliant script for retrieving the list of upgradable packages. It's just a wrapper around
```
MichaelSimons, line 3: Curious why bash is required?  Does the script get too complex/difficult written with sh?

eng/get-upgradable-packages.bash.sh
```diff
@@ -0,0 +1,174 @@
...
+outputPath="$1"
+args=( $@ )
+
+packagesToCheck=( "${args[@]:1}" )
+
+declare -A upgradablePackageVersions
+upgradablePackageVersions=()
+
+if type apt > /dev/null 2>/dev/null; then
+    getUpgradablePackageVersions "Apt"
```
MichaelSimons, line 159: Curious the value of this helper method versus inlining the call directly to getUpgradablePackageVersionsForApt?  It doesn't seem to be saving any complexity/lines of code.  I would argue it hurts the readability a little.

eng/get-upgradable-packages.sh
mthalman: The biggest thing that bash provides is support for arrays.

eng/get-upgradable-packages.bash.sh
mthalman: Unneeded.  Deleted.

eng/get-upgradable-packages.bash.sh
mthalman: Unneeded.  Deleted.

mthalman:
I've converted this to a draft because it's going to require some more changes.  After further testing, I've come to the realization that my original plan won't be feasible because of execution time.  It would require pulling all supported images and executing these scripts which ends up taking quite a long time.  Plus, it's wasteful to be running that every 4 hrs as our current scanning pipeline is configured to run.

I'm going to work on a different approach to mitigate these issues.

mthalman:
I've changed the scripts to take a different approach that doesn't require running the scripts on a container of every single image we publish.  The proposed solution now is to store image info metadata of which packages are installed for each of the images we publish.  This includes the versions of the packages.  Once we have that list, we no longer need the published image to determine whether a package upgrade is available.  All we need is the external base image that we can run scripts on using the stored metadata.

As an example, let's say we publish an image based on Alpine that has installed curl 7.78.0-r0.  That package and version info is stored in the image info file.  Later when determining whether the image has a current version of curl, we just need to run a base Alpine container and check whether curl 7.78.0-r0 is the latest or not.

mthalman:
I've been considering how to best handle the scenario of checking package updates for distroless containers. Obviously, the scripts defined in this PR will not work in a distroless container due to the lack of shell and package manager. But it would be possible to do with a set of scripts specific to each distroless implementation.  This PR will be sufficient for handling the standard, non-distroless, containers. There will need to be additional custom scripts to handle distroless.

eng/common/package-scripts/get-installed-packages.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env sh
+
+# This is the main script for retrieving the list of installed packages.
+
+imageTag=$1
+dockerfilePath=$2
+scriptDir=$(dirname $(realpath $0))
+
+docker run --rm -v $scriptDir:/scripts $imageTag /bin/sh /scripts/get-installed-packages.container.sh
```
MichaelSimons, line 9: A volume mount doesn't seem necessary here.  You can read the script and pass it into the docker run command.

`docker run -i --rm $imageTag /bin/sh < get-installed-packages.container.sh`

Note: the usage of the `-i` option otherwise you won't get the output.

eng/common/package-scripts/get-upgradable-packages.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env sh
+
+# This is the main script for retrieving the list of upgradable packages.
+
+imageTag=$1
+dockerfilePath=$2
```
MichaelSimons, line 6: Is this necessary?

eng/common/package-scripts/get-upgradable-packages.sh
```diff
@@ -0,0 +1,19 @@
...
+outputPath=$3
+
+shift 3
+packagelist=$@
+
+scriptDir="$(dirname $(realpath $0))"
+containerName="$(uuidgen)"
+containerOutputPath="/$containerName.txt"
+
+docker run --name $containerName -v $scriptDir:/scripts $imageTag /bin/sh -c "/scripts/get-upgradable-packages.container.sh $containerOutputPath $packagelist"
```
MichaelSimons, line 16: Same comment as earlier, the volume mount doesn't seem necessary.

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
```diff
@@ -0,0 +1,256 @@
...
+if type tdnf > /dev/null 2>/dev/null; then
+    echo "Updating package cache..."
+    tdnf makecache 1>/dev/null
+
+    for pkgName in "${packages[@]}"
+    do
+        installPackageWithTdnf $pkgName
+    done
+
+    tdnf install -y ${packages[@]}
```
MichaelSimons, line 249: Why is this necessary, isn't `installPackageWithTdnf` doing this already?

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
```diff
@@ -0,0 +1,256 @@
...
+    # If the package is not already in the list, add it
+    if [[ -z ${upgradablePackageVersions[$pkgName]} ]]; then
+        local versionData=$(echo "$currentVersion,$upgradeVersion")
+        upgradablePackageVersions[$pkgName]=$versionData
+
+        printPackageInfo "$pkgName" "$currentVersion" "$upgradeVersion"
+    fi
+}
+
+installPackageWithApt() {
```
MichaelSimons, line 25: The name is misleading as it doesn't always install the package.

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
```diff
@@ -0,0 +1,256 @@
...
+        local pkgVersion=${BASH_REMATCH[2]}
+
+        apt show 2>/dev/null $1
+        
+        # If this package version exists for install
+        if [[ $? == 0 ]]; then
+            echo "Installing package $1"
+            apt install -y $1
+        else
+            echo "Finding latest version of package $pkgName"
```
MichaelSimons, line 37: Why can't this code path get used all of the time?  Why does the package need to actually get installed when you can determine this information without installing the package?

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
```diff
@@ -0,0 +1,256 @@
...
+    mapfile -t aptPackages < <(apt upgrade -oDir::Etc::Sourcelist=/tmp/security.list -s 2>/dev/null | grep Inst | sort)
+
+    # Regex to parse the output from apt to get the package name, current version, and upgrade version
+    local regex="Inst\s(\S+)\s\[(\S+)]\s\((\S+)\s"
+
+    echo
+    echo "Installed packages available to upgrade:"
+    for pkg in "${aptPackages[@]}"
+    do
+        if [[ $pkg =~ $regex ]]; then
```
MichaelSimons, line 67: For these regex matches, it seems like you should always find a match.  For diagnostics, it feels like and else case would be helpful as a type of assert.

eng/common/package-scripts/get-installed-packages.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env sh
+
+# This is the main script for retrieving the list of installed packages.
+
+imageTag=$1
+dockerfilePath=$2
```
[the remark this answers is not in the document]
mthalman: It's intended as a means to make the script self-documenting. There's a contract here that the script needs to conform to. While it's not actually used, I thought it best to indicate that the argument is being passed as part of the contract.

eng/common/package-scripts/get-upgradable-packages.sh
mthalman: In this instance it is necessary because the container needs access to both the `get-upgradable-packages.container.sh` and `get-upgradable-packages.container.bash.sh` files.

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
mthalman: This is a good question. When I originally started implementing these scripts I went with the approach of only installing the package with the specific version and letting the package manager decide for me whether an upgrade was available. But when testing that, I realized that there will be cases where the specific version of the package to be installed doesn't even exist in the repository any more. That's when the `else` condition was added. But yeah, you're right. It would make sense to always follow that code path and avoid installing altogether.

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
mthalman: Removed

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
mthalman: Renamed to checkForUpgradableVersion

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
mthalman: Agreed. I've added error checking throughout.

eng/common/package-scripts/get-installed-packages.sh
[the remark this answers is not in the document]
MichaelSimons: Makes sense.

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
```diff
@@ -22,32 +27,32 @@ addUpgradeablePackageVersion() {
     fi
 }
 
-installPackageWithApt() {
+checkForUpgradableVersionWithApt() {
     if [[ $1 =~ $packageVersionRegex ]]; then
```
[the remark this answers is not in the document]
mthalman: Fixed

mthalman:
Also removed unnecessary functions now that the "install package"/"check for upgrade" pattern is no longer being used.

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
```diff
@@ -0,0 +1,194 @@
...
+args=( $@ )
+
+packages=( "${args[@]:1}" )
+
+declare -A upgradablePackageVersions
+upgradablePackageVersions=()
+
+packageVersionRegex="(\S+)=(\S+)"
+
+if type apt > /dev/null 2>/dev/null; then
```
MichaelSimons, line 157: Have you considered refactoring to have this pattern implemented once?  It's not obvious if it will help or hurt but it seems like it would reduce the amount of duplication.

```
echo "Updating package cache..."
updatePackageCache  # method which switches on package manager type to invoke updatePackageCache<PackageType>  

for pkgName in "${packages[@]}"
do
    if [[ ! pkgName  =~ $packageVersionRegex ]]; then
        writeError "Package version info for '$1' must be in the form of <pkg-name>=<pkg-version>"
    fi

    local pkgName=${BASH_REMATCH[1]}
    local pkgVersion=${BASH_REMATCH[2]}

    checkForUpgradableVersion $pkgName $pkgVersion  # method which switches on package manager type to invoke checkForUpgradableVersion<PackageType> 
done
outputPackagesToUpgrade
```

Building on that thought, you could do something like

```
if type apt > /dev/null 2>/dev/null; then
    $pkgType="Apt"
fi
...# logic for each supported package manager

echo "Updating package cache..."
eval updatePackageCache$pkgType

for pkgName in "${packages[@]}"
do
    if [[ ! pkgName  =~ $packageVersionRegex ]]; then
        writeError "Package version info for '$1' must be in the form of <pkg-name>=<pkg-version>"
    fi

    local pkgName=${BASH_REMATCH[1]}
    local pkgVersion=${BASH_REMATCH[2]}

    eval checkForUpgradableVersion$pkgType $pkgName $pkgVersion
done
outputPackagesToUpgrade
```

MichaelSimons (commented):
Looks good.  I had one comment about possibly reducing duplication.

eng/common/package-scripts/get-upgradable-packages.container.bash.sh
```diff
@@ -0,0 +1,190 @@
...
+if [[ $(type -t checkForUpgradableVersionWith$pkgType) != "function" ]]; then
+    writeError "Missing function named 'checkForUpgradableVersionWith$pkgType'"
+fi
+
+echo "Updating package cache..."
+updatePackageCacheWith$pkgType
+
+for pkgName in "${packages[@]}"
+do
+    checkForUpgradableVersionWith$pkgType $pkgName
```
MichaelSimons, line 188: Any reason to not apply the regex here and pass in the separated pkgName and pkgReg to reduce duplication?

mthalman: Sorry, I missed that from your earlier comment.

MichaelSimons (approved):
:shipit:

PR #838, opened 2021-08-10
Add command to post publish notifications

mthalman:
Defines a command that can be used to post notifications during the publish job.  The intent is to provide info at a glance that summarizes the results and which tags were published.

If the publish build had been queued by AutoBuilder, an attempt is made to locate the GitHub issue that was generated for the queue notification and to link that issue into the generated notification for the publish (labeled as "Queued by").  In order to find the correlated issue, there needs to be a way to find the ID of the queued build that the issue is associated with.  This is done by embedding the issue with metadata inside of an HTML comment.  This allows for flexible, programmatic access to metadata associated with an issue.

The task results section summarizes the results of a limited set of publish tasks, not all of the tasks within the publish job because it contains many tasks that are not too interesting.  To avoid Image Builder knowing the names of these tasks, the desired set of task names are passed in via command options.

Here's a screenshot of a fictitious example of the issue that would be generated:

image


src/Microsoft.DotNet.ImageBuilder/src/Services/VssConnectionFactory.cs
```diff
@@ -82,9 +82,15 @@ public void Dispose()
                 public Task<List<string>> AddBuildTagAsync(Guid project, int buildId, string tag) =>
                     _inner.AddBuildTagAsync(project, buildId, tag);
 
+				public Task<WebApi.Build> GetBuildAsync(Guid projectId, int buildId) =>
```
MichaelSimons, line 85: Odd indentation.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PostPublishNotificationCommand.cs
```diff
@@ -0,0 +1,295 @@
...
+            StringBuilder notificationMarkdown = new();
+            string buildUrl = string.Empty;
+            Dictionary<string, TaskResult?> taskResults = Options.TaskNames
+                .ToDictionary(name => name, name => (TaskResult?)null);
+            Dictionary<string, string> buildParameters = new();
+            BuildResult overallResult = BuildResult.Succeeded;
+            BuildReason buildReason = BuildReason.None;
+            string? correlatedQueueNotificationUrl = null;
+
+            if (!Options.IsDryRun)
```
MichaelSimons, line 52: Why not run this in the dry run?  I was expecting the command would gather all the info just not publish when dry running.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PostPublishNotificationCommand.cs
```diff
@@ -0,0 +1,295 @@
...
+                {
+                    throw new InvalidOperationException(
+                        $"Build task with name '{task}' could not be found in the build timeline.");
+                }
+
+                taskResults[task] = record.Result;
+
+                if (record.Result is not null)
+                {
+                    switch (record.Result.Value)
```
MichaelSimons, line 155: Curious why success isn't explicitly handled?  Is there danger in unexpected values resulting in success being reported because the overallResult is defaulted to success?

src/Microsoft.DotNet.ImageBuilder/src/Commands/PostPublishNotificationCommand.cs
```diff
@@ -0,0 +1,295 @@
...
+                    return issue.HtmlUrl;
+                }
+            }
+
+            return null;
+        }
+
+        private async Task<BuildResult> GetBuildTaskResultsAsync(Dictionary<string, TaskResult?> taskResults, IBuildHttpClient buildClient, TeamProject project)
+        {
+            BuildResult overallResult = BuildResult.Succeeded;
```
MichaelSimons, line 140: Is there a better way to retrieve the overall result versus calculating it? Does the logic below appropriate handle scenarios in which a task fails but a later one is set to always run and passes?

src/Microsoft.DotNet.ImageBuilder/src/Commands/PostPublishNotificationCommand.cs
mthalman: Gathering all the info requires connecting to online services which require authentication. This would all fail in a PR build if we didn't use the dry-run option.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PostPublishNotificationCommand.cs
mthalman: Not really, because this is getting executed within the build that it's reporting on. So the build hasn't finished and doesn't have a result, technically. The intent is to execute this command as one of the last steps in the publish job such that all the stuff being reported on has already been executed. Because the build is still in progress, we need to manually compute the result of what's currently been executed.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PostPublishNotificationCommand.cs
mthalman: Fixed by defaulting the result to None instead of Succeeded.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PostPublishNotificationCommand.cs
MichaelSimons: What you said makes sense as far as why it needs to be calculated.  Did my second question make sense and is that scenario handled?  I realize it may not be an issue with the specific tasks passed in but something to call out if it isn't handled.

MichaelSimons: Nevermind, I see you addressed this below.

MichaelSimons (approved):

src/Microsoft.DotNet.ImageBuilder/src/Commands/PostPublishNotificationCommand.cs
mthalman: > Does the logic below appropriate handle scenarios in which a task fails but a later one is set to always run and passes?

Yes, when it processes the task which failed, it'll set `overallResult` to `Failed`. If it then encounters a later task which succeeds, it won't overwrite the `overallResult` because of the conditional checks in the `Succeeded` and `SucceededWithIssues` cases.

PR #861, opened 2021-09-08
Track syndicated manifest digests

### LINKED ISSUE #847: Waiting for MCR ingestion can fail on syndication scenarios
An issue was identified during the publishing of Windows Server 2022 .NET Core images which caused a failure when checking for MCR ingestion.  The failure occurred because of a timeout checking for multi-arch tags for 3.1, 3.1.18, 3.1.412, and latest.  All of these tags were expected to be ingested by MCR but in fact none of them were.  The fact that they weren't ingested is actually the correct behavior.  It was the assumption that they would be ingested that was incorrect.  

This all stems from the configuration of the manifest file to syndicate tags to the old "core" repositories.  For WS 2022, the new tags that were added were not configured to be syndicated to the "core" repositories since they were new tags.  The only changed images for the WS 2022 release were the ones for WS 2022.  This meant that when the manifest tool regenerated the multi-arch tag for the 3.1-based tags in the "core" repositories, it ended up with the same digest value that is already published.  So when they were pushed, it was a no-op and MCR did not ingest them.  But Image Builder assumed they would because it built new images within the 3.1 multi-arch tag.  They key is that the new images were _not_ part of the old "core" repositories.  So the logic needs to be smarter to account for syndication to handle this properly.

This is an edge case to account for.  It wouldn't happen normally because on servicing releases, all of the Windows versions would be updated.  In such releases, the multi-arch tags mentioned above _would_ be ingested even though WS 2022 is not part of them because of the other Windows versions being updated.

mthalman:
As [described in the issue](https://github.com/dotnet/docker-tools/issues/847#issuecomment-912552874), multi-arch tags which are syndicated are not being tracked properly when waiting for MCR ingestion. There is no guarantee that the digest of a multi-arch tag will be the same digest in the syndicated destination because the underlying manifest list can be based on different images.

These changes fix this by explicitly tracking the digest of syndicated multi-arch tags in the image info file.  The logic for waiting for MCR ingestion consumes that data to ensure that both the primary and syndicated digests are successfully ingested.

Fixes #847

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/WaitForMcrImageIngestionCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/WaitForMcrImageIngestionCommand.cs
@@ -65,13 +65,6 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 string digestSha = DockerHelper.GetDigestSha(image.Manifest.Digest);
                 yield return new DigestInfo(digestSha, repo.Repo, image.Manifest.SharedTags);
 
-                foreach (string syndicatedDigest in image.Manifest.SyndicatedDigests)
-                {
-                    string syndicatedDigestSha = DockerHelper.GetDigestSha(syndicatedDigest);
-
-
-                }
-
                 // Find all syndicated shared tags grouped by their syndicated repo
                 IEnumerable<IGrouping<string, TagInfo>> syndicatedTagGroups = image.ManifestImage.SharedTags
                     .Where(tag => image.Manifest.SharedTags.Contains(tag.Name) && tag.SyndicatedRepo != null)
@@ -221,7 +214,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                         throw task.Exception;
                     }
 
-                    throw new NotImplementedException();
+                    throw new NotSupportedException();
                 }));
         }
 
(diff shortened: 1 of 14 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/WaitForMcrImageIngestionCommand.cs
```diff
@@ -64,6 +65,13 @@ private IEnumerable<DigestInfo> GetImageDigestInfos(ImageData image, RepoData re
                 string digestSha = DockerHelper.GetDigestSha(image.Manifest.Digest);
                 yield return new DigestInfo(digestSha, repo.Repo, image.Manifest.SharedTags);
 
+                foreach (string syndicatedDigest in image.Manifest.SyndicatedDigests)
```
MichaelSimons, line 68: I don't understand why this for loop is needed.

src/Microsoft.DotNet.ImageBuilder/src/Commands/WaitForMcrImageIngestionCommand.cs
```diff
@@ -202,12 +216,12 @@ private async Task<ImageResultInfo> ReportImageStatusWithContinuationAsync(IMcrS
...
                         }
                     }
-                    else if (task.IsFaulted)
+                    else if (task.Exception is not null)
                     {
                         throw task.Exception;
                     }
 
-                    return null;
+                    throw new NotImplementedException();
```
MichaelSimons, line 224: Would NotSupportedException be a better choice here?

src/Microsoft.DotNet.ImageBuilder/src/Commands/WaitForMcrImageIngestionCommand.cs
mthalman: This was not needed.

MichaelSimons (approved):

MichaelSimons (approved):

PR #863, opened 2021-09-09
Command to pull images

mthalman:
In order to execute a component governance scan over the supported images, we first need to pull those images to the build agent. The CG tool requires that the images to be scanned exist locally on the machine.

These changes add a command to pull the images which can be executed by the pipeline.

Related to dotnet/dotnet-docker#3001

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PullImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PullImagesCommand.cs
@@ -1,0 +1,60 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/PullImagesCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+                {
+                    pulledImages.Append(',');
+                }
+
+                pulledImages.Append(imageTag);
+            }
+
+            if (Options.OutputVariableName is not null)
+            {
+                _loggerService.WriteMessage(PipelineHelper.FormatOutputVariable(Options.OutputVariableName, pulledImages.ToString()));
```
MichaelSimons, line 53: It would simplify the code some as well as eliminate state if the output variable was not specified by utilizing string.join.

_loggerService.WriteMessage(PipelineHelper.FormatOutputVariable(Options.OutputVariableName, string.join ",", pulledImages)));

MichaelSimons (approved):

Issue #830, opened 2021-08-09
"Test Images" in PreBuildValidation assumes `testScriptPath` variable is defined


dagood:
https://github.com/dotnet/docker-tools/blob/7f59365dd99faff417457f5042a24db373553e4f/eng/common/templates/steps/test-images-linux-client.yml#L44-L51

If my repo doesn't have tests yet, I'm not sure what to do with this variable. To get things moving, I added:
```
condition: and(succeeded(), ne(variables['testScriptPath'], ''))
```

mthalman:
If there are no tests, then the test stage can be skipped.  This would be done by not including `test` in the `stages` variable.  So `stages` would just be set to `build;publish`.

mthalman:
Oh, but you're running into this in PreBuildValidation I see.  Ok, yeah.  I see the problem.

dagood:
> If there are no tests, then the test stage can be skipped. This would be done by not including `test` in the `stages` variable. So `stages` would just be set to `build;publish`.

Just to confirm: yep, already have it set to `build;publish`. (So I suppose conditioning this step on that variable instead of `ne(variables['testScriptPath'], '')` would work too.)

PR #866, opened 2021-09-14
Skip tests if test script path not set

mthalman:
In order to support other teams consuming the common infra, this removes the assumption that a test script will exist for the source repo.  If the `testScriptPath` variable is not set, it will skip execution of the PreBuildValidation job and any test steps.

Related to #830


dagood (approved):
Thanks!

MichaelSimons (approved):
It's unfortunate the ripple effect this change had on how many of the step templates now have to have conditions.  @dleeapho - this is an example of something I expect better support for from a CI system.

PR #873, opened 2021-09-17
Support for custom build matrix OS versions

mthalman:
Distroless Dockerfiles of CBL-Mariner will need to be executed on a custom agent pool. Currently, there is only one matrix defined for Linux Dockerfiles of a given architecture.  They're all lumped together targeting the same agent pool.  This is problematic because there's not a mechanism to separate out the distroless Dockerfiles to run them in a separate agent pool.  There needs to be a way for the generated build matrices to provide a distinct matrix just for the distroless Dockerfiles.  Once a separate matrix is provided, the pipeline can then explicitly reference that matrix and configure the appropriate agent pool name for that matrix.  

One way to do this is to change the matrix generation to not lump all Linux Dockerfiles together but rather define distinct matrices for each Linux OS version (e.g. mariner distroless, alpine3.14, buster, focal, etc).  That allows for maximum flexibility in that we can customize which agent pool each of those OS versions runs on. But that requires constant maintenance as supported OS versions change and also produces a lot of duplication in the pipeline YAML file.

Instead, the solution I've provided here uses an opt-in model to customize each platform so that it can define a custom OS version name for the purposes of matrix generation. In the case of distroless for Mariner, this means it can be configured to define a custom OS version (e.g. "mariner-distroless") instead of the default "Linux". This will produce a separate matrix for all platforms that use that same custom OS version. The pipeline YAML can then be configured to reference that matrix name.

Example showing both Mariner Distroless and Linux in the pipeline YAML:

```yaml
  - template: ../jobs/build-images.yml
    parameters:
      name: MarinerDistroless_amd64
      pool: ${{ parameters.marinerDistrolessAmd64BuildPool }}
      matrix: dependencies.GenerateBuildMatrix.outputs['matrix.MarinerDistrolessAmd64']
      dockerClientOS: linux
      buildJobTimeout: ${{ parameters.marinerDistrolessAmdBuildJobTimeout }}
      customInitSteps: ${{ parameters.customBuildInitSteps }}
      noCache: ${{ parameters.noCache }}
  - template: ../jobs/build-images.yml
    parameters:
      name: Linux_amd64
      pool: ${{ parameters.linuxAmd64BuildPool }}
      matrix: dependencies.GenerateBuildMatrix.outputs['matrix.LinuxAmd64']
      dockerClientOS: linux
      buildJobTimeout: ${{ parameters.linuxAmdBuildJobTimeout }}
      customInitSteps: ${{ parameters.customBuildInitSteps }}
      noCache: ${{ parameters.noCache }}
```

Related to #870

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest/Platform.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest/Platform.cs
@@ -67,12 +67,6 @@ namespace Microsoft.DotNet.ImageBuilder.Models.Manifest
         public CustomBuildLegGroup[] CustomBuildLegGroups { get; set; } = Array.Empty<CustomBuildLegGroup>();
 
         [Description(
-            "A custom OS version to use for the matrix instead of the default. This allows for certain platforms " +
-            "to be grouped into a distinct build matrix for when those platforms require custom configuration."
-            )]
-        public string? CustomMatrixOsVersion { get; set; }
-
-        [Description(
             "A label which further distinguishes the architecture when it " +
             "contains variants. For example, the ARM architecture has variants " +
             "named v6, v7, etc."
--- a/src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest/Platform.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest/Platform.cs
@@ -8,6 +8,7 @@ using System.ComponentModel;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
 
+#nullable enable
 namespace Microsoft.DotNet.ImageBuilder.Models.Manifest
 {
     [Description(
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Models/Manifest/Platform.cs
```diff
@@ -52,28 +53,35 @@ public class Platform
...
         [Description(
             "The custom build leg groups associated with the platform."
             )]
         public CustomBuildLegGroup[] CustomBuildLegGroups { get; set; } = Array.Empty<CustomBuildLegGroup>();
 
+        [Description(
+            "A custom OS version to use for the matrix instead of the default. This allows for certain platforms " +
+            "to be grouped into a distinct build matrix for when those platforms require custom configuration."
+            )]
+        public string? CustomMatrixOsVersion { get; set; }
```
MichaelSimons, line 73: This feels like somewhat redundant information (e.g stating this is Marinier) and something that doesn't belong in the manifest.  Is distroless mariner support going to be something that will be supported at the shared template level - e.g. https://github.com/dotnet/dotnet-docker/tree/main/eng/common/templates/stages?  If so, have you considered adding an option to the generate matrix command that is a list of OsVersions to group in a distinct matrix?  The benefit being that this requirement of building mariner on a special config would be encapsulated within the build templates and kept out of the manifest.

mthalman: Thanks for the suggestion.  I like that approach.

PR #878, opened 2021-09-22
Fix variable resolution logic

mthalman:
This fixes an issue introduced by https://github.com/dotnet/docker-tools/pull/873 where variables that are exclusively defined via the options are not exposed by VariableHelper. This specifically causes an issue when generating readmes in the dotnet/dotnet-docker repo. In that repo, a `branch` variable is provided:
https://github.com/dotnet/dotnet-docker/blob/7275e84505032552923953dce76fd478b42f63f9/eng/readme-templates/Get-GeneratedReadmes.ps1#L34

This variable does not get included by VariableHelper and subsequently not used by the GenerateReadmesCommand class. This causes the readme templates which rely on the branch variable to produce incorrect output.

I've fixed the logic to ensure that any variable exclusively defined via the options is included in the results.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
@@ -54,9 +54,18 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
                     ResolvedVariables.Add(kvp.Key, variableValue);
                 }
             }
-            else if (Options.Variables is not null)
+
+            // Include any variables exclusively defined in the options
+            if (Options.Variables is not null)
             {
-                ResolvedVariables = new Dictionary<string, string?>(Options.Variables);
+                foreach (KeyValuePair<string, string?> kvp in Options.Variables)
+                {
+                    if (!ResolvedVariables.ContainsKey(kvp.Key))
+                    {
+                        string? value = SubstituteValues(kvp.Value);
+                        ResolvedVariables.Add(kvp.Key, value);
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
```diff
@@ -54,9 +54,18 @@ public VariableHelper(Manifest manifest, IManifestOptionsInfo options, Func<stri
                     ResolvedVariables.Add(kvp.Key, variableValue);
                 }
             }
-            else if (Options.Variables is not null)
+
+            // Include any variables exclusively defined in the options
+            if (Options.Variables is not null)
             {
-                ResolvedVariables = new Dictionary<string, string?>(Options.Variables);
+                foreach (KeyValuePair<string, string?> kvp in Options.Variables)
```
MichaelSimons, line 61: If you iterated over the Options.Variables prior to the Manifest.Variables, it feels like the overall code would be a little simpler because you wouldn't have to check if the variable already existed.  In the Manifest.Variables loop, you would just skip any variables already contained in the ResolvedVariables - no else case needed.

MichaelSimons (approved):
One minor suggestion.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
mthalman: That would prevent the ability for a variable provided via options to reference a variable defined in the manifest.  The test I added checks for that scenario.  So in either case, you still need a check to determine if its exclusive to the options.

PR #881, opened 2021-09-27
Support for "FROM scratch" Dockerfiles

mthalman:
This updates the semantics of `PlatformInfo.FinalStageFromImage` to be nullable. When the value is null, it means the final stage of the Dockerfile has `FROM scratch`. Using the semantics of nullability provides better language support to identify all dependencies that would be affected by having a "FROM scratch" Dockerfile.

Related to #871


MichaelSimons (approved):
Nice change.

PR #895, opened 2021-11-08
Update .NET SDK version to Current/6.0

mthalman:
This also applies a [workaround to NuGet restore errors](https://github.com/NuGet/NuGet.Client/pull/4259) that have been showing up in builds.


MichaelSimons (approved):

mthalman:
I updated the projects in the repo to use .NET 6 as well. In .NET 6 SDK, there's now a warning if you don't specify `--self-contained` or `--no-self-contained` when you provide a RID to `dotnet publish`. That warning made me re-evaluate what we're doing in the Dockerfile for publish. I don't see a benefit to using a self-contained publish for this scenario. The downside to using it is that it just makes the publish execute slower.

MichaelSimons (approved):

PR #897, opened 2021-11-11
Track components of built images

mthalman:
These changes allow the components installed within built images to be tracked within the image info file. The word "component" used here is intended to be general purpose name to refer to any versioned thing that we want to track in the image. For now, only Linux packages will be tracked.

This consumes the scripts added by https://github.com/dotnet/docker-tools/pull/824 in order to get the list of packages and versions that are in the image. By default, it uses the script defined in that PR. But an option is available for a custom script to be used for certain Dockerfiles. This accounts for the distroless scenario where it's not possible to execute the commands within the container that are done for regular containers. Each distroless Dockerfile can be associated with its own custom script that can be used to retrieve the packages that are installed.

The downside to storing these components in the image info file, particularly as an array, is the amount of vertical space the text file has for each entry. There are a lot of components installed for some images so the list is lengthy.

Related to https://github.com/dotnet/dotnet-docker/issues/1455

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -183,8 +183,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                     .Split(Environment.NewLine)
                     .Select(val =>
                     {
-                        string[] pkgVersion = val.Split(new char[] { ',', '=' });
-                        return new Component(type: pkgVersion[0], name: pkgVersion[1], version: pkgVersion[2]);
+                        string[] pkgInfo = val.Split(new char[] { ',', '=' });
+                        return new Component(type: pkgInfo[0], name: pkgInfo[1], version: pkgInfo[2]);
                     })
                     .ToList();
             }
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -28,10 +28,11 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         private readonly IGitService _gitService;
         private readonly IProcessService _processService;
         private readonly ImageDigestCache _imageDigestCache;
-        private readonly List<TagInfo> _builtTags = new List<TagInfo>();
+        private readonly List<TagInfo> _processedTags = new List<TagInfo>();
+        private readonly HashSet<PlatformInfo> _builtPlatforms = new();
 
         // Metadata about Dockerfiles whose images have been retrieved from the cache
-        private readonly Dictionary<string, BuildCacheInfo> _cachedDockerfilePaths = new Dictionary<string, BuildCacheInfo>();
+        private readonly Dictionary<string, PlatformData> _cachedPlatforms = new Dictionary<string, PlatformData>();
 
         private ImageArtifactDetails? _imageArtifactDetails;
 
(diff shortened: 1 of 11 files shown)


MichaelSimons:
@mthalman - Can you provide a link a sample of the resulting image-info?

mthalman:
> @mthalman- Can you provide a link a sample of the resulting image-info?

[image-info.zip](https://github.com/dotnet/docker-tools/files/7548888/image-info.zip)

The baseline version of the image info file is about 6K lines. These changes push that up to 90K lines. I'm honestly not sure it really matters. From a usability perspective, the file being larger doesn't make much of a difference. Typical use cases of human interaction with the image info file is looking at diffs (which most of the time won't show any package differences and when there is, you'd want to see them) and searching for text.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -137,12 +145,51 @@ private void PublishImageInfo()
...
+            string args = $"{getInstalledPackagesScriptPath} {imageTag} {platform.PlatformInfo.DockerfilePath}";
+            string? output = _processService.Execute("/bin/sh", args, Options.IsDryRun);
+            if (output is not null)
+            {
+                platform.Components = output
+                    .Split(Environment.NewLine)
+                    .Select(val =>
+                    {
+                        string[] pkgVersion = val.Split(new char[] { ',', '=' });
+                        return new Component(type: pkgVersion[0], name: pkgVersion[1], version: pkgVersion[2]);
```
MichaelSimons, line 187: Do you envision that we would track multiple pkg types for a given distro?  I am wondering if the type can be stored one level up or if it is necessary to store at all.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -137,12 +145,51 @@ private void PublishImageInfo()
...
+
+            string args = $"{getInstalledPackagesScriptPath} {imageTag} {platform.PlatformInfo.DockerfilePath}";
+            string? output = _processService.Execute("/bin/sh", args, Options.IsDryRun);
+            if (output is not null)
+            {
+                platform.Components = output
+                    .Split(Environment.NewLine)
+                    .Select(val =>
+                    {
+                        string[] pkgVersion = val.Split(new char[] { ',', '=' });
```
MichaelSimons, line 186: I think renaming this to something more abstract helps the readability especially when indexing into the various parts below.
```suggestion
                        string[] pkgInfo = val.Split(new char[] { ',', '=' });
```

src/Microsoft.DotNet.ImageBuilder/src/Models/Image/Component.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+#nullable enable
+namespace Microsoft.DotNet.ImageBuilder.Models.Image
+{
+    public class Component : IComparable<Component>
```
MichaelSimons, line 10: Why name this component vs package?  The types/api/options etc to retrieve this info is named in terms of packages.  Do you see tracking other "components" (e.g. the .NET/PowerShell components installed)?

mthalman: I wanted to future proof the schema of the file to avoid the possibility of having to coordinate upgrades if the schema were to change because it was too specific. So yes, I was thinking of the possibility of tracking other components besides Linux packages. PowerShell was top of mind. For the .NET Framework Dockerfiles, there's even more options like NuGet, MSBuild, etc.

MichaelSimons (approved):
LGTM - There is still one open naming suggestion comment.

PR #900, opened 2021-11-12
Add display name for Ubuntu 21.10

mthalman:
Allows `impish` OS version in manifest to be translated to a display name of `Ubuntu 21.10`.


premun (approved):
Thanks! I wanted to get back to the other PR but I am swamped with some First Responder problems at the moment

PR #907, opened 2021-11-23
Pkg scripts support for custom pkg manager args

mthalman:
In order to support querying packages of distroless containers, custom package scripts need to be implemented rather than using the common scripts defined here. But in order to allow for reusability of the logic in these common scripts, they need the ability to allow the caller to pass custom arguments for the package manager. As an example, for distroless scenarios for CBL-Mariner, the custom args needed to pass to the dnf package manager would be `--releasever=$version --installroot /staging`. This allows the logic to target a specific install root location for querying the packages.

These changes update the scripts to support custom package manager args.

In addition, support for the dnf package manager was added because the CBL-Mariner distroless scripts also require the use of that package manager.

There was also an issue discovered with the `get-upgradable-packages.sh` script that prevent it from running properly within Image Builder due to the nested containers. The original implementation used volume mounting to provide access to the scripts within the container. This doesn't work when running within the Image Builder container because the source location of volume mounts must be from the host, even when running in a nested container. To work around this issue, the logic has been updated to build a derivative image which contains the scripts rather than relying on volume mounting.

### DIFF
--- a/eng/common/package-scripts/get-installed-packages.sh
+++ b/eng/common/package-scripts/get-installed-packages.sh
@@ -2,8 +2,50 @@
 
 # This is the main script for retrieving the list of installed packages.
 
+usage() {
+    echo "Usage: $0 [<OPTIONS>...] IMAGETAG DOCKERFILEPATH"
+    echo
+    echo "Queries a container to determine which packages are installed"
+    echo 
+    echo "Options:"
+    echo "  -h: Print this help message"
+    echo "  -a: Additional package manager arg"
+    echo "Arguments:"
+    echo "  IMAGETAG: The tag of the image to query"
+    echo "  DOCKERFILEPATH: The path to the image's Dockerfile"
+}
+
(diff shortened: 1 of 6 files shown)


eng/common/package-scripts/get-installed-packages.sh
```diff
@@ -2,8 +2,50 @@
...
 # This is the main script for retrieving the list of installed packages.
 
+usage() {
+    echo "Usage: $0 [<OPTIONS>...] IMAGETAG DOCKERFILEPATH"
+    echo
+    echo "Queries a container to determine which packages are installed"
+    echo 
+    echo "Options:"
+    echo "  -h: Print this help message"
+    echo "  -a: Additional package manager arg"
```
MichaelSimons, line 12:
```suggestion
    echo "  -a: Additional package manager args"
```

MichaelSimons (approved):

PR #908, opened 2021-11-23
Fix build script to work on Linux with wildcards

mthalman:
Adds quotes around arg values to allow for wildcards to work in Linux.

Fixes https://github.com/dotnet/dotnet-docker/issues/3288


dagood (approved):
Works on my machine. 👍

PR #918, opened 2021-12-09
Consolidate repos that share a name

mthalman:
The changes in https://github.com/dotnet/docker-tools/pull/916 were incomplete because it didn't consolidate repos from separate manifests that shared the same name.

The https://github.com/dotnet/dotnet-buildtools-prereqs-docker repo will end up having multiple sub-manifest files that all define a repo named `dotnet-buildtools/prereqs`. Simply concatenating the repos together into the model isn't sufficient because there is logic throughout Image Builder that assumes that there is only one repo in the model with a given name.

To fix this, I've updated the manifest loading to consolidate all repos that share the same name. It first validates that the repo definitions don't have conflicting state. It then combines all the images from those repos into one repo model.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using Microsoft.DotNet.ImageBuilder.Models.Manifest;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
```diff
@@ -195,10 +196,53 @@ private static Manifest LoadModel(string path, string manifestDirectory)
...
+
+        private static Repo ConsolidateReposWithSameName(IGrouping<string, Repo> grouping)
+        {
+            // Validate that all repos which share the same name also have the same values for the other settings
+            IEnumerable<PropertyInfo> propertiesToValidate = typeof(Repo).GetProperties()
+                .Where(prop => prop.Name != nameof(Repo.Images));
+            foreach (PropertyInfo property in propertiesToValidate)
+            {
+                List<string> distinctPropertyValues = grouping
+                    .Select(repo => property.GetValue(repo))
```
MichaelSimons, line 219: If the property is optional and is not specified, then I don't think this validation logic should apply.  It would be ideal if properties like the readme, could be declared in one spot from a maintenance perspective.  It should only flag specified properties that differ in value.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs
```diff
@@ -195,10 +196,55 @@ private static Manifest LoadModel(string path, string manifestDirectory)
...
+
+                if (distinctNonEmptyPropertyValues.Count > 1)
+                {
+                    throw new InvalidOperationException(
+                        "The manifest contains multiple repos with the same name that also do not have the same " +
+                        $"value for the '{property.Name}' property. Distinct values: {string.Join(", ", distinctNonEmptyPropertyValues)}");
+                }
+            }
+
+            Repo primaryRepo = grouping.First();
```
MichaelSimons, line 234: It's not guaranteed that the values will be in the first repo for optional values.

MichaelSimons (approved):

PR #922, opened 2021-12-14
Remove support for Windows 2004

mthalman:
Removing infra support for Windows Server, version 2004 because it has reached EOL today.


mthalman:
> Is there any cleanup work necessary on the scaleset side as well?

Yes, I'll follow up with DDFun

PR #940, opened 2022-01-13
Remove dead architecture modification code

### LINKED ISSUE #761: Remove dead code: architecture editing
Code was [added](https://github.com/dotnet/docker-tools/pull/152) to `BuildCommand` a while ago in order to workaround an issue where Docker was generating the incorrect architecture for an image's metadata.  When executed on an AArch64 machine, it would set the image's architecture to `aarch64` instead of `arm` for an Arm32 image.  To workaround this, the code directly edits the metadata of the built image so that it's correct.

However, based on the build logs of recent builds, this code path doesn't seem to ever be executed any more because Docker is setting the correct architecture metadata.

For that reason, we should probably just remove this method since it appears to be dead code.

https://github.com/dotnet/docker-tools/blob/8935e8e868ee722dfd6fac37da8fb85d34a890a5/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs#L479-L554

mthalman:
Fixes #761

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
@@ -89,6 +89,16 @@ namespace Microsoft.DotNet.ImageBuilder
             }
         }
 
+        public (Architecture Arch, string? Variant) GetImageArch(string image, bool isDryRun)
+        {
+            string archAndVariant = DockerHelper.ExecuteCommand(
+                "inspect", "Failed to retrieve image architecture", $"-f \"{{{{ .Architecture }}}}/{{{{ .Variant }}}}\" {image}", isDryRun);
+            string[] parts = archAndVariant.Split('/', StringSplitOptions.RemoveEmptyEntries);
+            Architecture arch = Enum.Parse<Architecture>(parts[0], ignoreCase: true);
+            string? variant = parts.Length > 1 ? parts[1] : null;
+            return (arch, variant);
+        }
+
         public bool LocalImageExists(string tag, bool isDryRun) => DockerHelper.LocalImageExists(tag, isDryRun);
 
         public long GetImageSize(string image, bool isDryRun) => DockerHelper.GetImageSize(image, isDryRun);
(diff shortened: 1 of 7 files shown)


MichaelSimons (commented):
These changes look good but what about [this comment](https://github.com/dotnet/docker-tools/issues/761#issuecomment-994989919)?  Do we need to add logic to set the platform according to what is stored in the manifest in order to support scratch images?

mthalman:
> These changes look good but what about [this comment](https://github.com/dotnet/docker-tools/issues/761#issuecomment-994989919)? Do we need to add logic to set the platform according to what is stored in the manifest in order to support scratch images?

Yes, that's true if we want to support scratch images. Given that there's not currently a need to support scratch images, I don't feel it's necessary to make any changes. I don't think logging an issue for it is even warranted since there's not currently a need. Thoughts?

MichaelSimons:
Ok, I got confused when reading https://github.com/dotnet/docker-tools/issues/761#issuecomment-994989919.  That Raspbian  image references builds correctly on our aarch64 build machines but wouldn't on arm64, correct?

mthalman:
Ok, I've updated the build logic to now pass the `--platform` option.

src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
```diff
@@ -76,7 +77,7 @@ public DockerService(IManifestToolService manifestToolService)
                 .Select(buildArg => $" --build-arg {buildArg.Key}={buildArg.Value}");
             string buildArgsString = string.Join(string.Empty, buildArgList);
 
-            string dockerArgs = $"build {tagArgs} -f {dockerfilePath}{buildArgsString} {buildContextPath}";
+            string dockerArgs = $"build --platform {platform} {tagArgs} -f {dockerfilePath}{buildArgsString} {buildContextPath}";
```
MichaelSimons, line 80: What do you think about adding some validation logic to ensure the platform aligns with the base image? Another option would be to only specify for scratch images but that doesn't seem as attractive as this allows for the potential to build images that don't match the current platform.  My concern is there may be incorrect entries in the manifest - (e.g variant).

mthalman: I like that idea. Updated to validate this.

PR #942, opened 2022-01-13
Allow build jobs to be canceled

mthalman:
Builds jobs do not cancel when the build is canceled. This was a regression introduced by https://github.com/dotnet/docker-tools/pull/866.

In that change, `succeeded()` was removed as a condition for the job because that condition interfered with having the PreBuildValidation job be skippable. If `succeeded()` was kept and the PreBuildValidation job was skipped, it would cause the build jobs to be skipped as well. This is why the `in(dependencies.PreBuildValidation.result, 'Succeeded', 'SucceededWithIssues', 'Skipped')` condition was added. However, by removing `succeeded()`, it removed the ability to stop the job when the build is canceled.

To fix this, an explicit condition for checking whether the build is canceled has been added.

### DIFF
--- a/eng/common/templates/jobs/build-images.yml
+++ b/eng/common/templates/jobs/build-images.yml
@@ -13,7 +13,7 @@ parameters:
 
 jobs:
 - job: ${{ parameters.name }}
-  condition: and(${{ parameters.matrix }}, in(dependencies.PreBuildValidation.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'))
+  condition: and(${{ parameters.matrix }}, not(canceled()), in(dependencies.PreBuildValidation.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'))
   dependsOn:
   - PreBuildValidation
   - CopyBaseImages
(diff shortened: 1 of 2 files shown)


eng/common/templates/jobs/build-images.yml
```diff
@@ -13,7 +13,7 @@ parameters:
 
 jobs:
 - job: ${{ parameters.name }}
-  condition: and(${{ parameters.matrix }}, in(dependencies.PreBuildValidation.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'))
+  condition: and(${{ parameters.matrix }}, not(canceled()), in(dependencies.PreBuildValidation.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'))
```
MichaelSimons, line 16: Should the other places that had the remove check removed in https://github.com/dotnet/docker-tools/pull/866 need/should check `not(canceled())` as well?  e.g. https://github.com/dotnet/docker-tools/pull/866/files#diff-55a3695e8e362367ab8c584042b267c367e7eee2de7e6e43ab3021a4edeef052R10

MichaelSimons (approved):

PR #947, opened 2022-01-19
Fix matrix generation to differentiate versions of separate repos

mthalman:
The changes from https://github.com/dotnet/docker-tools/pull/941 have exposed an issue with generation of the PlatformVersionedOs matrix type.

The scenario involves two repos that don't have a dependency on each other but share the same architecture, product version, and OS version. The current behavior will cause separate matrix legs to be generated, as expected, however the names of the legs are the same because the name is only based on the product version and OS version.

These changes update the leg name generation to also include the first platform's repo name as a differentiator.


MichaelSimons (approved):
It is a bit unfortunate to have to always add the repo for two reasons.

1. Longer names are difficult to view in certain places within AzDo
2. When there is a graph that spans multiple repos, showing the root is deceptive.

PR #970, opened 2022-02-15
Include SBOM generation in build

### LINKED ISSUE #961: Integrate SBOM generation into build pipeline
Since .NET Docker-related repos don't use Arcade, they need to explicitly integrate the [SBOM generation build task](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/ado-sbom-generator) into the pipeline.

mthalman:
This adds a pipeline stage which will generate an SBOM from the built images. It runs after the Post-Build stage, in parallel with the Test stage. It has the following workflow:

1. Consumes the image-info.json artifact produced by the Post-Build stage
2. Trims out any unchanged images that were pulled from the cache and not actually built
3. Pulls the images
4. Generate an SBOM from the images
5. Upload the SBOM as a pipeline artifact

Fixes #961

### DIFF
--- a/eng/common/templates/jobs/generate-sbom.yml
+++ b/eng/common/templates/jobs/generate-sbom.yml
@@ -4,6 +4,14 @@ parameters:
 jobs:
 - job: GenerateSBOM
   pool: ${{ parameters.pool }}
+  strategy:
+    matrix:
+      amd64:
+        arch: amd64
+      arm32:
+        arch: arm
+      arm64:
+        arch: arm64
   variables:
     sbomDirectory: $(Build.ArtifactStagingDirectory)/sbom
   steps:
@@ -18,7 +26,7 @@ jobs:
     displayName: Trim Unchanged Images
   - script: >
       $(runImageBuilderCmd) pullImages
-      --architecture '*'
+      --architecture '$(arch)'
       --manifest 'manifest.json'
       --output-var 'pulledImages'
       --image-info '$(artifactsPath)/image-info.json'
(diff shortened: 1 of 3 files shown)


riarenas (approved):
Changes LGTM with a minor comment about the package name. Do you have an example build where we can check out the produced sbom?

mthalman:
Converting this to draft for right now. Some things are going to need to change due to disk space issues on the build agent. It's not going to be possible to pull down all the images together on the agent.

mthalman:
Here's an example build containing the manifest output: https://dev.azure.com/dnceng/internal/_build/results?buildId=1619061&view=artifacts&pathAsName=false&type=publishedArtifacts

riarenas (approved):

MichaelSimons:
> Here's an example build containing the manifest output: https://dev.azure.com/dnceng/internal/_build/results?buildId=1619061&view=artifacts&pathAsName=false&type=publishedArtifacts

That doesn't contain the information I was expecting.  I don't see a mapping from packages to images.  What am I missing?

mthalman:
> > Here's an example build containing the manifest output: https://dev.azure.com/dnceng/internal/_build/results?buildId=1619061&view=artifacts&pathAsName=false&type=publishedArtifacts
> 
> That doesn't contain the information I was expecting. I don't see a mapping from packages to images. What am I missing?

You're right. There isn't. I've already provided this feedback. It's wrong that they allow multiple image tags to be provided but provide no distinction in the output.

MichaelSimons (approved):

PR #975, opened 2022-02-23
Support build retention in pipeline template

### LINKED ISSUE #973: Official Docker builds need to be retained for SBOM access
Now that we've [included SBOM generation in the build](https://github.com/dotnet/docker-tools/pull/970), the builds need to be configured so that they can be retained and not deleted when they expire. This allows for historical access to the SBOMs that get generated from our builds.

See [Retention rules for release build SBOMs](https://github.com/dotnet/arcade/blob/main/Documentation/SBOMGenerationGuidance.md#retention-rules-for-release-build-sboms) for more info.

mthalman:
This updates the pipeline template to support [build retention](#973) for official builds. By default, build retention is not enabled. Any pipeline that wants to enable it needs to be set with a `retainBuild` variable set to `true`.

The pipeline can run any stage independently. So a given build may end up executing only one of those stages. But we'd want to be able to retain any build, regardless of which stage it is running. In order to do that the build step used to retain the build is executed across all of the stages to ensure that it will be retained. If multiple stages run in a build, the step which retains the build essentially becomes a no-op after the first stage has run the step.

Fixes #973

### DIFF
--- a/eng/common/templates/steps/retain-build.yml
+++ b/eng/common/templates/steps/retain-build.yml
@@ -1,0 +1,9 @@
+steps:
+- powershell: >
+    $(engCommonPath)/Retain-Build.ps1
+    -BuildId $(Build.BuildId)
+    -AzdoOrgUri '$(System.CollectionUri)'
+    -AzdoProject '$(System.TeamProject)'
+    -Token '$(System.AccessToken)'
+  displayName: Enable permanent build retention
+  condition: and(succeeded(), eq(variables.retainBuild, 'true'))
--- a/eng/common/Retain-Build.ps1
+++ b/eng/common/Retain-Build.ps1
@@ -1,0 +1,43 @@
+# Adapted from https://github.com/dotnet/arcade/blob/main/eng/common/retain-build.ps1
+Param(
+    [Parameter(Mandatory = $true)][int] $BuildId,
+    [Parameter(Mandatory = $true)][string] $AzdoOrgUri, 
+    [Parameter(Mandatory = $true)][string] $AzdoProject,
+    [Parameter(Mandatory = $true)][string] $Token
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2.0
+
+function Get-AzDOHeaders(
+    [string] $Token) {
+    $base64AuthInfo = [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes(":${Token}"))
+    $headers = @{"Authorization" = "Basic $base64AuthInfo" }
+    return $headers
+}
+
+function Update-BuildRetention(
+    [string] $AzdoOrgUri,
+    [string] $AzdoProject,
+    [int] $BuildId,
+    [string] $Token) {
+    $headers = Get-AzDOHeaders -Token $Token
+    $requestBody = "{
+        `"keepForever`": `"true`"
+    }"
+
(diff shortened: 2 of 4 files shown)


eng/common/templates/steps/retain-build.yml
```diff
@@ -0,0 +1,9 @@
+steps:
+- powershell: >
+    $(engCommonPath)/Retain-Build.ps1
+    -BuildId $(Build.BuildId)
+    -AzdoOrgUri '$(System.CollectionUri)'
+    -AzdoProject '$(System.TeamProject)'
+    -Token '$(System.AccessToken)'
+  displayName: Enable permanent build retention
+  condition: and(succeeded(), eq(variables.retainBuild, 'true'))
```
MichaelSimons, line 9: Can you explain the larger vision on how retainBuild going to be set?

MichaelSimons (commented):
Do you know what is the e2e UX going to be for requesting an SBOM?  I am wondering how we will be able to map an image digest with the build that produced it.  Should we capture the build id in the version details to make this easier?

eng/common/templates/steps/retain-build.yml
mthalman: It's going to be set in the registered official pipeline as a variable.

This screenshot shows an example of the `retainBuild` variable be defined as a variable for the `dotnet-core.yml` pipeline in the AzDO UI:
image

mthalman:
> Do you know what is the e2e UX going to be for requesting an SBOM? I am wondering how we will be able to map an image digest with the build that produced it. Should we capture the build id in the version details to make this easier?

@riarenas - Do you know how this will work in the future? Will the build be the final resting place of the SBOM or will the SBOM be sent elsewhere? If it's the former, we'll need a way to cross-reference the build output with the build that produced it so that we can look up the SBOM, I assume?

riarenas:
> @riarenas - Do you know how this will work in the future? Will the build be the final resting place of the SBOM or will the SBOM be sent elsewhere? If it's the former, we'll need a way to cross-reference the build output with the build that produced it so that we can look up the SBOM, I assume?

We believe in the future the SBOM tooling itself will upload the SBOMs to a central repository. In practice I think we'll always need to keep the sboms in the builds that produced them, and yes, chaining of SBOMs to be able to reference them from other places is something that will happen at some point, although we haven't been given any specific guidance at the moment.

mthalman:
> > @riarenas - Do you know how this will work in the future? Will the build be the final resting place of the SBOM or will the SBOM be sent elsewhere? If it's the former, we'll need a way to cross-reference the build output with the build that produced it so that we can look up the SBOM, I assume?
> 
> We believe in the future the SBOM tooling itself will upload the SBOMs to a central repository. In practice I think we'll always need to keep the sboms in the builds that produced them, and yes, chaining of SBOMs to be able to reference them from other places is something that will happen at some point, although we haven't been given any specific guidance at the moment.

@MichaelSimons - I've logged https://github.com/dotnet/docker-tools/issues/977 so that we can have a way to cross-reference these things.

eng/common/templates/steps/retain-build.yml
MichaelSimons: Will we need to retain SBOMs for repos such as nightly and buildtools?

MichaelSimons (approved):

eng/common/templates/steps/retain-build.yml
mthalman: nightly = yes
buildtools-prereqs = no

PR #982, opened 2022-03-10
Generate SBOM for each container image

### LINKED ISSUE #979: SBOM generation needs to produce one distinct SBOM for each container image
Currently the SBOM generation stage of the pipeline ends up producing a single SBOM for all images built by the pipeline. But the requirement is that each container image have its own SBOM.

This was originally thought to be a bug in the pipeline task implementation because it accepts multiple images as input but only ever produces one SBOM as output. This is by design however. The reason it accepts multiple images is for scenarios where your pipeline is producing some artifact that has dependencies on multiple container images. The task allows you to combine all those dependencies as input to the task to produce the SBOM for your artifact.

mthalman:
This generates a separate SBOM for each image that gets built in the pipeline. This is now done in the Build stage where the images are locally available. The overhead of SBOM generation is fairly minimal. In the Post-Build stage, it consolidates all of the SBOMs generated from the build jobs into a single pipeline artifact.

Fixes #979


eng/common/templates/jobs/build-images.yml
```diff
@@ -105,14 +106,52 @@ jobs:
...
+  - powershell: |
+      $images = "$(BuildImages.builtImages)"
+      if (-not $images) { return 0 }
+      $taskDir = $(Get-ChildItem -Recurse -Directory -Filter "ManifestGeneratorTask*" -Path '$(Agent.WorkFolder)').FullName
+      $manifestToolDllPath = $(Get-ChildItem -Recurse -File -Filter "Microsoft.ManifestTool.dll" -Path $taskDir).FullName
+      $dotnetDir = $(Get-ChildItem -Recurse -Directory -Filter "dotnet-*" -Path $taskDir).FullName
+
+      # Call the manifest tool for each image to produce seperate SBOMs
+      # Manifest tool docs: https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/custom-sbom-generation-workflows
+      $images -Split ',' | ForEach-Object {
```
MichaelSimons, line 132: Have you considered utilizing `ForEach-Object -Parallel` here?

eng/common/templates/jobs/post-build.yml
```diff
@@ -4,14 +4,39 @@ parameters:
...
   - template: ../steps/init-docker-linux.yml
   - template: ../steps/download-build-artifact.yml
     parameters:
       targetPath: $(Build.ArtifactStagingDirectory)
-  - pwsh: |
+  - powershell: |
+      # Move all image-info artifacts to their own directory
+      New-Item -ItemType Directory -Path $(Build.ArtifactStagingDirectory)$(imageInfosSubDir)
+      Get-ChildItem -Directory -Filter "*-image-info-*" $(Build.ArtifactStagingDirectory) |
+        ForEach-Object {
```
MichaelSimons, line 19: Curious if you could eliminate the ForEach and pipe directly to `Move-Item` specifying `-Verbose` to get the desired traceability.

eng/common/templates/jobs/build-images.yml
mthalman: I didn't know about that. Unfortunately it can't be used with our current agent configuration. The Windows agents have PowerShell 5.1 by default which doesn't have the `-Parallel` option.

MichaelSimons: I was wondering if the agents had a new enough version, shucks.

MichaelSimons (approved):

mthalman:
<sigh> So the build is failing for Arm. It looks like the dotnet executable that comes with the task is x64 and not Arm. I've created a PR for the task to have support for Arm: https://github.com/microsoft/dropvalidator/pull/399

mthalman:
I've temporarily disabled SBOM generating for Arm until we can get https://github.com/microsoft/dropvalidator/pull/399.

MichaelSimons (approved):
LGTM,  Are you going to leave https://github.com/dotnet/docker-tools/issues/979 open or log a new issue to track enabling arm when the infrastructure support it?

mthalman:
> LGTM, Are you going to leave #979 open or log a new issue to track enabling arm when the infrastructure support it?

I'll leave it open

mthalman:
I had to make another change. Because the public PR builds do not end up pushing to a registry, they don't actually have a digest. Since the SBOM generation is based on using the digest value, that ends up being empty and causing conflicts between multiple images. So I've made the change to only do SBOM generation for internal builds which do push to a registry and therefore do have digests.

MichaelSimons (approved):

Issue #905, opened 2021-11-16
On Windows, manifest-tool ACR auth appears to work only on Docker pool, not default pool


dagood:
Here's a build using defaults, that fails:
https://dev.azure.com/dnceng/internal/_build/results?buildId=1470163&view=results

```
-- EXECUTING: manifest-tool inspect golangimages.azurecr.io/build-staging/1470163/oss/go/microsoft/alpha/golang:1.15.15-1-windowsservercore-1809-amd64 --raw
time="2021-11-15T21:30:10Z" level=error msg="Error trying v2 registry: Get \"https://golangimages.azurecr.io/v2/build-staging/1470163/oss/go/microsoft/alpha/golang/manifests/1.15.15-1-windowsservercore-1809-amd64\": unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information."
time="2021-11-15T21:30:10Z" level=fatal msg="Get \"https://golangimages.azurecr.io/v2/build-staging/1470163/oss/go/microsoft/alpha/golang/manifests/1.15.15-1-windowsservercore-1809-amd64\": unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information."
```

Here's one that switches to e.g. `Docker-2016-${{ variables['System.TeamProject'] }}`, and succeeds:
https://dev.azure.com/dnceng/internal/_build/results?buildId=1471447&view=results

```
-- EXECUTING: manifest-tool inspect mcr.microsoft.com/windows/nanoserver:1809 --raw
[{"Size":0,"MediaType":"applic...
```

Here's the commit that did the switch, the only change:
https://dev.azure.com/dnceng/internal/_git/microsoft-go-images/commit/e6a5cfd1013bee06e6b2524420ee2a0b8396129e/

---

It looks like `docker login` works differently. In the **failure** case, there are no warnings attached to `docker login`:

```
-- EXECUTING: docker login -u golangimages --password-stdin golangimages.azurecr.io
Login Succeeded
```

In the **success** case, there's a warning:

```
-- EXECUTING: docker login -u golangimages --password-stdin golangimages.azurecr.io
WARNING! Your password will be stored unencrypted in C:\Users\cloudtest\.docker\config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Login Succeeded
```

---

The version data printed at the top of the step is slightly different:

```diff
--- "a/.\\bad.txt"
+++ "b/.\\good.txt"
[...]

PR #998, opened 2022-04-19
Registry REST client to query manifest info

mthalman:
We've been dealing with several issues related to querying for manifest data using the [manifest-tool](https://github.com/estesp/manifest-tool):
* https://github.com/dotnet/docker-tools/issues/905
* https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/581#issuecomment-1061069818
* https://github.com/estesp/manifest-tool/issues/158

To solve the first two, we need to upgrade to v2 of the manifest-tool but the third issue is a blocker for adopting v2 of the tool.

These issues are mitigated with these changes by removing the dependency on the manifest-tool for the purposes of querying manifest data. We'll still using it for pushing manifests.

This is done by implementing a registry REST client that directly sends HTTPS requests to the container registry REST APIs from Image Builder. This also handles OAuth requests by using the creds that are passed in for each registry to the relevant Image Builder command. Further changes will need to be made to the common pipeline so that these creds are passed in appropriately for all relevant commands.

One other thing that was discovered and fixed in these changes was the layer data stored in the image info file. The layer digests ended up being alphabetically sorted when merged into the main image info file. This is incorrect as layer digests are specifically ordered according to their layer position in the image. This has been updated in the code to avoid sorting this data.


src/Microsoft.DotNet.ImageBuilder/src/ImageName.cs
```diff
@@ -0,0 +1,71 @@
...
+                registry = firstSegment;
+                imageName = imageName[(separatorIndex + 1)..];
+            }
+        }
+
+        string? tag = null;
+        string? digest = null;
+
+        separatorIndex = imageName.IndexOf('@');
+        if (separatorIndex < 0)
```
MichaelSimons, line 40: Personal preference - Negate this condition to provide alignment with the inner condition.  It also keeps the digest logic closer to the IndexOf('@') check.

src/Microsoft.DotNet.ImageBuilder/src/RegistryHttpClient.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            private static string? GetGroupValue(Match match, string groupName)
+            {
+                Group group = match.Groups[groupName];
+                return group.Success ? group.Value : null;
+            }
+
+            private static bool ValidateChallenge(string? challenge)
+            {
+                if (string.IsNullOrEmpty(challenge) || !s_bearerRegex.IsMatch(challenge))
```
MichaelSimons, line 191: Could simplify this into an expression bodied method.

src/Microsoft.DotNet.ImageBuilder/src/ManifestResult.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System.Text.Json.Nodes;
+
+namespace Microsoft.DotNet.ImageBuilder;
+
+#nullable enable
+/// <summary>
+/// Result of a manifest query from <see cref="IManifestService"/>.
+/// </summary>
+public class ManifestResult
```
MichaelSimons, line 12: This name feels a bit too abstract.  I can see it causing confusion.  What about something like `ManifestQueryResult`?

src/Microsoft.DotNet.ImageBuilder/src/ImageName.cs
```diff
@@ -0,0 +1,71 @@
...
+            repo = imageName[..separatorIndex];
+        }
+        else
+        {
+            repo = imageName;
+        }
+
+        if (autoResolveRepoName && registry is null && !repo.Contains('/'))
+        {
+            repo = $"library/{repo}";
```
MichaelSimons, line 65: This behavior is context specific.  I was expecting auto resolution logic for the registry.  Why one but not the other?

MichaelSimons (approved):

MichaelSimons (approved):

PR #1001, opened 2022-05-06
Support for generating multiple repo readmes

mthalman:
We need the ability to generate different sets of readmes depending on the site that will be hosting them. For example, readmes for Docker Hub will be different than the MCR portal.

These changes enable that by extending the manifest model for describing readmes.

This JSON snippet example shows the difference in the model, allowing for multiple readmes to be associated with a repo:
```diff
- "readmeTemplate": "eng/readme-templates/README.md",
- "mcrTagsMetadataTemplate": "eng/mcr-tags-metadata-templates/runtime-deps-tags.yml",
+ "readmes": [
+   {
+     "path": "README.runtime-deps.md",
+     "templatePath": "eng/readme-templates/README.md"
+   },
+   {
+     "path": ".mcr/portal/README.runtime-deps.portal.md",
+     "templatePath": "eng/readme-templates/README.mcr.md"
+   }
+ ],
```

An update was also needed in ReadmeHelper to add the tag listing to the readme. While I was in there, I realized the logic could be greatly simplified. Not sure what I was thinking 2 years ago when I wrote that. 🤷‍♂️ That allowed me to get rid of the Sprache dependency.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
@@ -65,12 +65,10 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
                 ValidateRepo(repo, manifestDirectory);
             }
 
-            ValidateFileReference(manifest.Readme, manifestDirectory);
-            ValidateFileReference(manifest.ReadmeTemplate, manifestDirectory);
-
-            if (manifest.ReadmeTemplate != null && manifest.Readme == null)
+            if (manifest.Readme is not null)
             {
-                throw new ValidationException("The manifest must specify a Readme since a ReadmeTemplate is specified");
+                ValidateFileReference(manifest.Readme.Path, manifestDirectory);
+                ValidateFileReference(manifest.Readme.TemplatePath, manifestDirectory);
             }
         }
 
--- a/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
@@ -58,6 +58,8 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
                 throw new ValidationException($"The manifest must contain at least one repo.");
             }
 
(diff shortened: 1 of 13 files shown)


src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -108,17 +133,24 @@ private static void ValidatePathIsRelative(string path)
...
 
-            if (repo.ReadmeTemplate != null && repo.Readme == null)
-            {
-                throw new ValidationException($"The repo '{repo.Name}' must specify a Readme since a ReadmeTemplate is specified");
-            }
-            if (repo.McrTagsMetadataTemplate != null && repo.ReadmeTemplate == null)
+            foreach (Readme readme in repo.Readmes)
             {
-                throw new ValidationException($"The repo '{repo.Name}' must specify a ReadmeTemplate since a McrTagsMetadataTemplate is specified");
+                string readmeFilename = Path.GetFileName(readme.Path);
```
MichaelSimons, line 140: Why is this called?  I'm not seeing `readmeFilename` referenced anywhere.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
```diff
@@ -108,17 +133,24 @@ private static void ValidatePathIsRelative(string path)
...
+
+
+                ValidateFileReference(readme.Path, manifestDirectory);
+                ValidateFileReference(readme.TemplatePath, manifestDirectory);
+
+                if (readme.TemplatePath != null && readme.Path == null)
```
MichaelSimons, line 146: Isn't the readme.Path required?  How would this condition ever evaluate to true?

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ModelExtensions.cs
mthalman: Fixed. This was leftover code from before I refactored some things.

mthalman:
I realized that this new `Readme` model type should also be used for the root Manifest type and not just the Repo type. I've updated the model to now be consistent.

MichaelSimons (approved):

PR #1022, opened 2022-07-05
Support multiple OS versions in tests

mthalman:
These are changes to support a fix for https://github.com/dotnet/dotnet-docker/issues/3870. The issue there is that we're excluding distroless Mariner images from being tested in jobs where both Mariner and distroless Mariner images were built (e.g. PR builds).

In order to fix this, the intention is to make use of the data we have available in the generated matrix. It includes an `osVersions` field which specifies the set of OS versions represented by the Dockerfile paths it has included to be built/tested in matrix job. In the case of this scenario, that field contains both "cbl-mariner2.0" and "cbl-mariner2.0-distroless", for example. By passing both of these values to the `run-tests.ps1` script, it can correctly include both of those image types to be in its testing.

This is a breaking change with the contract between the common pipeline and the test script because it changes the name and type of the OS parameter.

### DIFF
--- a/eng/common/templates/stages/build-test-publish-repo.yml
+++ b/eng/common/templates/stages/build-test-publish-repo.yml
@@ -54,7 +54,16 @@ stages:
       testJobTimeout: ${{ parameters.linuxAmdTestJobTimeout }}
       preBuildValidation: true
       internalProjectName: ${{ parameters.internalProjectName }}
-      customInitSteps: ${{ parameters.customTestInitSteps }}
+      customInitSteps:
+        - ${{ parameters.customTestInitSteps }}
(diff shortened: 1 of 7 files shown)


dagood (approved):
For now, we don't have any tests running in the Go images repo yet, and we only build distroful Mariner images (because Go's runtime is always self-contained in the app), but good to know.

mthalman:
> For now, we don't have any tests running in the Go images repo yet, and we only build distroful Mariner images (because Go's runtime is always self-contained in the app), but good to know.

"Distroful"? Is that the oppposite of distroless? 😆 Love it.

MichaelSimons (approved):

PR #1028, opened 2022-07-13
Consolidate build jobs in check-base-image-updates pipeline

### LINKED ISSUE #887: Disallow silent failure of CopyBaseImages stages for check-base-image-updates pipeline
In order to prevent issues like https://github.com/dotnet/docker-tools/pull/886 from going unnoticed, the check-base-image-updates pipeline should fail if it's unable to copy base images. The CopyBaseImages stage is shared between this pipeline and product builds so it was originally designed to not fail the build if the copy failed.  But in the case of check-base-image-updates, the whole point of that pipeline relies on having the latest version of base images to compare against. If the copy failed, then we're not necessarily comparing with the latest versions. So the proposed solution is that CopyBaseImages should fail the build if it fails in the check-base-image-updates pipeline but not for product builds.

mthalman:
This refactors the check-base-image-updates pipeline and its underlying copy-base-images template. There was a waste in the number of build jobs being used in the pipeline.

Originally, this command was pulling images down to get their digest values. But that was changed in https://github.com/dotnet/docker-tools/pull/620 to only rely on manifest querying. Because of that, there's no need to have platform-specific build jobs that are running this command. It's essentially platform-agnostic at this point, meaning that we can check the status of Windows images from a Linux build job, for example.

So these changes simplify the pipeline by just consolidating everything into one job. And since I was in the copy-base-images template and I fixed #887 as well so that we'll detect failures of mirroring images when running from this pipeline.

Fixes #887


eng/pipelines/check-base-image-updates.yml
```diff
@@ -14,93 +14,40 @@ variables:
...
-        osType: linux
-        architecture: arm64
-        dockerClientOS: linux
-        staleImagePathsVariableName: linux-arm64-stale-image-paths
-        subscriptionsPath: $(checkBaseImageSubscriptionsPath)
-
-  - job: Get_Stale_Images_Windows_AMD
-    pool: Docker-1809-Internal
-    steps:
-    - template: templates/steps/get-stale-images.yml
```
MichaelSimons, line 70: Is this template being referenced anymore after these changes?

mthalman: Nope, deleted.

MichaelSimons (approved):

PR #1030, opened 2022-07-21
Move NuGet.config to repo root

mthalman:
By moving the NuGet.config file to the root of the repo, it ensures that it gets applied to all projects in the repo, not just the update-dependencies project. Specifically, the tool projects in the eng directory weren't making use of it. In that case, they were falling back to whatever NuGet.config file existed on the system hosting this repo. We want to be explicit for the repo about which package sources are used.


MichaelSimons (approved):

mthalman:
The path change for the NuGet.config file requires some changes for the Dockerfiles that exist. Both the file-pusher and yaml-updater projects had their own local NuGet.config copy. I've deleted them in favor of the file at the repo root. This can be done because the Dockerfiles are configured to use the repo root as the build context.

The same is not true for the ImageBuilder Dockerfile. That Dockerfile is built by the common infrastructure which requires that the build context be the same as the Dockerfile location. I've experimented with file linking such that a NuGet.config link be placed in src\Microsoft.DotNet.ImageBuilder so that we can still maintain just one central copy of a NuGet.config. But Docker requires the file to be a hard link, not a soft link. And git doesn't preserve hard links on merge. So I just decided to keep a copy of the NuGet.config file in src\Microsoft.DotNet.ImageBuilder.

The diff shows that the NuGet.config file I "deleted" from eng\src\file-pusher was actually "moved" to src\Microsoft.DotNet.ImageBuilder, since I added a new copy there.

MichaelSimons (approved):

PR #1031, opened 2022-07-28
Allow base image tags to be overriden

mthalman:
There's a need to be able to pull the base images that our Dockerfiles are dependent on from a registry location other than the Docker Hub registry. But we still want to maintain our Dockerfile content as-is to allow the community to make use of them with the canonical source of these base images.

This can be implemented in a build by pulling from the overridden location, tagging it with the same tag referenced by the Dockerfile, and proceeding with the rest of the build as usual. This allows us to keep the Dockerfile unchanged but still allows us to make use of a base image from a different source.

By ensuring that images from the overridden location are the same images that exist on Docker Hub, we can maintain our commitment to the community that the images we publish are effectively based on the location described by the Dockerfiles.

This is implemented by providing two command options: `base-override-regex` and `base-override-sub`. When used together, these options allow a caller to specify a regex that, when matched on a base image tag, will replace it with the substitution pattern. This provides a dynamic way for callers to customize how the base image tag should be overridden. It's consumed by `CopyBaseImagesCommand`, to allow images from the overridden location to be mirrored to our ACR, and `BuildCommand`, to do the retagging logic described earlier.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BaseImageOverrideOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BaseImageOverrideOptions.cs
@@ -1,0 +1,65 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.CommandLine;
+using System.Linq;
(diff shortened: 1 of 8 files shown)


MichaelSimons (commented):
This turned out nice.  I like the simplicity.

MichaelSimons (approved):

PR #1034, opened 2022-08-03
Fixes to better support tag override

mthalman:
The changes from https://github.com/dotnet/docker-tools/pull/1031 are incomplete when it comes to the scenario of checking for whether an update to the base image exists. When the Dockerfile's base tag is being overridden, we want to compare against the digest of the tag that has the override applied, not the tag that's defined in the Dockerfile.

This requires a change to the `GetStaleBaseImagesCommand` using the same pattern that was used for the other commands in https://github.com/dotnet/docker-tools/pull/1031. It simply applies the override, if applicable, when it needs to look up the digest value that gets compared to the digest from the image info file.

Similarly, `BuildCommand` also needs to be updated so that it outputs the correct base image digest when generating the image info file. Again, it just needs get the digest of the image tag that has the override applied.


MichaelSimons (approved):
LGTM - This showcases the value of the way the code is factored!

PR #1038, opened 2022-08-05
Split base image subscriptions

mthalman:
The changes in https://github.com/dotnet/dotnet-docker/pull/3971 require a corresponding change in the execution of the `getStaleImages` command. That command needs to have the same `--base-override-regex` and `--base-override-sub` options in order to correctly interpret the image digests that would be written to the image info file by those builds (see https://github.com/dotnet/docker-tools/pull/1034).

However, we can't make that change in all cases of when the `getStaleImages` command is called. We want to evaluate the changes from https://github.com/dotnet/dotnet-docker/pull/3971 in the nightly branch before rolling out to main. Since the `getStaleImages` command applies to all subscriptions defined in https://github.com/dotnet/docker-tools/blob/main/eng/check-base-image-subscriptions.json, that would mean the tag override options would get applied to checks in the dotnet-docker/main branch as well. That would cause diffs to be detected in the base image digest because that branch isn't yet configured to use the tag override. And those diffs would cause rebuild attempts; not what we want.

To avoid that, we need to split apart the subscriptions to limit the scope of where the new tag override options are applied. I've refactored the pipeline to have run two jobs in parallel that run the same logic, but parameterized in order to use different command options and to target different subscription files. The subscriptions for the dotnet-docker/main branch have been split into a separate subscription file. This allows the tag override options to not be applied for the secondary subscription file.

This PR needs to be merged in conjunction with https://github.com/dotnet/dotnet-docker/pull/3971 to ensure that image info content is kept in sync.

### DIFF
--- a/eng/check-base-image-subscriptions.json
+++ b/eng/check-base-image-subscriptions.json
@@ -24,25 +24,6 @@
     "manifest": {
       "owner": "dotnet",
       "repo": "dotnet-docker",
-      "branch": "main",
-      "path": "manifest.json"
-    },
-    "imageInfo": {
-      "owner": "dotnet",
-      "repo": "versions",
-      "branch": "main",
-      "path": "build-info/docker/image-info.dotnet-dotnet-docker-main.json"
-    },
-    "pipelineTrigger": {
-      "id": 373,
-      "pathVariable": "imageBuilder.pathArgs"
-    },
-    "osType": "linux"
-  },
-  {
-    "manifest": {
-      "owner": "dotnet",
-      "repo": "dotnet-docker",
       "branch": "nightly",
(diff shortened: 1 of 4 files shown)


MichaelSimons (commented):
Do you see this as something temporary and will get reverted or this is permanent?  Either way I think some comments in the code to explain this or to state it is temporary would be beneficial.

mthalman:
> Do you see this as something temporary and will get reverted or this is permanent? Either way I think some comments in the code to explain this or to state it is temporary would be beneficial.

Hopefully it's temporary but I've added a comment.

MichaelSimons (approved):

PR #1046, opened 2022-08-23
Compensate for varying .NET install location from manifest task

mthalman:
The change to the manifest AzDO task that made the changes in https://github.com/dotnet/docker-tools/pull/1045 necessary have been rolled back. That rollback now means the changes in https://github.com/dotnet/docker-tools/pull/1045 cause build failures because it's attempting to look for the .NET install location in the wrong directory.

I've updated the script to handle either scenario. This allows it to be more robust and at least handle either of these conditions. The team which owns this AzDO task do intend to add roll out there changes again. They were testing it temporarily.


MichaelSimons (approved):
Hopefully they can provide a better UX for our usage scenario because this has been a little ridiculous.

PR #1068, opened 2022-10-18
Define version of Image Builder image with Azure CLI installed

mthalman:
In order to implement https://github.com/dotnet/dotnet-docker/issues/4139, we need the ability to authenticate to the internal ACR. This requires the use of the Azure CLI to be able to authenticate into Azure with a certificate.

Installing the Azure CLI adds 1 GB to the image size, a 270% increase over the original image size of Image Builder. The large install size is already tracked by https://github.com/Azure/azure-cli/issues/7387. Given that the use of the Azure CLI is only needed in scenarios to target internal builds, it doesn't seem like a good idea to have it installed for the Image Builder image used everywhere.

With these changes, I've defined a separate image that adds the Azure CLI to the base Image Builder image. This will allow the pipeline runs that target public builds to not be subjected to the very large image size. It will only be relevant for runs that target internal builds.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/Dockerfile.az.linux
+++ b/src/Microsoft.DotNet.ImageBuilder/Dockerfile.az.linux
@@ -1,0 +1,29 @@
+ARG PYTHON_VERSION=3.10
+
+# Azure CLI installer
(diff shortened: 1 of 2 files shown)


mthalman:
Converted to draft. Need to make some more changes here.

MichaelSimons (approved):

PR #1074, opened 2022-11-01
Fix YAML parsing in Get-ImageBuilder.ps1

mthalman:
The changes in https://github.com/dotnet/docker-tools/pull/1070 broke the use of the `Get-ImageBuilder.ps1` script because it's unable to properly parse the `docker-images.yml` file. The YAML file now includes comments and blank lines that need to be filtered out.


MichaelSimons (approved):

mthalman:
I missed that the `Invoke-CleanupDocker.ps1` script does the same parsing. So I've factored out the parsing into a common script that can be reused.

MichaelSimons (approved):

PR #1089, opened 2022-12-08
Use "docker manifest" command instead of manifest-tool

### LINKED ISSUE #1084: Image Builder waits for MAR ingestion of multi-arch tags that aren't ingested for retries
In the scenario where the Publish pipeline stage needs to be retried due to a failure like a MAR ingestion timeout, the next attempt will end up failing if any multi-arch tags are included in the published tags. It fails in the step which waits for those multi-arch tags to be ingested by MAR. But MAR never received notification that the multi-arch tags were republished (the webhook was never triggered for those tags).

This behavior is due to the new version of the manifest-tool: https://github.com/dotnet/docker-tools/pull/1067. Compared to the old 1.0 behavior, the new 2.0 behavior of the manifest-tool is to skip pushing the multi-arch tag if it's a duplicate. It executes a HTTP HEAD request first to check what's already in the target registry. Because of that, the multi-arch tags don't get pushed in the retry attempt of the Publish stage. And so no webhooks are triggered for MAR to respond to. But the Image Builder logic is expecting MAR to have ingestion activities associated with these tags so it keeps querying waiting for them and eventually times out.

Possible solutions:

1. Change the "wait for MAR ingestion" logic to not use a minimum queue time for multi-arch tags. This is configured at https://github.com/dotnet/docker-tools/blob/fd1e72af5a9b5f00e07e4608a598494767040bfd/eng/common/templates/steps/wait-for-mcr-image-ingestion.yml#L16 Because this minimum queue time is set currently, it causes the originally published multi-arch tags from the 1st publish attempt to get filtered out. Since we know we can't get duplicate queue attempts for the same multi-arch tag with this behavior of the manifest-tool, that's safe to do.
2. Make use of the `docker manifest create/push` command instead of the manifest-tool. The explicit push that would be used does cause the tag to be pushed regardless of whether it exists or not, causing the webhook to fire.

mthalman:
Due to the behavior described in #1084, these changes remove the usage of the manifest-tool and instead make use of the `docker manifest` command. While this is technically an experimental feature, it's been that way for years and is widely used. I think the use of this command also puts us in a better position to be supported.

Making use of the `docker manifest push` command always causes the manifest to be pushed even if it already exists. This is the behavior that we need and fixes #1084.

The usage patterns are different between the two such that it required a bit of code rejiggering to spit out the commands we need. The manifest-tool was capable of creating and pushing multiple manifest list tags in one command. The `docker manifest` command requires an individual command invocation for each manifest list tag to be created and again to be pushed. Since we have a lot of manifest list tags, this results in a lot of command executions. But on average, all of these commands seem to be able to be executed in 1m 30s for all the content that we have in the nightly branch.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
@@ -7,7 +7,6 @@ using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.IO;
 using System.Linq;
-using System.Text;
 using System.Threading.Tasks;
 using Microsoft.DotNet.ImageBuilder.Models.Image;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
@@ -19,6 +18,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     public class PublishManifestCommand : DockerRegistryCommand<PublishManifestOptions, PublishManifestOptionsBuilder>
     {
         private readonly IManifestService _manifestToolService;
+        private readonly IDockerService _dockerService;
         private readonly ILoggerService _loggerService;
         private readonly IDateTimeService _dateTimeService;
         private List<string> _publishedManifestTags = new List<string>();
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
```diff
@@ -44,29 +46,21 @@ public override async Task ExecuteAsync()
...
-                        _manifestToolService.PushFromSpec(manifestFilename, Options.IsDryRun);
-                    }
-                    finally
-                    {
-                        File.Delete(manifestFilename);
-                    }
+                DateTime createdDate = _dateTimeService.UtcNow;
+                Parallel.ForEach(_publishedManifestTags, tag =>
+                {
+                    _dockerService.PushManifestList(tag, Options.IsDryRun);
```
MichaelSimons, line 63: Would it make sense to push immediately after the manifest was created (e.g. generate and push within the same loop)?

MichaelSimons (approved):
Nice to finally get off the manifest tool

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
mthalman: I'd rather do it in separate stages like this. That way if there's an error creating a tag for whatever reason, we don't end up pushing anything.

PR #1091, opened 2022-12-19
Image caching redesign

### LINKED ISSUE #821: Different digest value of cached image encountered on push
Some odd behavior with Docker has been discovered which has the effect of producing a new digest value for a pushed image that had been pulled as a cache image in a build.  That is not the expected behavior.  Pulling an image, tagging it, and pushing it is expected to yield the same digest value in the new registry/repository as the pulled image. This seems to be the same issue that is logged at https://github.com/moby/moby/issues/31919.

This behavior can only manifest in a build which encounters an image cache hit and needs to pull down an already built image. If a new digest was to be created for the pushed image, it would result in the publish failing during the task which waits for MCR ingestion. That's because the digest being queried in MCR's status API is the pulled digest, not the pushed digest. No status will exist for the pulled digest and it will keep querying until the task times out.  Failing at this point prevents the publishing of the image info file and ingestion of Kusto data.

This scenario was encountered in the build for Fedora images in the dotnet/dotnet-buildtools-prereqs-docker repo. In this particular build, a cache hit occurred for the `src/fedora/32/amd64/Dockerfile` file. The digest value stored in the image info for this image was `sha256:e1b2d84583e96d98ea53aa1fde916d5f4b317c82024de5a140713df2831a6239`. When the build later pushed the image, it had a digest value of `sha256:73fe8e96fcc4d0bbc92ca0a6985fe19359dd0daba94c89383dd0a72c399d6b4e`.

The build log is attached: [build-log.txt](https://github.com/dotnet/docker-tools/files/6803832/build-log.txt)

mthalman:
A situation can sometimes occur in image caching scenarios where the cached image is pulled from MAR and pushed to the ACR resulting in a different image digest in ACR than it had in MAR. This causes the incorrect data to exist in the image info which affects caching behavior of subsequent builds. It can also affect the logic for querying MAR's image ingestion status because we'll be querying the expected digest which doesn't exist. The reason we'd be querying for image ingestion status for a cached image is if it was assigned new tags.

This is fixed by changing the workflow for how image caching is done. When a cache hit occurs, this is the current workflow:

1. Pull image from MAR by digest
2. Add desired tags based on ACR staging repo
3. Push tags to ACR

Here is the new workflow:

1. [Import/copy image](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-import-images) from MAR to ACR staging repo for all the desired tags
2. Pull the image from ACR staging repo by digest
3. Add desired tags based on ACR staging repo

This new workflow means that, for cached images, we never pull the image locally and push it to the ACR. We rely on the ACR import logic instead. Hopefully the ACR import logic won't run into this issue of the digest being different. If they do, I would consider it a bug to be logged on ACR. And if it does occur, it will actually be caught much sooner in the pipeline now. Because we pull the image by its expected digest from the staging repo in step 2 above, if it doesn't exist we'll get an error at that point. This is much better than running into the issue during the publish stage when images would have already been published.

Because the image import logic was already implemented by `CopyImagesCommand`, I've factored out that logic into a separate service, `CopyImagesService`, that can be reused by `BuildCommand`. Changes were also necessary to the build options to provide the necessary options (subscription, ACR tenant, etc.) for executing the import operation on the ACR.

Fixes #821

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -12,7 +12,6 @@ using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.DotNet.ImageBuilder.Models.Image;
-using Microsoft.DotNet.ImageBuilder.Models.Manifest;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
 
 #nullable enable
@@ -25,6 +24,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         private readonly ILoggerService _loggerService;
         private readonly IGitService _gitService;
         private readonly IProcessService _processService;
+        private readonly ICopyImageService _copyImageService;
         private readonly ImageDigestCache _imageDigestCache;
         private readonly List<TagInfo> _processedTags = new List<TagInfo>();
         private readonly HashSet<PlatformData> _builtPlatforms = new();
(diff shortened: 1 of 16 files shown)


MichaelSimons:
For completeness the new process still involves step 3 (Push tags to ACR) from the current workflow.

mthalman:
> For completeness the new process still involves step 3 (Push tags to ACR) from the current workflow? Or is that only done locally?

All relevant tags get imported in step 1 of the new workflow. The import isn't on a per-image basis; it's on a per-tag basis. So step 3 of the old workflow is incorporated in step 1 of the new workflow with the difference being that it's handled by the import process and not pushed from the build machine.

MichaelSimons (approved):

axljoker03 (approved):

PR #1117, opened 2023-02-01
Change tagging from parallel to serial

mthalman:
Lately there have been random build timeouts popping up at the point where the `build` command tags cached images. It's not clear what is hanging. It could either be a deadlock of some kind in ImageBuilder's caching code (which seems unlikely since that code hasn't changed) or maybe a new version of Docker that doesn't tolerate parallel tagging operations very well, or maybe something else.

Regardless, this should be fixed by changing the logic to create the tags serially rather than in parallel.


mthalman:
> Curious what the impact is to the build times?

It's very negligible. A given `tag` operation only takes about 1/10 of a second. At worse, we might have 100 tags to create in a given job. That amounts to 10 secs.

MichaelSimons (approved):

PR #1120, opened 2023-02-15
Add CodeQL 3000 Scanning

lbussell:
With this change, the CodeQL tasks capture and upload a snapshot that will be analyzed by CodeQL 3000. Then, the results of the scan will be filed as work items under the `areaPath` on devdiv.visualstudio.com. Here is an [example of the scan](https://1es.lgtm.microsoft.com/logs/a28e14d48b8430bceef606eaade0dc5e5f0aba0d/lang:csharp) [internal Microsoft link]. That scan had the issue-filing turned off before we confirm what area path to use.

I added as much as possible to `eng/common` for re-use in the other .NET Docker repos.


mthalman (approved):

.config/tsaoptions.json
```diff
@@ -0,0 +1,10 @@
+{
+    "instanceUrl": "https://devdiv.visualstudio.com/",
+    "template": "TFSDEVDIV",
+    "projectName": "DEVDIV",
+    "areaPath": "DevDiv\\NET Fundamentals\\.NET Acquisition\\Tools and Infrastructure",
```
MichaelSimons, line 5:
```suggestion
    "areaPath": "DevDiv\\NET Fundamentals\\.NET Acquisition\\Docker",
```

garath (approved):

MichaelSimons (approved):

PR #1132, opened 2023-04-18
Enable BuildKit in ImageBuilder

lbussell:
This is needed to facilitate https://github.com/dotnet/dotnet-docker/issues/4447 and https://github.com/dotnet/dotnet-docker/pull/4542

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/ExecuteHelper.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ExecuteHelper.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Text;
@@ -21,9 +22,20 @@ namespace Microsoft.DotNet.ImageBuilder
             string args,
             bool isDryRun,
             string? errorMessage = null,
-            string? executeMessageOverride = null)
+            string? executeMessageOverride = null,
+            Dictionary<string, string>? envVars = null)
         {
-            return Execute(new ProcessStartInfo(fileName, args), isDryRun, errorMessage, executeMessageOverride);
+            ProcessStartInfo processInfo = new ProcessStartInfo(fileName, args);
+
+            if (envVars != null)
+            {
+                foreach(var envVar in envVars)
+                {
+                    processInfo.EnvironmentVariables.Add(envVar.Key, envVar.Value);
+                }
+            }
+
+            return Execute(processInfo, isDryRun, errorMessage, executeMessageOverride);
         }
 
         public static string? Execute(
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/ExecuteHelper.cs
```diff
@@ -21,9 +22,20 @@ public static class ExecuteHelper
...
-            return Execute(new ProcessStartInfo(fileName, args), isDryRun, errorMessage, executeMessageOverride);
+            ProcessStartInfo processInfo = new ProcessStartInfo(fileName, args);
+
+            if (envVars != null)
+            {
+                foreach(var envVar in envVars)
+                {
+                    processInfo.EnvironmentVariables.Add(envVar.Key, envVar.Value);
+                }
+            }
```
mthalman, line 36: Consider factoring this out into a separate method so that it can be shared with the implementation of `ExecuteWithRetry`.

src/Microsoft.DotNet.ImageBuilder/src/ExecuteHelper.cs
```diff
@@ -89,6 +81,15 @@ public static class ExecuteHelper
             );
         }
 
+        private static ProcessStartInfo AttachEnvironmentVars(ProcessStartInfo info, Dictionary<string, string> envVars)
```
mthalman, line 84: I was thinking the shared method would be a helper that creates a properly configured `ProcessStartInfo` instance. So in the method it would check whether `envVars` is set.

```c#
private static ProcessStartInfo CreateProcessStartInfo(string fileName, string args, Dictionary<string, string>? envVars)
{
    ProcessStartInfo processInfo = new ProcessStartInfo(fileName, args);
    if (envVars != null)
    {
        foreach (var envVar in envVars)
        {
            info.EnvironmentVariables.Add(envVar.Key, envVar.Value);
        }
    }
    return info;
}
```

lbussell: Yeah, I think this makes more sense, thanks.

mthalman (approved):

mthalman:
I'm concerned that this may have affected the metadata of the outputted image. It doesn't look like the architecture variant is getting set for the built image (specific to Arm). This [build result](https://dev.azure.com/dnceng-public/public/_build/results?buildId=265781&view=logs&jobId=d08a0e1a-d569-5d90-0afc-9b3ad3be8bd5&j=d08a0e1a-d569-5d90-0afc-9b3ad3be8bd5&t=ddf57a14-90b8-5bbb-abff-6c99aa588205) is an example of the issue.
```
Unhandled exception: System.InvalidOperationException: Platform 'Dockerfile.az.linux' is configured with an architecture that is not compatible with the base image 'mcr.microsoft.com/dotnet-buildtools/image-builder:linux-arm64':
Manifest platform:
	Architecture: ARM64
	Variant: v8
Base image:
	Architecture: ARM64
	Variant: 
```

lbussell:
Couple findings -

1. I tried building an arm64 image on my Windows machine with Docker Desktop which has buildkit enabled by default. Produced an arm64v8 image from `mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine3.17-arm64v8` and it kept the v8 in the variant. I checked by running `docker inspect -f "{{ .Architecture }}/{{ .Variant }}" $image` like the pipeline.
2. I pulled the offending arm64v8 ImageBuilder from the `dotnetdocker` ACR staging location and it still had the correct variant. I pulled the exact ImageBuilder tag (`mcr.microsoft.com/dotnet-buildtools/image-builder:linux-arm64-2174742`) and verified it still had the correct `v8` variant. Thinking about it more this is expected, because when we use the new ImageBuilder to produce itself, it drops the `v8` variant off of the new image. The existing images have the correct variant or else they wouldn't have made it through CI.

I also retried failed jobs on that pipeline and it still had the same error. This feels like a bug somewhere in ImageBuilder and requires more investigation, so for now I have prepared a PR that reverts the BuildKit changes: https://github.com/dotnet/docker-tools/pull/1136

PR #1133, opened 2023-04-20
Allow passing docker secret mounts through ImageBuilder

lbussell:
I tried to make the new CliHelper Option and the build command arguments more general. I'm working on the tests now.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildOptions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         public ServicePrincipalOptions ServicePrincipal { get; set; } = new();
         public string? Subscription { get; set; }
         public string? ResourceGroup { get; set; }
-        public IEnumerable<(string Id, string Src)>? SecretInfo { get; set; }
+        public IEnumerable<(string Id, string Src)> SecretInfo { get; set; } = Enumerable.Empty<(string Id, string Src)>();
     }
 
     public class BuildOptionsBuilder : DockerRegistryOptionsBuilder
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildOptions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         public ServicePrincipalOptions ServicePrincipal { get; set; } = new();
         public string? Subscription { get; set; }
         public string? ResourceGroup { get; set; }
-        public IEnumerable<(string Id, string Src)> SecretInfo { get; set; } = Enumerable.Empty<(string Id, string Src)>();
+        public IEnumerable<(string Id, string Src)> SecretInfos { get; set; } = Enumerable.Empty<(string Id, string Src)>();
     }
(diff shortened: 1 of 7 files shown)


mthalman:
@lbussell - Can you post an example of what the syntax looks like for making use of this new option? That would be helpful.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildOptions.cs
```diff
@@ -29,6 +29,7 @@ public class BuildOptions : DockerRegistryOptions, IFilterableOptions
         public ServicePrincipalOptions ServicePrincipal { get; set; } = new();
         public string? Subscription { get; set; }
         public string? ResourceGroup { get; set; }
+        public IEnumerable<(string Id, string Src)>? SecretInfo { get; set; }
```
mthalman, line 32: This can then clean up the calling code to not have to check for null.
```suggestion
        public IEnumerable<(string Id, string Src)> SecretInfos { get; set; } = Enumerable.Empty<(string Id, string Src)>();
```

src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
```diff
@@ -78,6 +78,7 @@ public DockerService(IManifestService manifestToolService)
             string platform,
             IEnumerable<string> tags,
             IDictionary<string, string?> buildArgs,
+            IEnumerable<string> additionalArgs,
```
mthalman, line 81: I feel like these should either be combined into one parameter or `IEnumerable<string> additionalArgs` should be changed to `IEnumerable<(string Id, string Src)> secretInfos`. Because it's inconsistent in the responsibilities of the caller. For build args, they're not responsible for constructing `--build-arg` options, but they are for `--secret` options. I prefer the latter approach of using `IEnumerable<(string Id, string Src)> secretInfos`, so that the caller isn't responsible for constructing the `--secret` options.

lbussell:
> @lbussell - Can you post an example of what the syntax looks like for making use of this new option? That would be helpful.

Yes, the syntax should be exactly the same as the docker syntax - 
```
--secret id=<your secret's id>,src=<your literal secret or secret file>
```
https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet-docker?path=/eng/pipelines/stages/build-test-publish-repo.yml&version=GBdev/loganbussell/fips2&line=48&lineEnd=49&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents (internal link)

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildOptions.cs
```diff
@@ -29,7 +29,7 @@ public class BuildOptions : DockerRegistryOptions, IFilterableOptions
         public ServicePrincipalOptions ServicePrincipal { get; set; } = new();
         public string? Subscription { get; set; }
         public string? ResourceGroup { get; set; }
-        public IEnumerable<(string Id, string Src)>? SecretInfo { get; set; }
+        public IEnumerable<(string Id, string Src)> SecretInfo { get; set; } = Enumerable.Empty<(string Id, string Src)>();
```
mthalman, line 32: Considering using the plural form for the name: `SecretInfos`.

mthalman (approved):

mthalman (approved):

lbussell:
This feature is being put on hold and we can re-open this PR if we need it in the future.

PR #1134, opened 2023-05-04
Disable BuildKit on Windows

lbussell:
This should fix the CI failure in https://github.com/dotnet/docker-tools/pull/1129

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
@@ -91,7 +91,7 @@ namespace Microsoft.DotNet.ImageBuilder
 
             Dictionary<string, string> envVars = new();
             // BuildKit is not supported on Windows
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
```diff
@@ -89,10 +89,12 @@ public DockerService(IManifestService manifestToolService)
 
             string dockerArgs = $"build --platform {platform} {tagArgs} -f {dockerfilePath}{buildArgsString} {buildContextPath}";
 
-            Dictionary<string, string> envVars = new Dictionary<string, string>
+            Dictionary<string, string> envVars = new();
+            // BuildKit is not supported on Windows
+            if (platform.Contains("linux"))
```
mthalman, line 94:
```suggestion
            if (DockerHelper.OS == OS.Linux)
```

mthalman (approved):

PR #1142, opened 2023-06-01
Fix platformVersionedOs leg generation for shared cached parent images

### LINKED ISSUE #1141: Duplicate matrix leg name from introduction of aspnet composite Dockerfiles
The introduction of the [aspnet composite Dockerfiles](https://github.com/dotnet/dotnet-docker/pull/4594) have caused there to be situations where the matrix generation fails. It ends up producing duplicate leg names.

Example: https://dev.azure.com/dnceng/internal/_build/results?buildId=2189050&view=results (internal link)

```
Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.InvalidOperationException: Duplicate leg name(s) found in matrix 'linuxArm32': 8.0-alpine3.17-aspnet
   at Microsoft.DotNet.ImageBuilder.Commands.GenerateBuildMatrixCommand.GenerateMatrixInfo() in /image-builder/src/Commands/GenerateBuildMatrixCommand.cs:line 448
   at Microsoft.DotNet.ImageBuilder.Commands.GenerateBuildMatrixCommand.ExecuteAsync() in /image-builder/src/Commands/GenerateBuildMatrixCommand.cs:line 46
   at Microsoft.DotNet.ImageBuilder.Commands.Command`2.<GetCliCommand>b__9_0(TOptions options) in /image-builder/src/Commands/Command.TOptions.cs:line 46
```

Looking at the matrix legs that were generated, I see this:

image

One is targeting the normal aspnet Dockerfiles and the other is for composite but they have the same leg name.

lbussell:
Fixes #1141.

I verified that the new test fails with the old behavior and passes with the new changes. Also verified that the exact scenario from the linked issue works now by using the failing dotnet-docker sha and image info files from CI.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
@@ -128,7 +128,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
                     buildLegsByDockerfilePath.Add(dockerfilePath, leg);
                 }
-                
+
                 matrix.Legs.Add(leg);
 
                 AddImageBuilderPathsVariable(dockerfilePaths, leg);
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -262,21 +262,28 @@ private static void AddImageBuilderPathsVariable(string[] dockerfilePaths, Build
...
             IGrouping<PlatformId, PlatformInfo> platformGrouping)
         {
-            // Pass 1: Get the set of subgraphs grouped by their FROM dependencies as well as any integral custom leg dependencies.
+            IEnumerable<PlatformInfo> allPlatforms = Manifest.GetAllPlatforms();
+
+            // Pass 1: Get the set of subgraphs of all platforms grouped by their FROM dependencies as well as any integral custom leg dependencies.
             IEnumerable<IEnumerable<PlatformInfo>> subgraphs = platformGrouping
                 .GetCompleteSubgraphs(platform =>
-                    Manifest.GetParents(platform, platformGrouping)
+                    Manifest.GetParents(platform, allPlatforms)
```
mthalman, line 270: A comment should be added here explaining why we're using _all_ platforms. Something like:

> In some cases, there can be a root Dockerfile that isn't to be included in the matrix leg. An [example](https://github.com/dotnet/docker-tools/issues/1141) of this is when generating a test matrix where the runtime-deps Dockerfile hasn't been built (pulled from cache) and so it isn't included in the matrix leg because it doesn't need to be tested. But it has two sibling dependencies: runtime and aspnet-composite. These would end up getting placed in separate build legs if the set of parents that we used to walk their graph was limited to those platforms (Dockerfiles) which are to be included in the matrix leg. For that reason, we use _all_ platforms as the source set from which to find parents to walk the graph. And then we later filter out those platforms which are not intended to be included in the matrix leg.

Feel free to modify.

src/Microsoft.DotNet.ImageBuilder/tests/GenerateBuildMatrixCommandTests.cs
```diff
@@ -689,6 +689,147 @@ public void PlatformVersionedOs_Cached(bool isRuntimeCached, bool isAspnetCached
             }
         }
 
+        /// <summary>
+        /// Verifies that the platformVersionedOs matrix type doesn't create mutiple legs
+        /// for a scenario where platforms in the same repo share a cached parent image.
```
mthalman, line 694: Add a link to https://github.com/dotnet/docker-tools/issues/1141

mthalman (approved):
Wonderful. I have to say I'm surprised this worked with minimal changes. Just a few suggestions on code comments.

PR #1143, opened 2023-06-06
Fix build matrix graph generation to include grandparents in calculation

lbussell:
Fixes https://github.com/dotnet/docker-tools/issues/1141

I did not test https://github.com/dotnet/docker-tools/pull/1142 enough and it turns out it was still an issue in some cases. In this case, the PlatformVersionedOS matrix graph generation would create multiple legs for the same OS when we needed to build images with a shared cached parent that were not direct siblings. For example, generating a matrix when we have built both `aspnet` and `aspnet-composite` would end up creating two separate test graphs - one with `[runtime-deps, aspnet-composite]` and the other with `[runtime, aspnet, sdk]` (that's before we filter the images down in a later step).

```mermaid
flowchart TD
    A[runtime-deps] --> B
    A --> E[aspnet-composite]
    B[runtime] --> C
    C[aspnet] --> D[sdk]
```

The issue is that the `CreateNodeList` method didn't include parents-of-parents in its calculation. I changed it to be recursive. We don't need to dig deeper into children because ~~our Dockerfiles don't branch beneath runtime-deps, no children have multiple descendants. However I could be convinced to add walking the children graph to `CreateNodeList`.~~ because parent images that aren't cached imply all their children will also be not cached.

I also added some extra testing to cover this case. I added SDK to the test graph to make it more representative of the actual build but I don't think it makes a difference.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/tests/GenerateBuildMatrixCommandTests.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/tests/GenerateBuildMatrixCommandTests.cs
@@ -697,9 +697,9 @@ namespace Microsoft.DotNet.ImageBuilder.Tests
         /// https://github.com/dotnet/docker-tools/issues/1141
         /// </remarks>
         [Theory]
-        [InlineData(false, false, "--path 1.0/runtime/os/Dockerfile --path 1.0/aspnet/os-composite/Dockerfile --path 1.0/aspnet/os/Dockerfile")]
-        [InlineData(true, false, "--path 1.0/aspnet/os/Dockerfile --path 1.0/aspnet/os-composite/Dockerfile")]
-        [InlineData(true, true, "--path 1.0/aspnet/os-composite/Dockerfile")]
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/tests/GenerateBuildMatrixCommandTests.cs
```diff
@@ -812,7 +825,22 @@ public void PlatformVersionedOs_CachedParent()
...
+                    {
+                        Repo = "sdk",
+                        Images =
+                        {
+                            new ImageData
+                            {
+                                ProductVersion = "1.0",
+                                Platforms =
+                                {
+                                    CreateSimplePlatformData(sdkDockerfilePath, isCached: true)
```
mthalman, line 843: Based on my other comment about the parent hierarchy and caching, it doesn't seem valid for the sdk to always be cached when it's possible for its parent to not be cached. For purposes of this test, can't `isCached` always be set to `false` here and still give us the test coverage we need?

mthalman (approved):

lbussell:
I accidentally created an sdk->sdk cycle in the manifest defined in the test. Fixed that. Glad it wasn't an issue with the actual code - we should never have cycles in our actual Dockerfiles/manifests.

PR #1163, opened 2023-07-24
Switch Windows Server 2022 CI legs to 1ES hosted pools

lbussell:
This mitigates an issue that we're having with our Server 2022 pools at the moment. I ran some internal tests and noticed no issues using the 1ES hosted pools. We can switch back once our custom pools are functional again.


mthalman (approved):
cc @dagood

Issue #1159, opened 2023-07-12
Enable use of BuildKit in CI


lbussell:
We want to be able to build Dockerfiles and samples with newer Docker functionality (multi-platform builds, etc). See https://github.com/dotnet/dotnet-docker/pull/4742.

We previously tried to enable BuildKit in ImageBuilder, but encountered an issue where multi-stage Dockerfiles will lose their variant (e.g. arm64v8 becomes arm64), likely due to us building all arm32v7 and arm64v8 Dockerfiles on an arm64 host, and not specifying the build target for the final stage in the Dockerfile build. See #1132 and #1136.

Now we should investigate whether or not we can install and use buildx on our build agents to resolve this issue.

lbussell:
Related:
- https://github.com/dotnet/docker-tools/issues/1017

This is the same issue, but what I have observed is the opposite behavior. @mthalman mentions in https://github.com/dotnet/dotnet-docker/issues/4438#issuecomment-1440681555 that the issue might have been resolved due to build agent updates, so that might have changed the behavior here.

lbussell:
@richlander since you are the resident expert on multi-platform Docker builds, can you provide some insight here? If having no `--platform=` specified in the `FROM` instruction implies that `$TARGETPLATFORM` will be the default platform for the final stage, why would the output of the image be different from what's passed in the `--platform` argument to `docker build`? Could this be a consequence of not using `buildx`? We are not yet trying to build multiple platforms at once, just one at a time.

I am still having troubles getting buildx working on Azure build agents. Docker has a ["setup buildx" GitHub Action](https://github.com/marketplace/actions/docker-setup-buildx), but GitHub Actions runners already have buildx installed.

richlander:
Let's assume that `$BUILDPLATFORM` is not used in the Dockerfile to make this discussion simple.

The value of `--platform` (via the CLI) will be used for two things:

- The platform to use to select an image from any manifest tag.
- The platform to write to the final image, even if it is wrong.

Here's a good example.

```bash
$ echo FROM arm32v7/alpine > Dockerfile
$ docker build -t test --platform linux/amd64 .
$ docker inspect test | grep Arch
        "Architecture": "amd64",
$ docker run --rm test uname -a
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
[...]

PR #1169, opened 2023-07-28
Move back to alpine floating tag and don't check variant for arm64/(v8)

### LINKED ISSUE #1161: Align ImageBuilder image platform validation with containerd
There is a bug in the Docker legacy builder that results in `arm64` images being produced with the `arm64/v8` variant. When switching to BuildKit, the `arm64/v8` platform gets normalized to `arm64` with no variant due to [this code ](https://github.com/containerd/containerd/blob/96de54db4385279a5f8474b5de6658f6336b6fd8/platforms/database.go#L87-L92)in containerd.

Currently, ImageBuilder checks the variant here, but doesn't take into account that `arm64` and `arm64/v8` are supposedly compatible:
https://github.com/dotnet/docker-tools/blob/1bc4ea539f067c09d19627ca16ec9fcc38b240b5/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs#L412-L437

See https://github.com/moby/buildkit/issues/4039 for more details.

This is blocking https://github.com/dotnet/docker-tools/issues/1159

lbussell:
This should fix #1161 and #1159

I am listening for feedback on how else I can validate that this change won't cause problems.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/manifest.json
+++ b/src/Microsoft.DotNet.ImageBuilder/manifest.json
@@ -36,8 +36,7 @@
               "tags": {
                 "linux-arm64": {},
                 "linux-arm64-$(UniqueId)": {}
-              },
-              "variant": "v8"
+              }
             },
             {
               "buildArgs": {
--- a/src/Microsoft.DotNet.ImageBuilder/manifest.json
+++ b/src/Microsoft.DotNet.ImageBuilder/manifest.json
@@ -36,7 +36,8 @@
               "tags": {
                 "linux-arm64": {},
                 "linux-arm64-$(UniqueId)": {}
-              }
+              },
+              "variant": "v8"
             },
             {
               "buildArgs": {
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -427,6 +427,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             // We still want to check variants if either variant is not "v8" or empty.
             // See https://github.com/moby/buildkit/issues/4039
             bool skipVariantCheck = platform.Model.Architecture == Models.Manifest.Architecture.ARM64
+                && baseImageArch == Models.Manifest.Architecture.ARM64
                 && ((platform.Model.Variant == "v8" || string.IsNullOrEmpty(platform.Model.Variant))
                     && (baseImageVariant == "v8" || string.IsNullOrEmpty(baseImageVariant)));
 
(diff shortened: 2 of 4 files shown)


src/Microsoft.DotNet.ImageBuilder/manifest.json
```diff
@@ -36,8 +36,7 @@
               "tags": {
                 "linux-arm64": {},
                 "linux-arm64-$(UniqueId)": {}
-              },
-              "variant": "v8"
```
mthalman, line 40: Why is this removed? It doesn't seem necessary and is also causing the build to fail since CI is still using the "previous" version of Image Builder without these changes.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -419,10 +419,18 @@ private void ValidatePlatformIsCompatibleWithBaseImage(PlatformInfo platform)
...
-            (Models.Manifest.Architecture arch, string? variant) =
+            (Models.Manifest.Architecture baseImageArch, string? baseImageVariant) =
                 _dockerService.GetImageArch(baseImageTag, Options.IsDryRun);
 
-            if (platform.Model.Architecture != arch || platform.Model.Variant != variant)
+            // Containerd normalizes arm64/v8 to arm64 with no variant.
+            // In other words, arm64/v8 and arm64/ are compatible.
+            // We still want to check variants if either variant is not "v8" or empty.
+            // See https://github.com/moby/buildkit/issues/4039
+            bool skipVariantCheck = platform.Model.Architecture == Models.Manifest.Architecture.ARM64
```
mthalman, line 429: We also only want to skip the check if the base image architecture is Arm64. Otherwise, we could erroneously skip the check if the base image was amd64, for example.

src/Microsoft.DotNet.ImageBuilder/manifest.json
lbussell: I thought that it was appropriate given that the image we produce won't have the v8 variant. However, you're right, it will have the variant until we are actually using the ImageBuilder created by these changes... we can worry about that later.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
lbussell: I figured that that would be caught by the architecture check, which is never skipped. But I do think your suggestion will make the variable more correct so I added it.

src/Microsoft.DotNet.ImageBuilder/manifest.json
mthalman: I don't feel that Docker's behavior of putting the variant in the metadata or not should dictate how the manifest is described. For example, the dotnet-docker's tagging pattern uses the variant metadata. So we can't remove the variant from the manifest in that repo without impacting the tags we produce. So we should be consistent with that here as well.

PR #1173, opened 2023-08-09
Add workaround for incorrect BuildKit file caching behavior

lbussell:
This fixes the issue that we saw for patch Tuesday where the wrong Dockerfile was cached for some 7.0 runtime Dockerfiles.

Please see https://github.com/moby/buildkit/issues/1368

It's fixed in Docker version 23, but the bug is in the agent code and not the client, so we would have to wait until our build agents are updated.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/DockerService.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.IO;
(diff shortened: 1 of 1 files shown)


mthalman (approved):
Great sleuthing to find the root issue.

MichaelSimons:
Please log a follow-up issue to remove this workaround once the buildkit fix is picked up.

lbussell:
> Please log a follow-up issue to remove this workaround once the buildkit fix is picked up.

Filed https://github.com/dotnet/docker-tools/issues/1174

Issue #1162, opened 2023-07-20
dotnet-docker-internal publish result exceeds GitHub issue character limit


lbussell:
In [a recent internal build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2225611&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=307ab20e-da09-58a9-37fd-320ab9135ca4) (internal link), the issue body for the publish result notification exceeded Github's issue character limit of 65536:
```
{"message":"Validation Failed","errors":[{"resource":"Issue","code":"custom","field":"body","message":"body is too long (maximum is 65536 characters)"}],"documentation_url":"[https://docs.github.com/rest/issues/issues#create-an-issue"}](https://docs.github.com/rest/issues/issues#create-an-issue%22})
```

mthalman:
This is a reoccurrence of https://github.com/dotnet/docker-tools/issues/1013.

We're just expanding the number of images. For example, we'd added aspnet composite images. So this increases the likelihood of hitting this limit.

The main desire is to have the digests included in the issue content so that we can search by digest and find the issue.

One solution may be to simply list all the digests in the content without any associated repos or tags. Then include an attached txt file that has the same content that we have today with repo grouping and tags associated with each digest. That will allow us to still search by digest but then dive into more detail by opening the attached txt file. Not ideal, but we're limited.

Another option is to store the content as a markdown file in the repo and have the issue link to that. In that case, maybe it makes more sense for it to be a PR that adds the file and we rely on notification via PR rather than issue. Then the PR can be automatically merged.

mthalman:
[Triage]
Another option might be to break up the total content so that it has multiple discussion posts in the issue. For example, each Docker repo could be a separate post that lists the images for that repo.

lbussell:
I think that it would make sense to keep the existing issue, and post each individual repo's images as a comment on the issue. Each comment has its own character limit of 65536 so I'd be impressed if we grew past that. I will file an issue in Arcade to add comment support to `Microsoft.DotNet.Git.IssueManager`.

lbussell:
Submitted and self-assigned https://github.com/dotnet/arcade/issues/14048

Issue #1181, opened 2023-09-25
Move .NET Docker projects to .NET 8.0 TFM


lbussell:
Here's a search query that should return all instances of the `net7.0` TFM in our repos: https://github.com/search?q=net7.0+repo%3Adotnet%2Fdotnet-docker+OR+repo%3Adotnet%2Fdocker-tools+OR+repo%3Adotnet%2Fdotnet-buildtools-prereqs-docker+OR+repo%3Amicrosoft%2Fdotnet-framework-docker++path%3A*.csproj&type=code

### .NET Docker

- [x] update-dependencies
- [x] update-dependencies Dockerfile
- [x] Microsoft.DotNet.Docker.Tests

### Docker tools

- [x] file-pusher
- [x] file-pusher Dockerfile
- [x] yaml-updater
- [x] yaml-updater Dockerfile
- [x] ImageBuilder
- [x] ImageBuilder Dockerfile
- [x] ImageBuilder Tests

lbussell:
[Triage] This is completed with the above referenced PRs.

PR #1186, opened 2023-10-13
Break publish notifications up into comments

lbussell:
Fixes https://github.com/dotnet/docker-tools/issues/1162

New Arcade packages target .NET 8.0, so I needed to upgrade ImageBuilder to .NET 8 as well. Took the opportunity to make the Dockerfile multi-platform as well (I have been using this modified Dockerfile for some time to test ImageBuilder in arm64 CI anyways). And as far as I could tell, the LibGit2Sharp issues are all closed so I removed the RID weirdness.

I tested this manually by pointing at a private repo of mine.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/manifest.json
+++ b/src/Microsoft.DotNet.ImageBuilder/manifest.json
@@ -33,19 +33,21 @@
             },
             {
               "buildArgs": {
-                "WINDOWS_BASE": "servercore:ltsc2016-amd64"
+                "WINDOWS_BASE": "servercore:ltsc2019-amd64",
+                "WINDOWS_SDK": "windowsservercore-ltsc2019"
               },
               "dockerfile": "Dockerfile.windows",
               "os": "windows",
               "osVersion": "windowsservercore-ltsc2019",
               "tags": {
-                "windowsservercore-ltsc2016-amd64": {},
-                "windowsservercore-ltsc2016-amd64-$(UniqueId)": {}
+                "windowsservercore-ltsc2019-amd64": {},
+                "windowsservercore-ltsc2019-amd64-$(UniqueId)": {}
               }
             },
             {
               "buildArgs": {
-                "WINDOWS_BASE": "nanoserver:1809-amd64"
+                "WINDOWS_BASE": "nanoserver:1809-amd64",
+                "WINDOWS_SDK": "nanoserver-1809"
               },
               "dockerfile": "Dockerfile.windows",
               "os": "windows",
@@ -57,7 +59,8 @@
             },
             {
               "buildArgs": {
-                "WINDOWS_BASE": "nanoserver:ltsc2022-amd64"
+                "WINDOWS_BASE": "nanoserver:ltsc2022-amd64",
+                "WINDOWS_SDK": "nanoserver-ltsc2022"
               },
               "dockerfile": "Dockerfile.windows",
               "os": "windows",
--- a/src/Microsoft.DotNet.ImageBuilder/manifest.json
+++ b/src/Microsoft.DotNet.ImageBuilder/manifest.json
@@ -33,15 +33,15 @@
             },
             {
               "buildArgs": {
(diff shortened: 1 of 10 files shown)


mthalman (approved):

src/Microsoft.DotNet.ImageBuilder/manifest.json
```diff
@@ -33,19 +33,21 @@
...
+                "WINDOWS_SDK": "windowsservercore-ltsc2019"
               },
               "dockerfile": "Dockerfile.windows",
               "os": "windows",
               "osVersion": "windowsservercore-ltsc2019",
               "tags": {
-                "windowsservercore-ltsc2016-amd64": {},
-                "windowsservercore-ltsc2016-amd64-$(UniqueId)": {}
+                "windowsservercore-ltsc2019-amd64": {},
+                "windowsservercore-ltsc2019-amd64-$(UniqueId)": {}
```
mthalman, line 44: This should all stay at ltsc2016. We need a version of Image Builder that can be used for our WS 2016 builds.

src/Microsoft.DotNet.ImageBuilder/src/NotificationService.cs
```diff
@@ -33,20 +33,20 @@ public NotificationService(ILoggerService loggerService)
         {
             IssueManager issueManager = new(gitHubAccessToken);
 
-            int issueId = 0;
+            Uri issueUrl = new("https://github.com/404");
```
mthalman, line 36: To avoid confusion, I think this should just be a nullable `Uri` that gets defaulted to `null` instead of inventing a fake URL.

src/Microsoft.DotNet.ImageBuilder/src/NotificationService.cs
```diff
@@ -56,6 +56,16 @@ public NotificationService(ILoggerService loggerService)
...
+            if (comments != null)
+            {
+                for (int i = 0; i < comments.Count(); i++)
+                {
+                    _loggerService.WriteMessage($"====COMMENT {i + 1} MARKDOWN===");
+                    _loggerService.WriteMessage(comments.ElementAt(i));
+                }
+            }
+
             _loggerService.WriteMessage($"====END DESCRIPTION MARKDOWN===");
```
mthalman, line 69: This should be placed above the start of the comments markdown.

src/Microsoft.DotNet.ImageBuilder/manifest.json
lbussell: So then should the `osVersion` be ltsc2016 as well?

mthalman: You can try it. I think there was a reason I needed to make it `ltsc2019` to work around a problem.

mthalman: Oh, right. I remember. It determines which build agent is used. We want this to build on a WS 2019 machine so it should stay as is.

src/Microsoft.DotNet.ImageBuilder/manifest.json
```diff
@@ -41,7 +33,8 @@
             },
             {
               "buildArgs": {
-                "WINDOWS_BASE": "servercore:ltsc2016-amd64"
+                "WINDOWS_BASE": "servercore:ltsc2016-amd64",
+                "WINDOWS_SDK": "windowsservercore-ltsc2019"
```
mthalman, line 37: This might as well be `nanoserver-1809`. It's smaller and will pull faster. We don't need WSC just to build the project.

src/Microsoft.DotNet.ImageBuilder/manifest.json
lbussell: Yep, that's what I thought right after I asked that.

mthalman (approved):

lbussell:
This contains part of #1181

PR #1214, opened 2024-03-14
Convert the check-base-image-updates pipeline to use 1ES templates

MichaelSimons:
Related to: https://github.com/dotnet/dotnet-docker-internal/issues/4475

Test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2404434 (internal Microsoft link)

PR #1217, opened 2024-03-15
Fix pipeline stage names

MichaelSimons:
The check-base-image-updates pipeline stage name that was added in https://github.com/dotnet/docker-tools/pull/1214 is invalid as Names cannot contain spaces.  Fixing that as well as adding a more descriptive name for the upload-file pipeline stage.


eng/pipelines/check-base-image-updates.yml
```diff
@@ -27,7 +27,8 @@ extends:
     customBuildTags:
     - ES365AIMigrationTooling
     stages:
-    - stage: Check Base Images
+    - stage:
```
lbussell, line 30:
```suggestion
    - stage: CheckBaseImages
```

I haven't seen any documentation that suggests an empty stage or job name will be accepted.

eng/pipelines/upload-file.yml
```diff
@@ -30,7 +30,8 @@ extends:
     customBuildTags:
     - ES365AIMigrationTooling
     stages:
-    - stage: stage
+    - stage:
```
lbussell, line 33:
```suggestion
    - stage: UploadFile
```

lbussell (changes requested):

eng/pipelines/check-base-image-updates.yml
MichaelSimons: FWIW the pipeline works without one.

lbussell: You learn something new every day.

mthalman (approved):

Issue #1213, opened 2024-03-14
Consider removing the image cleanup logic in pipelines


mthalman:
The common pipeline implementation contains logic to cleanup Docker state before and after images are built:
* https://github.com/dotnet/docker-tools/blob/main/eng/common/templates/steps/cleanup-docker-linux.yml
* https://github.com/dotnet/docker-tools/blob/main/eng/common/templates/steps/cleanup-docker-windows.yml

These steps can add a fair amount of execution time, especially on Windows agents. Some recent runs have shown this taking 5 minutes on a Windows agent.

At the time this was implemented, build agents were potentially reused between runs and it was important to keep the state of the agents clean. Since then, however, we're using stateless pools that provide fresh agents each time. So there really isn't much need to clean the agent. The one exception might be for cases where we would run out of disk space. That's most likely to happen in the .NET Framework pipeline.

So, by default, I propose that we eliminate the clean steps. If it's necessary to clean for disk space purposes, this is something that can be opted into.

lbussell:
Removed the majority of the cleanup steps with https://github.com/dotnet/docker-tools/pull/1229.

PR #1229, opened 2024-03-22
Remove certain instances of Docker image cleanup

lbussell:
Part of #1213 

These changes are intended to...

Remove the Docker cleanup step:
- From the end of pipeline runs
- Before pre-build validation and matrix generation

Leave the cleanup step:
- Before building any images
- Before mirroring/pulling any base images 

Test run (still running at time of submission, but it's been through a few iterations): https://dev.azure.com/dnceng/internal/_build/results?buildId=2412187&view=results

### DIFF
--- a/eng/common/templates/jobs/copy-base-images.yml
+++ b/eng/common/templates/jobs/copy-base-images.yml
@@ -16,4 +16,3 @@ jobs:
       additionalOptions: ${{ parameters.additionalOptions }}
       publicProjectName: ${{ parameters.publicProjectName }}
       continueOnError: true
-  - template: /eng/common/templates/steps/cleanup-docker-linux.yml@self
--- a/eng/common/templates/jobs/copy-base-images.yml
+++ b/eng/common/templates/jobs/copy-base-images.yml
@@ -10,8 +10,6 @@ jobs:
   pool: ${{ parameters.pool }}
   steps:
   - template: /eng/common/templates/steps/init-docker-linux.yml@self
-    parameters:
-      cleanupDocker: false
   - ${{ parameters.customInitSteps }}
   - template: /eng/common/templates/steps/copy-base-images.yml@self
     parameters:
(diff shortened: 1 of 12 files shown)


eng/common/templates/jobs/copy-base-images.yml
mthalman, line 1: This should also skip cleanup in the init step. This job isn't building anything.

eng/common/templates/jobs/publish.yml
mthalman, line 1: This should also skip cleanup in the init step. This job isn't building anything.

eng/common/templates/jobs/wait-for-ingestion.yml
mthalman, line 1: This should also skip cleanup in the init step. This job isn't building anything.

eng/pipelines/cleanup-acr-images-custom.yml
mthalman, line 1: This should also skip cleanup in the init step. This job isn't building anything.

eng/pipelines/cleanup-acr-images.yml
mthalman, line 1: This should also skip cleanup in the init step. This job isn't building anything.

eng/pipelines/templates/jobs/check-base-image-updates.yml
mthalman, line 1: This should also skip cleanup in the init step. This job isn't building anything.

lbussell:
@mthalman thoughts on changing the default value of `cleanupDocker` to false? It really only needs to be set in builds. Would save some lines of code over what's in this PR.

mthalman:
> @mthalman thoughts on changing the default value of `cleanupDocker` to false? It really only needs to be set in builds. Would save some lines of code over what's in this PR.

Yes, that sounds like a good idea.

lbussell:
New test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2413966&view=results

mthalman (approved):

PR #1235, opened 2024-04-02
Use Mariner ARM64 VM image for public builds

lbussell:
I am hoping that this will address some public build flakiness. This will also bring the public CI in-line with the internal CI by using the same VM image.

### DIFF
--- a/eng/common/templates/stages/dotnet/build-test-publish-repo.yml
+++ b/eng/common/templates/stages/dotnet/build-test-publish-repo.yml
@@ -76,20 +76,20 @@ stages:
     linuxArm64Pool:
       os: linux
       hostArchitecture: Arm64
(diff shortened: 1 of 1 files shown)


mthalman (approved):

lbussell:
Oops, I merged this too early and assumed AzDo resolved the correct image. It in fact resolved the existing ubuntu image which is the default for the pool. It makes no difference but the yaml is misleading now. I'll submit a follow-up PR, or make sure the new image is actually working. https://dev.azure.com/dnceng-public/public/_build/results?buildId=627622&view=logs&j=bbd5db2d-92c1-5c19-fb34-868147c7bf3c&t=5877679a-4ada-4857-a3e6-f02e9004b252&l=7

PR #1239, opened 2024-04-12
Use custom templates for 1ES Pipeline Templates

lbussell:
This PR:
- Should fix https://github.com/dotnet/dotnet-docker/issues/5332
  - It's hard to validate that it actually fixes this. However I've got a totally green build from dotnet-docker-nightly-playground [here](https://dev.azure.com/dnceng/internal/_build/results?buildId=2427926&view=results) [internal link]. I think it is ignoring the /versions directory correctly.
- Reduces lots of duplication
- Centralizes the location where we define which pool we use for SDL stages
  - We could probably do better with the definition. I put it in a parameter, since parameters let you hold arbitrary yaml objects. However we could just inline the whole thing and put all the values in variables.

### DIFF
--- a/eng/common/templates/1es-official.yml
+++ b/eng/common/templates/1es-official.yml
@@ -1,0 +1,48 @@
+# When extending this template, pipelines using a repository resource containing versions files for image caching must
+# do the following:
+#
+# - Do not rely on any source code from the versions repo so as to not circumvent SDL and CG guidelines
+# - The versions repo resource must be named `InternalVersionsRepo` or `PublicVersionsRepo` to avoid SDL scans
+# - The versions repo must be checked out to `$(Build.SourcesDirectory)/versions` to avoid CG scans
+#
+# If the pipeline is not using a separate repository resource, ensure that there is no source code checked out in
+# `$(Build.SourcesDirectory)/versions`, as it will not be scanned.
+#
+# The `cgDryRun` parameter will run CG but not submit the results, for testing purposes.
+
+parameters:
+- name: cgDryRun
+  type: boolean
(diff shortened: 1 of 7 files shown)


eng/pipelines/dotnet-buildtools-image-builder-official.yml
```diff
@@ -6,27 +6,21 @@ trigger:
...
     - src/*
-
 pr: none
 
+resources:
+  repositories:
+  - repository: InternalVersionsRepo
+    type: github
+    endpoint: dotnet
+    name: dotnet/versions
```
mthalman, line 16: Why is this needed?

eng/common/templates/1es-official.yml
```diff
@@ -0,0 +1,36 @@
...
+    pool: ${{ parameters.pool }}
+    sdl:
+      componentgovernance:
+        ignoreDirectories: $(Build.SourcesDirectory)/versions
+        whatIf: ${{ parameters.cgDryRun }}
+        showAlertLink: true
+      sourceRepositoriesToScan:
+        exclude:
+        - repository: InternalVersionsRepo
+        - repository: PublicVersionsRepo
```
mthalman, line 35: There's an implicit contract here with consuming pipelines, which is fine because it simplifies things but it should be documented. Namely, there should be comments at the top that specify that the use of the versions repo means that the consuming pipeline must have a repository resource named either `InternalVersionsRepo` or `PublicVersionsRepo`, and be mapped to the `versions` directory in sources.

mthalman (approved):

PR #1240, opened 2024-04-12
Add support for Azure Linux to ImageBuilder

MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker/issues/5331

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
@@ -238,6 +238,10 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
                 {
                     displayName = FormatVersionableOsName(os, name => name.FirstCharToUpper());
                 }
+                else if (os.Contains("azure-linux"))
+                {
+                    displayName = FormatVersionableOsName(os, name => "Azure Linux");
+                }
                 else if (os.Contains("cbl-mariner"))
                 {
                     displayName = FormatVersionableOsName(os, name => "CBL-Mariner");


mthalman (approved):

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs
```diff
@@ -238,6 +238,10 @@ public string GetOSDisplayName()
                 {
                     displayName = FormatVersionableOsName(os, name => name.FirstCharToUpper());
                 }
+                else if (os.Contains("azure-linux"))
```
lbussell, line 241: Should this be `azurelinux`? Following the tagging scheme from the base image:

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/58623a3bc566341cec68e76782a8e75fceb9a8d5/src/azurelinux/3.0/crossdeps-builder-net8.0/Dockerfile#L1

and the tags for buildtools-prereqs:

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/58623a3bc566341cec68e76782a8e75fceb9a8d5/src/azurelinux/manifest.json#L12-L17

mthalman: Good point. That identifier is already used https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/58623a3bc566341cec68e76782a8e75fceb9a8d5/src/azurelinux/manifest.json#L11 so we might as well go with that. Otherwise, we'd need to change it in that repo.

MichaelSimons: I agree as well.  Thanks for pointing that out.

MichaelSimons: https://github.com/dotnet/docker-tools/pull/1242

PR #1244, opened 2024-04-17
Fix test result publishing

lbussell:
See https://github.com/dotnet/docker-tools/issues/1246

- Testing this fix in PR validation since it affects public builds
- Rearrange docker copy and test results check commands


lbussell:
The correct solution to the issue of having no test results to publish is https://github.com/dotnet/docker-tools/issues/1250.

lbussell:
This change should go in as-is and https://github.com/dotnet/docker-tools/issues/1250 should be addressed separately so that test results are available again as soon as possible.

mthalman (approved):

Issue #1176, opened 2023-08-28
Microsoft.IdentityModel.Clients.ActiveDirectory is deprecated


mthalman:
Image Builder has a dependency on Microsoft.IdentityModel.Clients.ActiveDirectory which is marked as deprecated. This should be [migrated to use MSAL.NET instead](https://learn.microsoft.com/entra/msal/dotnet/#migration-from-azure-active-directory-authentication-library-adal). This has shown up in the component governance report.

mthalman:
This should take precedence over the work in https://github.com/dotnet/docker-tools/issues/1123.

lbussell:
This now has a hard due date of April 13 2024.

lbussell:
This is blocking ImageBuilder builds since last Friday 03/15/2024 - https://dev.azure.com/dnceng/internal/_build/results?buildId=2405989&view=logs&j=ac422da1-c6cf-5423-8ede-22ab7060639a&t=898b1438-1a8f-5482-ed55-51c989cf5404&l=818

lbussell:
The due date was extended to 04/01/2024.

lbussell:
Related https://github.com/dotnet/dotnet-docker-internal/issues/4648 [internal issue]

Issue #1123, opened 2023-02-27
Microsoft.Azure.Management.Fluent package is deprecated


mthalman:
Image Builder has a dependency on `Microsoft.Azure.Management.Fluent` which is marked as deprecated. The [deprecation notice](https://www.nuget.org/packages/Microsoft.Azure.Management.Fluent) indicates that `Azure.ResourceManager` be used as an alternative and they have a migration guide at https://aka.ms/dotnet/t2/migration. This has shown up in the component governance report.

mthalman:
The work for https://github.com/dotnet/docker-tools/issues/1176 should be done before this work.

PR #1252, opened 2024-04-24
Migrate Import Image to use Azure.ResourceManager.ContainerRegistry

mthalman:
This migrates the "Import Image" logic from the deprecated Azure Fluent API to the modern Azure.ResourceManager.ContainerRegistry package.

This affects the functionality of the `copyAcrImages` and `copyBaseImages` commands. I've tested both of those commands with these changes and they're both working.

Overall, it's a pretty straightforward 1:1 migration. Just a new API to adjust to.


lbussell (approved):
Thanks!

lbussell:
Part of https://github.com/dotnet/docker-tools/issues/1176 and https://github.com/dotnet/docker-tools/issues/1123

Issue #1254, opened 2024-04-24
Tenant variable not set in public validation pipeline


mthalman:
The changes from https://github.com/dotnet/docker-tools/pull/1252 are causing a failure in the docker-tools-eng-validation-pr pipeline during the `Copy Images` step of the Publish stage:

```
Unhandled exception: System.ArgumentException: Invalid tenant id provided. You can locate your tenant id by following the instructions listed here: https://learn.microsoft.com/partner-center/find-ids-and-domain-names (Parameter 'tenantId')
   at Azure.Identity.Validations.ValidateTenantId(String tenantId, String argumentName, Boolean allowNull)
   at Azure.Identity.ClientSecretCredential..ctor(String tenantId, String clientId, String clientSecret, TokenCredentialOptions options, CredentialPipeline pipeline, MsalConfidentialClient client)
   at Azure.Identity.ClientSecretCredential..ctor(String tenantId, String clientId, String clientSecret)
   at Microsoft.DotNet.ImageBuilder.CopyImageService.ImportImageAsync(String subscription, String resourceGroup, ServicePrincipalOptions servicePrincipalOptions, String[] destTagNames, String destRegistryName, String srcTagName, String srcRegistryName, ResourceIdentifier srcResourceId, ContainerRegistryImportSourceCredentials sourceCredentials, Boolean isDryRun) in /image-builder/src/CopyImageService.cs:line 48
   at Microsoft.DotNet.ImageBuilder.Commands.CopyAcrImagesCommand.ExecuteAsync() in /image-builder/src/Commands/CopyAcrImagesCommand.cs:line 60
   at System.CommandLine.Invocation.CommandHandler.GetResultCodeAsync(Object value, InvocationContext context)
   at System.CommandLine.Invocation.ModelBindingCommandHandler.InvokeAsync(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
```

This happens specifically because this is a public pipeline. The YAML is configured to pass a tenant from a variable:

https://github.com/dotnet/docker-tools/blob/736b6b0f3ea2923a754a51ec7726074e7fe5ebb7/eng/common/templates/jobs/publish.yml#L71

But that variable is not set in the pipeline. It only exists in the internal pipeline via a variable group.

lbussell:
[...]

Issue #1256, opened 2024-04-25
Failing CopyBaseImages step doesn't fail the PreBuildValidation pipeline stage


lbussell:
See the output in this public PR validation pipeline run: https://dev.azure.com/dnceng-public/public/_build/results?buildId=656237&view=logs&j=1a3858d4-df16-5910-9f2f-b79bed34300b&t=753b3247-b9b0-566e-d813-4e57c0e7e6de&l=108

Even though the CopyBaseImages step has failed with an exception, the pipeline doesn't fail on the PreBuildValidation stage.

mthalman:
That's intentional: https://github.com/dotnet/docker-tools/blob/8fa03e26afbf7676865f2335c0b2f78baf5dbfe9/eng/common/templates/jobs/copy-base-images.yml#L18

It's to ensure our build won't be blocked by unavailability of the base images (e.g. Docker Hub registry down or something). But yes, it can be a problem for legitimate issues on our side.

lbussell:
OK. There's probably some solution we could make where an exception in ImageBuilder fails but an exception from PowerShell (for example) throws a warning and skips the rest of the stage. But it's not a pressing issue so closing as not planned unless we start having more issues.

PR #1257, opened 2024-04-25
Define default values for common variables

mthalman:
The changes that were applied in https://github.com/dotnet/docker-tools/pull/1251/commits/6c89f3faa08770f2b757d7e70818bee0c474b724 to fix https://github.com/dotnet/docker-tools/issues/1254 aren't sufficient. Public builds for other pipelines also run into the same issue (https://github.com/dotnet/docker-tools/issues/1256).

I've moved these variables to the common variable template so that all pipelines can make use of these values. They will still be overridden by any variable which consumes the variable group that has these actual values.

### DIFF
--- a/eng/common/templates/jobs/generate-matrix.yml
+++ b/eng/common/templates/jobs/generate-matrix.yml
@@ -10,6 +10,7 @@ jobs:
 - job: ${{ parameters.name }}
   pool: ${{ parameters.pool }}
   steps:
+  - script: echo $(acr.servicePrincipalTenant)
   - template: /eng/common/templates/steps/retain-build.yml@self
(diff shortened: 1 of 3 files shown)


eng/common/templates/jobs/generate-matrix.yml
```diff
@@ -10,6 +10,7 @@ jobs:
 - job: ${{ parameters.name }}
   pool: ${{ parameters.pool }}
   steps:
+  - script: echo $(acr.servicePrincipalTenant)
```
lbussell, line 13: Was this change just for debugging?

mthalman: Yes indeed. Removed.

lbussell (approved):

PR #1259, opened 2024-04-26
Remove dependency on Azure Fluent package

mthalman:
This removes Image Builder's dependency on Microsoft.Azure.Management.Fluent, which is marked as deprecated. The remaining code actually mainly had a dependency on a transitive dependency of Microsoft.Azure.Management.Fluent: Microsoft.Rest.ClientRuntime.

Fixes https://github.com/dotnet/docker-tools/issues/1123

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/BasicAuthenticationCredentials.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/BasicAuthenticationCredentials.cs
@@ -4,9 +4,5 @@
 namespace Microsoft.DotNet.ImageBuilder;
 
 #nullable enable
-public class BasicAuthenticationCredentials(string userName, string password)
-{
-    public string UserName { get; } = userName;
-    public string Password { get; } = password;
-}
+public record struct BasicAuthenticationCredentials(string UserName, string Password);
 #nullable disable
--- a/src/Microsoft.DotNet.ImageBuilder/src/BasicAuthenticationCredentials.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/BasicAuthenticationCredentials.cs
@@ -1,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.ImageBuilder;
+
+#nullable enable
+public class BasicAuthenticationCredentials(string userName, string password)
+{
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.ImageBuilder/src/BasicAuthenticationCredentials.cs
```diff
@@ -0,0 +1,12 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.ImageBuilder;
+
+#nullable enable
+public class BasicAuthenticationCredentials(string userName, string password)
+{
+    public string UserName { get; } = userName;
+    public string Password { get; } = password;
+}
```
lbussell, line 11:
```suggestion
public record struct BasicAuthenticationCredentials(string UserName, string Password);
```

src/Microsoft.DotNet.ImageBuilder/src/RegistryServiceClient.cs
```diff
@@ -69,33 +69,21 @@ private async Task<HttpResponseMessage> SendRequestAsync(HttpRequestMessage requ
...
-            throw new HttpOperationException(
-                $"Response status code does not indicate success: {response.StatusCode}. Reason: '{response.ReasonPhrase}'. Error content:{Environment.NewLine}{errorContent}")
-            {
-                Body = errorContent,
-                Request = new HttpRequestMessageWrapper(request, requestContent),
-                Response = new HttpResponseMessageWrapper(response, errorContent)
-            };
+            throw new HttpRequestException(
+                $"Response status code does not indicate success: {response.StatusCode}. Reason: '{response.ReasonPhrase}'. Error content:{Environment.NewLine}{errorContent}");
         }
```
lbussell, line 87: Isn't this exactly what EnsureSeccussStatusCode below does?

lbussell: https://source.dot.net/#System.Net.Http/System/Net/Http/HttpResponseMessage.cs,161

mthalman: Not "exactly", no. `EnsureSuccessStatusCode` is only returning status code information. That's not very helpful. Because registry errors include content with their response, this exception includes the content in the exception message which is great for diagnostic purposes.

lbussell (approved):

PR #1261, opened 2024-04-29
Remove service principal authentication from CopyImages commands

lbussell:
I have the necessary pipeline changes ready to go when this version of ImageBuilder is built and the automated update PR is created.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/CopyImageService.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/CopyImageService.cs
@@ -11,7 +11,6 @@ using Azure.Identity;
 using Azure.ResourceManager;
 using Azure.ResourceManager.ContainerRegistry;
 using Azure.ResourceManager.ContainerRegistry.Models;
-using Azure.ResourceManager.Resources;
 using Microsoft.VisualStudio.Services.Common;
 
 namespace Microsoft.DotNet.ImageBuilder;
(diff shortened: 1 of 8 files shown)


src/Microsoft.DotNet.ImageBuilder/src/CopyImageService.cs
```diff
@@ -20,35 +19,50 @@ namespace Microsoft.DotNet.ImageBuilder;
...
 }
 
 [Export(typeof(ICopyImageService))]
 public class CopyImageService : ICopyImageService
 {
     private readonly ILoggerService _loggerService;
 
+    private ArmClient? _armClient;
+
+    private ArmClient ArmClient => _armClient ??= new ArmClient(new DefaultAzureCredential());
```
mthalman, line 40: This can be simplified. Since it's a field, it should have the `_` prefix and camelCasing. The consuming code needs to be updated as well.

```suggestion
    private readonly ArmClient _armClient = new ArmClient(new DefaultAzureCredential());
```

lbussell: Got it. I made it lazy since I wasn't sure if it'd blow up when we run it in the public - https://dev.azure.com/dnceng-public/public/_build/results?buildId=659648&view=logs&j=1a3858d4-df16-5910-9f2f-b79bed34300b&t=753b3247-b9b0-566e-d813-4e57c0e7e6de&l=11

But I'm actually not sure why it doesn't blow up now, since we're giving it bogus credentials, and this seems like a typo:
https://github.com/dotnet/docker-tools/blob/10c82d12cf90e24693067e9768eab929153e1fa2/eng/common/templates/steps/copy-base-images.yml#L22-L24

mthalman (approved):

PR #1262, opened 2024-04-29
Define common API for querying registries

mthalman:
This provides a common API, `IRegistryContentClient`, that provides an interface for querying container registries. This provides a common interface for both the ACR that we own as well as 3rd party registries. For our ACR, it uses the Azure library which provides the Azure auth implementation we require. The logic for determining which implementation to use is in `RegistryClientFactory`. For now, this uses a hardcoded registry name to determine which implementation it should use. This is temporary. Once I get signoff on these changes, I'll replace that so that the value can be passed in via the command call. That will modify a whole bunch of files so I want to wait until the end to make that change.


src/Microsoft.DotNet.ImageBuilder/src/ContainerRegistryContentClientWrapper.cs
```diff
@@ -0,0 +1,26 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text.Json.Nodes;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Containers.ContainerRegistry;
+
+namespace Microsoft.DotNet.ImageBuilder;
+
+internal class ContainerRegistryContentClientWrapper : IRegistryContentClient
```
lbussell, line 11:
```suggestion
#nullable enable
internal class ContainerRegistryContentClientWrapper : IRegistryContentClient
```

src/Microsoft.DotNet.ImageBuilder/src/RegistryClientFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+using Azure.Identity;
+
+namespace Microsoft.DotNet.ImageBuilder;
+
+#nullable enable
+[Export(typeof(IRegistryContentClientFactory))]
+[method: ImportingConstructor]
+internal class RegistryClientFactory(IHttpClientProvider httpClientProvider) : IRegistryContentClientFactory
+{
+    private readonly RegistryHttpClient _httpClient = httpClientProvider.GetRegistryClient();
```
lbussell, line 16: Is there an issue with getting this client when the RegistryClientFactory is initialized? It will be created but not used when we are interacting with an ACR. Perhaps we could make it lazy or get the client inline?

src/Microsoft.DotNet.ImageBuilder/src/RegistryClientFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+            DockerHelper.DockerHubApiRegistry :
+            registry!;
+
+        string ownedAcr = "dotnetdocker";
+
+        if (!ownedAcr.EndsWith(DockerHelper.AcrDomain))
+        {
+            ownedAcr = $"{ownedAcr}.{DockerHelper.AcrDomain}";
+        }
+
```
lbussell, line 31: The point of this code is unclear to me. The condition will always be true. Is the `ownedAcr` intended to be passed in as a parameter?

src/Microsoft.DotNet.ImageBuilder/src/IRegistryContentClient.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.ImageBuilder;
+
+#nullable enable
+public interface IRegistryContentClient
+{
+    Task<ManifestQueryResult> GetManifestAsync(string tagOrDigest);
+}
+#nullable disable
```
lbussell, line 13: My IDE says these closing `#nullable` directives are not necessary. I recognize that they're in most files in this repo, so if you're following the convention that's fine. Just thought I'd call it out since this is adding a new one.

src/Microsoft.DotNet.ImageBuilder/src/RegistryClientFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.ComponentModel.Composition;
+using Azure.Containers.ContainerRegistry;
+using Azure.Identity;
+
+namespace Microsoft.DotNet.ImageBuilder;
+
+#nullable enable
+[Export(typeof(IRegistryContentClientFactory))]
+[method: ImportingConstructor]
+internal class RegistryClientFactory(IHttpClientProvider httpClientProvider) : IRegistryContentClientFactory
```
lbussell, line 14: Is there a reason this isn't `RegistryContentClientFactory`?

src/Microsoft.DotNet.ImageBuilder/src/RegistryClientFactory.cs
mthalman: > Is the `ownedAcr` intended to be passed in as a parameter?

Yes. From my PR description:
> For now, this uses a hardcoded registry name to determine which implementation it should use. This is temporary.

src/Microsoft.DotNet.ImageBuilder/src/RegistryClientFactory.cs
mthalman: No issue, but it's not necessary to create it at this point. I've moved it down to where it's needed.

mthalman:
I was able to revert all the changes which flowed `RegistryAuthContext` all over the place and replaced it with a DI solution: https://github.com/dotnet/docker-tools/pull/1262/commits/02e1e8c3ed34f200488474dade29bc90491862ad.

lbussell (approved):

mthalman:
I'm going to hold off on merging this until https://github.com/dotnet/docker-tools/pull/1260 is merged and rolled out.

PR #1263, opened 2024-04-30
Fix branch validation

### LINKED ISSUE #1247: `Validate branch` step isn't validating correctly
The intent of the [`Validate branch` step](https://github.com/dotnet/docker-tools/blob/3d4a494c5e7a9a741b7359f220ef70a779ce1dc7/eng/common/templates/steps/validate-branch.yml#L5-L14) is to prevent publishing from occurring from unofficial branches. However, it's not preventing anything. It's check for an official branch is always succeeding, allowing publishing to occur from any branch.

It seems to be caused by this condition: https://github.com/dotnet/docker-tools/blob/3d4a494c5e7a9a741b7359f220ef70a779ce1dc7/eng/common/templates/steps/validate-branch.yml#L8

That check was added as part of https://github.com/dotnet/docker-tools/pull/1044. Not exactly sure what the intent of it is (yes, I know I authored it).

In addition, as a safety precaution, it might be better to inverse the logic so that it fails the build by default and only allows it if the condition is met.

lbussell:
Fixes #1247 and applies the suggestion that we should fail by default and only pass if specific conditions are met (inverted the existing conditions).

- ~~Still needs to have an additional passing condition for publish prefixes that aren't in the `officialRepoPrefixes`~~

### DIFF
--- a/eng/common/templates/steps/validate-branch.yml
+++ b/eng/common/templates/steps/validate-branch.yml
@@ -4,11 +4,19 @@ parameters:
 steps:
 - ${{ if and(eq(variables['System.TeamProject'], parameters.internalProjectName), ne(variables['Build.Reason'], 'PullRequest')) }}:
   - powershell: |
-      if (-not "$(officialBranches)".Split(',').Contains("$(sourceBranch)") `
-          -and "$(officialBranches)".Split(',').Contains("$(publishRepoPrefix)") `
-          -and "$(overrideOfficialBranchValidation)" -ne "true")
+      if ("$(officialBranches)".Split(',').Contains("$(sourceBranch)") `
+          -and "$(officialRepoPrefixes)".Split(',').Contains("$(publishRepoPrefix)"))
       {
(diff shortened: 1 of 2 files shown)


eng/common/templates/steps/validate-branch.yml
```diff
@@ -4,11 +4,25 @@ parameters:
...
+        exit 0
+      }
+
+      if ("$(overrideOfficialBranchValidation)" -eq "true")
+      {
+        echo "Variable overrideOfficialBranchValidation is set to true, continuing..."
+        exit 0
+      }
+
+      echo "##vso[task.logissue type=error]Official builds must be done from an official branch: $(officialBranches)"
```
mthalman, line 26:
```suggestion
      echo "##vso[task.logissue type=error]Official builds must be done from an official branch ($(officialBranches)) and repo prefix ($(officialRepoPrefixes))."
```

mthalman (approved):

PR #1266, opened 2024-05-02
Pass internalProjectName to copy-base-images from base image update pipeline

lbussell:
An issue with the base image update pipeline was introduced in https://github.com/dotnet/docker-tools/pull/1261. In order to run with authentication, the copy-base-images step needs to know if it's in the internal project or not, but the parameter was not passed in.


lbussell:
This change was tested internally and is working. I'm merging this now since this is needed to maintain our SLA for base image updates. Any review comments can be addressed in a follow-up PR.

Issue #1264, opened 2024-04-30
Update ImageBuilder commands that use Service Principal authentication


lbussell:
Commands that need to be updated:
- [x] [CopyImages](https://github.com/dotnet/docker-tools/blob/10c82d12cf90e24693067e9768eab929153e1fa2/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyImagesOptions.cs#L31) - @lbussell 
  - [x] https://github.com/dotnet/docker-tools/pull/1261
  - [x] https://github.com/dotnet/docker-tools/pull/1260
  - [x] https://github.com/dotnet/dotnet-docker/pull/5421
- [x] [CleanAcrImages](https://github.com/dotnet/docker-tools/blob/99579deabbf928327291d8cd4271aa486bd34660/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs#L39) - @mthalman 
    - [x] https://github.com/dotnet/docker-tools/pull/1272
    - [x] https://github.com/dotnet/docker-tools/pull/1271
- [x] [WaitForMcrDocIngestion](https://github.com/dotnet/docker-tools/blob/99579deabbf928327291d8cd4271aa486bd34660/src/Microsoft.DotNet.ImageBuilder/src/Commands/WaitForMcrDocIngestionOptions.cs#L46) and [WaitForMcrImageIngestionOptions](https://github.com/dotnet/docker-tools/blob/99579deabbf928327291d8cd4271aa486bd34660/src/Microsoft.DotNet.ImageBuilder/src/Commands/WaitForMcrImageIngestionOptions.cs#L44) - @lbussell 
    - [x] https://github.com/dotnet/docker-tools/pull/1269
    - [x] https://github.com/dotnet/docker-tools/pull/1271
- [x] [Build](https://github.com/dotnet/docker-tools/blob/99579deabbf928327291d8cd4271aa486bd34660/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildOptions.cs#L38) - @lbussell 
    - [x] https://github.com/dotnet/docker-tools/pull/1293
- [x] [IngestKustoImageInfo](https://github.com/dotnet/docker-tools/blob/99579deabbf928327291d8cd4271aa486bd34660/src/Microsoft.DotNet.ImageBuilder/src/Commands/IngestKustoImageInfoOptions.cs#L46) - @mthalman
    - [x] https://github.com/dotnet/docker-tools/pull/1277
[...]

PR #1269, opened 2024-05-03
Remove service principal authentication from McrStatusClient

lbussell:
Part of #1264

I haven't tested this in-place yet since it only runs on real builds, so marking as a draft for now. Early next week I'll run a test by modifying a pipeline to skip publishing and just check the status of some existing build.


lbussell:
Tried to queue a test but I'm not sure it did anything [internal link] - https://dev.azure.com/dnceng/internal/_build/results?buildId=2444600&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=487ef77f-c8ee-5b2f-147b-5cff06a12f4e

Any other tips for testing this @mthalman? Or since it wouldn't block publishing should we just merge it and revert if there are any big issues?

mthalman:
> Tried to queue a test but I'm not sure it did anything [internal link] - https://dev.azure.com/dnceng/internal/_build/results?buildId=2444600&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=487ef77f-c8ee-5b2f-147b-5cff06a12f4e
> 
> Any other tips for testing this @mthalman? Or since it wouldn't block publishing should we just merge it and revert if there are any big issues?

It didn't do anything because no images were published (due to image caching). You could test this by running Image Builder locally targeting the image info file from a previous publish run.

src/Microsoft.DotNet.ImageBuilder/src/AuthHelper.cs
```diff
@@ -16,5 +18,12 @@ public static async Task<string> GetAadAccessTokenAsync(string resource, string
                 resource, new ClientCredential(username, password));
             return result.AccessToken;
         }
+
+        public static async Task<string> GetMsiAadAccessTokenAsync(string resource)
+        {
+            ManagedIdentityCredential credential  = new ManagedIdentityCredential();
```
mthalman, line 24: Consider changing this to `DefaultAzureCredential`, if that works. That will allow testing of this command to still work on a dev machine.

lbussell: It's working with default. I'll update it.

mthalman (approved):

PR #1275, opened 2024-05-08
Perf improvements for CleanAcrImagesCommand

mthalman:
The `CleanAcrImagesCommand` command makes use of `IAsyncEnumerable` but it does so in a blocking manner. This causes poor performance in the build job that consumes that command. In one scenario it's taking 23 minutes to run when the previous implementation (prior to https://github.com/dotnet/docker-tools/pull/1272) was taking just 3 minutes.


src/Microsoft.DotNet.ImageBuilder/src/Microsoft.DotNet.ImageBuilder.csproj
```diff
@@ -31,6 +31,7 @@
     <PackageReference Include="System.ComponentModel.Composition" Version="8.0.0-rc.2.23479.6" />
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="8.0.0-rc.2.23479.6" />
     <PackageReference Include="System.Data.SqlClient" Version="4.8.6" />
+    <PackageReference Include="System.Linq.Async" Version="6.0.1" />
```
lbussell, line 34: Cool, it's a .NET Foundation project 👍

lbussell (approved):

lbussell:
Merging in latest now to start PR validation.

PR #1280, opened 2024-05-10
Fix casing for tasks of publish notification

mthalman:
The `Post Publish Notification` task fails with this error:

```
Unhandled exception: System.InvalidOperationException: Build task with name 'Wait For Image Ingestion' could not be found in the build timeline.
```

Fixed by using the correct casing that matches the actual task.


lbussell (approved):
Thanks for fixing this while I'm busy with servicing!

Issue #1286, opened 2024-05-15
Wait for image ingestion step failing due to having multiple user assigned identities


lbussell:
```
- ManagedIdentityCredential authentication unavailable. The requested identity has not been assigned to this resource.
Status: 400 (Bad Request)

Content:
{"error":"invalid_request","error_description":"Multiple user assigned identities exist, please specify the clientId / resourceId of the identity in the token request"}
```

We don't control which identities are attached to the VMs we get. We should consider using an identity with no special permissions to access this API.

lbussell:
I have created a new identity and service connection, and am working out how to use it to work around this issue.

PR #1287, opened 2024-05-15
Use service connection for MAR ingestion and use .default scope

lbussell:
This should fix https://github.com/dotnet/docker-tools/issues/1286 once the new ImageBuilder is built and flows in.

Using the service connection by itself yielded a new error:

```
 ---> Azure.Identity.AuthenticationFailedException: ClientAssertionCredential authentication failed: AADSTS1002012: The provided value for scope <snip> is not valid. Client credential flows must have a scope value with /.default suffixed to the resource identifier (application ID URI).
```

So that's why the ImageBuilder change is necessary. I will file a follow-up issue to use a pipeline variable for the status API scope since hardcoding it isn't great.

### DIFF
--- a/eng/common/templates/steps/wait-for-mcr-image-ingestion.yml
+++ b/eng/common/templates/steps/wait-for-mcr-image-ingestion.yml
@@ -16,6 +16,7 @@ steps:
       '${{ parameters.imageInfoPath }}'
       --manifest '$(manifest)'
       --repo-prefix '$(publishRepoPrefix)'
+      --min-queue-time '${{ parameters.minQueueTime }}'
       --timeout '$(mcrImageIngestionTimeout)'
       $(manifestVariables)
       ${{ parameters.dryRunArg }}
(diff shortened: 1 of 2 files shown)


eng/common/templates/steps/wait-for-mcr-image-ingestion.yml
```diff
@@ -9,12 +9,13 @@ steps:
...
     displayName: Wait for Image Ingestion
     condition: and(${{ parameters.condition }}, eq(variables['waitForIngestionEnabled'], 'true'))
+    serviceConnection: $(marStatus.serviceConnectionName)
+    internalProjectName: 'internal'
     args: >
       waitForMcrImageIngestion
       '${{ parameters.imageInfoPath }}'
       --manifest '$(manifest)'
       --repo-prefix '$(publishRepoPrefix)'
-      --min-queue-time '${{ parameters.minQueueTime }}'
```
mthalman, line 17: Why is this removed?

lbussell:
Filed https://github.com/dotnet/docker-tools/issues/1288

eng/common/templates/steps/wait-for-mcr-image-ingestion.yml
lbussell: For testing. I will revert this change.

mthalman (approved):

PR #1290, opened 2024-05-16
Use ACR service connection to login to ACR for test job

mthalman:
This removes the use of credentials for logging into the ACR in test jobs and replaces it with the use of a service connection.

This requires the changes in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1057.

Contributes to https://github.com/dotnet/docker-tools/issues/1264

### DIFF
--- a/eng/common/templates/steps/run-pwsh-with-auth.yml
+++ b/eng/common/templates/steps/run-pwsh-with-auth.yml
@@ -4,7 +4,7 @@ parameters:
   default: "RunPwshWithAuth"
 - name: displayName
   type: string
-  default: "Run PowerShell with Auth"
+  default: "Run PowerShell"
 - name: serviceConnection
   type: string
   default: ""
@@ -24,7 +24,7 @@ parameters:
 steps:
 - task: AzureCLI@2
   name: ${{ parameters.name }}
-  displayName: ${{ parameters.displayName }}
+  displayName: ${{ parameters.displayName }} (Authenticated)
   continueOnError: ${{ parameters.continueOnError }}
   condition: and(succeeded(), ${{ parameters.condition }})
   inputs:
--- a/eng/common/templates/steps/run-pwsh-with-auth.yml
+++ b/eng/common/templates/steps/run-pwsh-with-auth.yml
@@ -1,0 +1,38 @@
+parameters:
+- name: name
+  type: string
+  default: "RunPwshWithAuth"
+- name: displayName
+  type: string
+  default: "Run PowerShell with Auth"
+- name: serviceConnection
+  type: string
+  default: ""
+- name: command
+  type: string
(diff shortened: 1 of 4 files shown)


eng/common/templates/steps/run-pwsh-with-auth.yml
```diff
@@ -0,0 +1,38 @@
...
+  type: string
+  default: "linux"
+- name: condition
+  type: string
+  default: true
+
+steps:
+- task: AzureCLI@2
+  name: ${{ parameters.name }}
+  displayName: ${{ parameters.displayName }}
```
lbussell, line 27: Just an idea - maybe we should append something like "(Authenticated)" to the displayName here to make it easier to tell which pipeline task was chosen. Would require change to default displayName above.

```suggestion
  displayName: ${{ parameters.displayName }} (Authenticated)
```

eng/common/templates/steps/test-images-windows-client.yml
```diff
@@ -10,11 +10,16 @@ steps:
...
+  - template: /eng/common/templates/steps/run-pwsh-with-auth.yml@self
+    parameters:
+      displayName: Docker login
+      serviceConnection: $(acr.serviceConnectionName)
+      dockerClientOS: windows
+      condition: and(succeeded(), ${{ parameters.condition }})
+      command: >
+        az login --service-principal --tenant $env:tenantId -u $env:servicePrincipalId --federated-token $env:idToken;
+        $accessToken = $(az acr login -n $(acr.server) --expose-token --query accessToken --output tsv);
+        docker login $(acr.server) -u 00000000-0000-0000-0000-000000000000 -p $accessToken
```
lbussell, line 22: General comment: I read on [the documentation](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#individual-login-with-microsoft-entra-id) that this token should last for 3 hours. Our tests shouldn't last for 3 hours so I don't think that's a concern.

lbussell (approved):

eng/common/templates/steps/run-pwsh-with-auth.yml
mthalman: Wouldn't internal runs always be authenticated and public runs would not?

mthalman: > Wouldn't internal runs always be authenticated and public runs would not?

I guess that's not true. There would be scenarios for internal runs where no service connection is in scope.

lbussell: I was just thinking about from a debugging point of view, it'd require less knowledge of how the tasks are structured to see if it's authenticating when you want it to.

PR #1302, opened 2024-05-22
Remove ServicePrincipalOptions

lbussell:
This class is unnecessary since #1264 is completed.


mthalman (approved):

lbussell:
Would be good to merge https://github.com/dotnet/docker-tools/pull/1306 first 🙂

PR #1303, opened 2024-05-22
Add Registry credential options back to GetStaleImages command

### LINKED ISSUE #1299: RegistryCredentialsOptions was mistakenly removed from GetStaleImagesCommand
...in https://github.com/dotnet/docker-tools/pull/1293.

We need the RegistryCredentialsOptions on GetStaleImagesCommand so that we can pass in our DockerHub credentials to avoid rate-limiting. The option was mistakenly removed in the above linked PR.

#1298 can be reverted after this is fixed.

lbussell:
Fixes #1299

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesOptions.cs
@@ -22,7 +22,6 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         public RegistryCredentialsOptions CredentialsOptions { get; set; } = new RegistryCredentialsOptions();
 
         public BaseImageOverrideOptions BaseImageOverrideOptions { get; set; } = new();
-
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesOptions.cs
```diff
@@ -20,34 +19,40 @@ public class GetStaleImagesOptions : Options, IFilterableOptions, IGitOptionsHos
 
         public string VariableName { get; set; } = string.Empty;
 
+        public RegistryCredentialsOptions CredentialsOptions { get; set; } = new RegistryCredentialsOptions();
+
         public BaseImageOverrideOptions BaseImageOverrideOptions { get; set; } = new();
+
     }
```
mthalman, line 26:
```suggestion
        public BaseImageOverrideOptions BaseImageOverrideOptions { get; set; } = new();
    }
```

PR #1320, opened 2024-05-30
Make better use of layer caching in Dockerfile

mthalman:
When iteratively building the Dockerfile after making changes to the Image Builder code, it causes the ORAS tool to be downloaded each time. This is unnecessary and a waste of time. To improve this experience, I've moved that logic up in the Dockerfile, before Image Builder gets built, so that it can take advantage of the layer caching.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
+++ b/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
@@ -34,11 +34,11 @@ RUN apk add --no-cache \
         docker \
         git
 
+# install oras tool
+COPY --from=build-env ["/oras-install/oras", "/usr/local/bin"]
+
 # install image-builder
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -5,6 +5,15 @@
 # build Microsoft.DotNet.ImageBuilder
 FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-alpine AS build-env
 ARG TARGETARCH
+
+# download oras package tarball
```
MichaelSimons, line 9: Would it make sense to make the same change in the final stage as well?

MichaelSimons (approved):

Issue #1316, opened 2024-05-29
Implement proper solution for updating ACR access token


mthalman:
The changes made in https://github.com/dotnet/docker-tools/pull/1310 and https://github.com/dotnet/docker-tools/pull/1314 were temporary workarounds in order to get a valid token for connecting to the ACR. These changes should be updated properly so that it's not a hack.

PR #1310, opened 2024-05-24
Refresh login token before publish

mthalman:
The Docker login token is expiring after a certain period for builds that take a long time (e.g. 3 hrs). This causes the `docker push` operation to fail with an `unauthorized` error.

Resolving this by refreshing the token by logging in again right before the push.

PR #1321, opened 2024-05-30
Improve Azure token logic

mthalman:
Fixes https://github.com/dotnet/docker-tools/issues/1316

The original changes from https://github.com/dotnet/docker-tools/pull/1310 and https://github.com/dotnet/docker-tools/pull/1314 were hacks in order to deal with token expiration that occurs with a long-running command. If execution of the command takes longer than an hour and then an attempt is made to retrieve an Azure access token, it will fail. This is resolved by retrieving the token at the beginning of the command instead.

This defines a provider component that returns an Azure credential that's configured with a pre-cached access token. The access tokens are retrieved during the startup of the command during component composition.

In order to get the tokens, we have to know the scope ahead of time. So these are explicitly defined and a token is retrieved for each one of them. When the caller needs the credential, it specifies which scope is needed and the appropriate credential for that scope is returned.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/AuthHelper.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/AuthHelper.cs
@@ -5,24 +5,23 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading.Tasks;
 using Azure.Core;
-using Azure.Identity;
 using Azure.ResourceManager;
 
 namespace Microsoft.DotNet.ImageBuilder
 {
     public static class AuthHelper
     {
-        private const string DefaultScope = "https://management.azure.com/.default";
+        public const string DefaultAzureManagementScope = "https://management.azure.com/.default";
+        public const string ContainerRegistryScope = "https://containerregistry.azure.net/.default";
+        public const string McrStatusScope = "api://c00053c3-a979-4ee6-b94e-941881e62d8e/.default";
 
-        public static async Task<(string token, Guid tenantId)> GetDefaultAccessTokenAsync(ILoggerService loggerService, string resource = DefaultScope)
-        {
-            DefaultAzureCredential credential  = new();
-            AccessToken token = await credential.GetTokenAsync(new TokenRequestContext([ resource ]));
-            Guid tenantId = GetTenantId(loggerService, credential);
-            return (token.Token, tenantId);
-        }
+        public static readonly string[] AllScopes =
+            [
+                DefaultAzureManagementScope,
+                ContainerRegistryScope,
+                McrStatusScope
+            ];
 
         public static Guid GetTenantId(ILoggerService loggerService, TokenCredential credential)
         {
--- a/src/Microsoft.DotNet.ImageBuilder/src/AuthHelper.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/AuthHelper.cs
@@ -12,17 +12,6 @@ namespace Microsoft.DotNet.ImageBuilder
 {
     public static class AuthHelper
     {
-        public const string DefaultAzureManagementScope = "https://management.azure.com/.default";
-        public const string ContainerRegistryScope = "https://containerregistry.azure.net/.default";
-        public const string McrStatusScope = "api://c00053c3-a979-4ee6-b94e-941881e62d8e/.default";
-
-        public static readonly string[] AllScopes =
-            [
-                DefaultAzureManagementScope,
-                ContainerRegistryScope,
-                McrStatusScope
-            ];
-
         public static Guid GetTenantId(ILoggerService loggerService, TokenCredential credential)
         {
             ArmClient armClient = new(credential);
--- a/src/Microsoft.DotNet.ImageBuilder/src/AzureScopes.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/AzureScopes.cs
@@ -1,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.ImageBuilder;
+
+#nullable enable
+internal static class AzureScopes
+{
+    public const string ScopeSuffix = "/.default";
+    public const string DefaultAzureManagementScope = "https://management.azure.com" + ScopeSuffix;
+    public const string ContainerRegistryScope = "https://containerregistry.azure.net" + ScopeSuffix;
+    public const string McrStatusScope = "api://c00053c3-a979-4ee6-b94e-941881e62d8e" + ScopeSuffix;
+    public static readonly string[] AllScopes =
+        [
+            DefaultAzureManagementScope,
+            ContainerRegistryScope,
+            McrStatusScope
+        ];
+}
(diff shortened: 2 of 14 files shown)


src/Microsoft.DotNet.ImageBuilder/src/AzureTokenCredentialProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+                if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("AZURE_FEDERATED_TOKEN_FILE")))
+                {
+                    loggerService.WriteMessage($"Getting token for scope '{scope}'");
+                    AccessToken token = new WorkloadIdentityCredential().GetToken(new TokenRequestContext([scope]), CancellationToken.None);
+                    credential = new StaticTokenCredential(token);
+                }
+                else
+                {
+                    loggerService.WriteMessage($"No credentials set for scope '{scope}'");
+                }
```
lbussell, line 52: In the case where I'm running a local test using ImageBuilder in a container, does this make all commands that might require auth fail? Is that a valid scenario to consider anymore?

src/Microsoft.DotNet.ImageBuilder/src/AzureTokenCredentialProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+
+    [ImportingConstructor]
+    public AzureTokenCredentialProvider(ILoggerService loggerService)
+    {
+        loggerService.WriteSubheading("Initializing Azure credentials");
+
+        // Pre-cache the credentials for all the scopes we need. This is done at startup to ensure we have a valid OIDC token from the pipeline before it expires.
+        // Otherwise, long-running commands that don't attempt to request an Azure access token until much later run the risk of failing due to the expired OIDC token.
+        _credentialsByScope = AuthHelper.AllScopes
+            .ToDictionary<string, string, TokenCredential?>(scope => scope, scope =>
```
lbussell, line 29: How does this behave when we try to cache a token when the Workload Identity we're using doesn't give any access to a specific scope? For example, what happens when we try to cache the ACR token when running the MAR Status API commands?

src/Microsoft.DotNet.ImageBuilder/src/CopyImageService.cs
```diff
@@ -34,13 +33,13 @@ public interface ICopyImageService
 public class CopyImageService : ICopyImageService
 {
     private readonly ILoggerService _loggerService;
-
-    private readonly ArmClient _armClient = new(new DefaultAzureCredential());
+    private readonly Lazy<ArmClient> _armClient;
```
lbussell, line 36: Why is this `Lazy`? Can it be created in the constructor?

src/Microsoft.DotNet.ImageBuilder/src/AuthHelper.cs
```diff
@@ -5,24 +5,23 @@
...
-            AccessToken token = await credential.GetTokenAsync(new TokenRequestContext([ resource ]));
-            Guid tenantId = GetTenantId(loggerService, credential);
-            return (token.Token, tenantId);
-        }
+        public static readonly string[] AllScopes =
+            [
+                DefaultAzureManagementScope,
+                ContainerRegistryScope,
+                McrStatusScope
+            ];
```
lbussell, line 24: Since this is only used by `AzureTokenCredentialProvider`, should it live in that class?

src/Microsoft.DotNet.ImageBuilder/src/Services/KustoClientWrapper.cs
```diff
@@ -33,6 +33,9 @@ public KustoClientWrapper(ILoggerService loggerService)
         {
             KustoConnectionStringBuilder connectionBuilder =
                 new KustoConnectionStringBuilder($"https://{cluster}.kusto.windows.net")
+                    // We can't use AzureTokenCredentialProvider here to get the credential because that would require
+                    // knowing the scope at startup which is based on the cluster name. It's fine to not use a pre-cached
+                    // token anyway since kusto ingestion happens right away in the command invocation.
                     .WithAadAzureTokenCredentialsAuthentication(new DefaultAzureCredential());
```
lbussell, line 39: If `AzureTokenCredentialProvider` could be given the scope in its constructor, and it was lazily constructed, would that work for this scenario? It smells to me that we may be creating a solution that only works for 2/3 scenarios.

lbussell (commented):
I think having the `AzureTokenCredentialProvider` as a token cache is a great idea.

General comments - you can take or leave these ideas :)

I think all of the commands already use their auth right away - the only reason we had failures in the `BuildCommand` is because `ExecuteWithCredentialsAsync(...)` and the `ManifestService` were getting their own credentials, right? (one at the start, and the other after the build finished). If `AzureTokenCredentialProvider` were just a general cache for token(s) instead of trying to predict which tokens we may need, I think it might still solve the problem while being simpler. *When* we decide to get the tokens from the `AzureTokenCredentialProvider` at any point in a command could be a separate consideration. As long as the build command uses its auth right away and the token is cached, we wouldn't have to do anything fancy there.

src/Microsoft.DotNet.ImageBuilder/src/AzureTokenCredentialProvider.cs
mthalman: That scenario wouldn't work even with our current implementation. It can't use your dev configuration because that doesn't exist in the container.

src/Microsoft.DotNet.ImageBuilder/src/CopyImageService.cs
mthalman: Attempting to get a token during component composition results in the following exception in the dependency injection unit test because it doesn't run in an environment setup for auth:

```
System.ComponentModel.Composition.CompositionException : The composition produced a single composition error. The root cause is provided below. Review the CompositionException.Errors property for more detailed information.



EnvironmentCredential authentication unavailable. Environment variables are not fully configured. See the troubleshooting guide for more information. https://aka.ms/azsdk/net/identity/environmentcredential/troubleshoot


Resulting in: Multiple exceptions were encountered while attempting to authenticate. (EnvironmentCredential authentication unavailable. Environment variables are not fully configured. See the troubleshooting guide for more information. https://aka.ms/azsdk/net/identity/environmentcredential/troubleshoot) (WorkloadIdentityCredential authentication unavailable. The workload options are not fully configured. See the troubleshooting guide for more information. https://aka.ms/azsdk/net/identity/workloadidentitycredential/troubleshoot) (ManagedIdentityCredential authentication unavailable. The requested identity has not been assigned to this resource.
Status: 400 (Bad Request)
```

src/Microsoft.DotNet.ImageBuilder/src/AzureScopes.cs
```diff
@@ -0,0 +1,19 @@
...
+    public const string ScopeSuffix = "/.default";
+    public const string DefaultAzureManagementScope = "https://management.azure.com" + ScopeSuffix;
+    public const string ContainerRegistryScope = "https://containerregistry.azure.net" + ScopeSuffix;
+    public const string McrStatusScope = "api://c00053c3-a979-4ee6-b94e-941881e62d8e" + ScopeSuffix;
+    public static readonly string[] AllScopes =
+        [
+            DefaultAzureManagementScope,
+            ContainerRegistryScope,
+            McrStatusScope
+        ];
```
lbussell, line 18: This appers to be unused.

lbussell (approved):

lbussell (approved):

lbussell (approved):

PR #1333, opened 2024-06-13
Add pipeline for mirroring base images to cache ACR on a schedule

lbussell:
1/2 of https://github.com/dotnet/dotnet-docker-internal/issues/5504

Implements a new subscription type (manifest-only subscription) and pipeline that consumes that subscription. The pipeline runs an import of all the base images into our cache ACR. Some minor refactoring of the CopyBaseImages pipeline templates was required.

### DIFF
--- a/eng/common/templates/jobs/copy-base-images-staging.yml
+++ b/eng/common/templates/jobs/copy-base-images-staging.yml
@@ -1,0 +1,30 @@
+parameters:
+- name: name
+  type: string
+  default: null
+- name: pool
+  type: object
+  default: {}
+- name: customInitSteps
+  type: stepList
+  default: []
+- name: additionalOptions
+  type: string
+  default: ''
+- name: continueOnError
+  type: string
+  default: false
+
+jobs:
+- template: /eng/common/templates/jobs/copy-base-images.yml@self
+  parameters:
+    name: ${{ parameters.name }}
+    pool: ${{ parameters.pool }}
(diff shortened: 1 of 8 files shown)


eng/common/templates/variables/docker-images.yml
```diff
@@ -1,5 +1,5 @@
 variables:
-  imageNames.imageBuilderName: mcr.microsoft.com/dotnet-buildtools/image-builder:2472089
+  imageNames.imageBuilderName: dotnetdockerdev.azurecr.io/dotnet-buildtools/image-builder:mirrorbaseimages
```
mthalman, line 2: Needs to be reverted.

eng/mirror-base-images-subscriptions.json
mthalman, line 1: Is there a reason we can't use the existing subscription files? I don't really like the idea of having duplicate content here.

lbussell: I chose to do this so that we could have the same file/job for both dotnet-docker and buildtools prereqs, and so I could exclude the framework repo. Although it may be fine to use the existing files, since filtering to external from images only will make framework a no-op.

lbussell: That'll vastly simplify this, though... I'll work on some changes

mthalman (approved):

mthalman (approved):

Issue #1201, opened 2024-01-09
Add lifecycle annotations to images pushed to MCR


MichaelSimons:
A broader set of services are starting to consume image annotations.  Specifically Azure container scanning honors the lifecycle annotation to treat EOL images specially.  As a result, the images produced with the .NET Docker infrastructure should have these lifecycle annotations added.

Related links:
* [OCI Annotations Spec](https://specs.opencontainers.org/image-spec/annotations/#:~:text=Annotations%20MUST%20be%20a%20key,reverse%20domain%20notation%20%2D%20e.g.%20com)
* [Thread & Vulnerability Management guidance](https://eng.ms/docs/microsoft-security/cloud-ecosystem-security/azure-security/security-health-analytics/sha-intelligence/threat-vulnerability-management/remediate/container) (MSFT internal link)

  ```
  How can I mark an image in my ACR as end of life or dictate how long it should be scanned and reported on?
  Attach annotations* to the image called 'lifecycle annotations' which requires:
  
  The ACR Metadata Service to be onboarded for your registry (see callout below)
  [ORAS CLI](https://oras.land/docs/installation/) must be installed
  ORAS CLI must be authenticated to your registry [(See How to Authenticate ORAS CLI)](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-oci-artifacts#sign-in-to-a-registry)
  You can use the ORAS CLI to attach lifecycle annotations by running the following command:
  
  Variables:
  
  $IMAGE refers to a fully qualified image name or digest (e.g. myregistry.azurecr.io/myimage:v1 OR myregistry.azurecr.io/myimage@sha256:fdbd1bc674f77e7a3b8129ecac062a9b7c22de74968e8ea3a46f071f61f3e99f)
  Lifecycle.end-of-life.date should be in the expected format of yyyy-mm-dd
  oras attach --artifact-type application/vnd.microsoft.artifact.lifecycle --annotation "vnd.microsoft.artifact.lifecycle.end-of-life.date=2023-05-12T21:15:10.8343521Z" $IMAGE
  ```

Work Items:

- [x] https://github.com/dotnet/docker-tools/pull/1334
- [x] https://github.com/dotnet/docker-tools/pull/1340
- [x] https://github.com/dotnet/docker-tools/pull/1358
- [x] https://github.com/dotnet/docker-tools/pull/1364
- [x] https://github.com/dotnet/docker-tools/pull/1367
[...]

PR #1334, opened 2024-06-13
Add a new command for annotating image digests for EOL

NikolaMilosavljevic:
Contributes to: https://github.com/dotnet/docker-tools/issues/1201

The work is being split into smaller PRs to aid in review process.

Some notes:
- We allow two scenarios for annotations - with and without checking for existing annotations.
- Check for existing annotations relies on `oras discover` command and its stdout. This will likely be updated once Oras completed the work on managed library.
- `oras login` command was not needed, as we already do `docker login`
- We do not fail if we fail to annotate one digest, we proceed until the end and log all failures, at which point we fail the command.
- There are 3 tests right now. We could add more later.

JSON file format (with sample dates):
```
{
  "eolDate": "2024-05-27",
  "eolDigests": [
    {
      "digest": "digest-value"
    },
    {
      "digest": "digest-value",
      "eolDate" : "2022-03-01"
    }
  ]
}
```

EOL data can be specified as a global value (optional) for all digests in the file. Each digest can have its own specific EOL date, which overrides the global value (if set).


NikolaMilosavljevic:
Hmm, some test failures on Linux. This worked fine on Windows.

Working on a fix.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    public class AnnotateEolDigestsOptionsBuilder : DockerRegistryOptionsBuilder
+    {
+        public override IEnumerable<Option> GetCliOptions() =>
+            base.GetCliOptions()
+                .Concat(
+                    new Option[]
+                    {
+                        CreateOption<bool>("no-check", nameof(AnnotateEolDigestsOptions.NoCheck),
+                            "Annotate always, without checking if digest is already annotated for EOL"),
```
mthalman, line 34: Consider using `force` as the name of the option instead. It's more commonly used.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+                    }
+                );
+
+        public override IEnumerable<Argument> GetCliArguments() =>
+            base.GetCliArguments()
+                .Concat(
+                    new Argument[]
+                    {
+                        new Argument<string>(nameof(AnnotateEolDigestsOptions.EolDigestsListPath),
+                            "Eol annotations digests list path")
```
mthalman, line 48:
```suggestion
                            "EOL annotations digests list path")
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            string eolAnnotationsJson = File.ReadAllText(eolDigestsListPath);
+            EolAnnotationsData? eolAnnotations = JsonConvert.DeserializeObject<EolAnnotationsData>(eolAnnotationsJson);
+            return eolAnnotations is null
+                ? throw new JsonException($"Unable to correctly deserialize path '{eolAnnotationsJson}'.")
+                : eolAnnotations;
+        }
+    }
+}
+#nullable restore
```
mthalman, line 100: I learned from @lbussell  that these `#nullable` directives at the end of the file aren't needed. The scope of `#nullable enable` only applies to the file.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+    public class AnnotateEolDigestsOptions : DockerRegistryOptions
+    {
+        public string EolDigestsListPath { get; set; } = string.Empty;
+        public string? Subscription { get; set; }
+        public string? ResourceGroup { get; set; }
+        public bool NoCheck { get; set; } = false;
+
+        public AnnotateEolDigestsOptions() : base()
+        {
+        }
```
mthalman, line 23: Unnecessary

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/EolDigestData.cs
```diff
@@ -0,0 +1,17 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.ImageBuilder.Models.EolAnnotations
+{
+    public class EolDigestData
+    {
+        [JsonProperty(Required = Required.Always)]
+        public string Digest { get; set; }
+
+        public DateOnly? EolDate { get; set; }
```
mthalman, line 15: This should be a `DateTime`.

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/EolAnnotationsData.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+        }
+
+        public EolAnnotationsData(DateOnly eolDate, List<EolDigestData> eolDigests)
+        {
+            EolDate = eolDate;
+            EolDigests = eolDigests;
+        }
+
+        [JsonProperty(Required = Required.Always)]
```
mthalman, line 24: I question whether this should be required. If I'm hand editing the file with two digests that have differing dates, I'd prefer to just set the EOL date for both individual digests rather than make one the default.

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class OrasService : IOrasService
+    {
+        public bool IsDigestAnnotatedForEol(string digest, bool isDryRun)
+        {
+            string? stdOut = ExecuteHelper.ExecuteWithRetry(
+                "oras",
+                $"discover --artifact-type application/vnd.microsoft.artifact.lifecycle {digest}",
+                isDryRun);
+
+            if (!string.IsNullOrEmpty(stdOut) && stdOut.Contains("Discovered 0 artifact"))
```
mthalman, line 21: Does it have a status code that indicates positive/negative result instead of needing to check stdout?

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace Microsoft.DotNet.ImageBuilder
+{
+    [Export(typeof(IOrasService))]
+    public class OrasService : IOrasService
+    {
+        public bool IsDigestAnnotatedForEol(string digest, bool isDryRun)
+        {
+            string? stdOut = ExecuteHelper.ExecuteWithRetry(
+                "oras",
+                $"discover --artifact-type application/vnd.microsoft.artifact.lifecycle {digest}",
```
mthalman, line 18: Consider defining `application/vnd.microsoft.artifact.lifecycle` as a constant that is reused here and in the annotate method.

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
```diff
@@ -0,0 +1,48 @@
...
+            try
+            {
+                ExecuteHelper.ExecuteWithRetry(
+                    "oras",
+                    $"attach --artifact-type application/vnd.microsoft.artifact.lifecycle --annotation \"vnd.microsoft.artifact.lifecycle.end-of-life.date={date}\" {digest}",
+                    isDryRun);
+            }
+            catch (InvalidOperationException ex)
+            {
+                loggerService.WriteMessage($"Failed to annotate EOL for digest '{digest}': {ex.Message}");
```
mthalman, line 40:
```suggestion
                loggerService.WriteError($"Failed to annotate EOL for digest '{digest}': {ex.Message}");
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+                        }
+                    });
+
+                },
+                registryName: Manifest.Registry,
+                ownedAcr: Options.RegistryOverride);
+
+            if (_failedAnnotations.Count > 0)
+            {
+                _loggerService.WriteMessage("JSon file for rerunning failed annotations:");
```
mthalman, line 77:
```suggestion
                _loggerService.WriteMessage("JSON file for rerunning failed annotations:");
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+                _loggerService.WriteMessage("");
+                _loggerService.WriteMessage(JsonConvert.SerializeObject(new EolAnnotationsData(globalEolDate, [.. _failedAnnotations])));
+                _loggerService.WriteMessage("");
+                throw new InvalidOperationException($"Failed to annotate {_failedAnnotations.Count} digests for EOL.");
+            }
+        }
+
+        private static EolAnnotationsData LoadEolAnnotationsData(string eolDigestsListPath)
+        {
+            if (eolDigestsListPath == null)
```
mthalman, line 87: This is unnecessary since nullable checks are enabled and the fact that it's a required argument in the command.

src/Microsoft.DotNet.ImageBuilder/tests/AnnotateEolDigestsCommandTests.cs
```diff
@@ -0,0 +1,172 @@
...
+                    tempFolderContext,
+                    out loggerServiceMock,
+                    noCheck: true,
+                    digestAlreadyAnnotated: true,
+                    digestAnnotationIsSuccessful: true);
+            command.LoadManifest();
+            await command.ExecuteAsync();
+
+            loggerServiceMock.Verify(o => o.WriteMessage($"Annotating EOL for digest 'digest1', date '{_globalDate}'"));
+            loggerServiceMock.Verify(o => o.WriteMessage($"Annotating EOL for digest 'digest2', date '{_specificDigestDate}'"));
```
mthalman, line 49: I don't find that validating logging messages is all that useful. Instead, we should be validating that the calls to `OrasService` are being made with the expected input.

src/Microsoft.DotNet.ImageBuilder/tests/AnnotateEolDigestsCommandTests.cs
```diff
@@ -0,0 +1,172 @@
...
+                    tempFolderContext,
+                    out loggerServiceMock,
+                    noCheck: false,
+                    digestAlreadyAnnotated: true,
+                    digestAnnotationIsSuccessful: true);
+            command.LoadManifest();
+            await command.ExecuteAsync();
+
+            loggerServiceMock.Verify(o => o.WriteMessage("Digest 'digest1' is already annotated for EOL."));
+            loggerServiceMock.Verify(o => o.WriteMessage("Digest 'digest2' is already annotated for EOL."));
```
mthalman, line 92: Repeating my comment above about what to validate here. Here, we'd want to validate that `IsDigestAnnotatedForEol` is being called with the expected input and that `AnnotateEolDigest` is _not_ being called at all.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
NikolaMilosavljevic: `force` was the option I also considered. Yeah, it makes sense to use it - will fix.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
NikolaMilosavljevic: Right, forgot to remove - thanks.

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/EolAnnotationsData.cs
NikolaMilosavljevic: Yeah, good point. I'll update this and related code.

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/EolDigestData.cs
NikolaMilosavljevic: OK, that would also simplify some code. `DateOnly` did seem convenient originally so I stuck with it.

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
NikolaMilosavljevic: Unfortunately no - this case is considered a success.

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
NikolaMilosavljevic: Missed this - thanks.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
NikolaMilosavljevic: Right, not sure why I added this - will remove.

src/Microsoft.DotNet.ImageBuilder/tests/AnnotateEolDigestsCommandTests.cs
NikolaMilosavljevic: Oh, that makes more sense - thanks.

src/Microsoft.DotNet.ImageBuilder/tests/AnnotateEolDigestsCommandTests.cs
NikolaMilosavljevic: Thanks - that is better - will fix.

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
lbussell: This will behave unexpectedly if ran on an image that has any other artifacts - example:

```
PS> oras discover docker.io/bitnami/oras:latest
Discovered 1 artifact referencing latest
Digest: sha256:a50c6a13aac080ee67e6ed19e2cdad6f427cd303dd9bc37a83448c8945e2bd2f

Artifact Type                           Digest
application/vnd.cncf.notary.signature   sha256:1b77ffff8ddb39f17edd79cd64498af3099d1712aaf2076af29bc56e9aadc627
```

It's not a lifecycle digest, but a different type of artifact.

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/EolDigestData.cs
NikolaMilosavljevic: Keeping this as `DateOnly` as API requires date-only in a specific format - `yyyy-mm-dd`, according to spec: https://eng.ms/docs/microsoft-security/cloud-ecosystem-security/azure-security/security-health-analytics/sha-intelligence/threat-vulnerability-management/remediate/container#container-annotations

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
NikolaMilosavljevic: OK. I will update this code to get the full manifest json and look for elements of a specific artifact type.

NikolaMilosavljevic: Actually, no need for a change, we query the specific artifact type:

```
PS> oras discover docker.io/bitnami/oras:latest
Discovered 1 artifact referencing latest
Digest: sha256:a50c6a13aac080ee67e6ed19e2cdad6f427cd303dd9bc37a83448c8945e2bd2f

Artifact Type                           Digest
application/vnd.cncf.notary.signature   sha256:1b77ffff8ddb39f17edd79cd64498af3099d1712aaf2076af29bc56e9aadc627

PS> oras discover --artifact-type application/vnd.microsoft.artifact.lifecycle docker.io/bitnami/oras:latest
Discovered 0 artifact referencing latest
Digest: sha256:a50c6a13aac080ee67e6ed19e2cdad6f427cd303dd9bc37a83448c8945e2bd2f
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/12fe0a4610b7ad7bb860efd8b975f3f06980f23a

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/12fe0a4610b7ad7bb860efd8b975f3f06980f23a

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/12fe0a4610b7ad7bb860efd8b975f3f06980f23a

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/EolAnnotationsData.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/12fe0a4610b7ad7bb860efd8b975f3f06980f23a

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/12fe0a4610b7ad7bb860efd8b975f3f06980f23a

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/12fe0a4610b7ad7bb860efd8b975f3f06980f23a

src/Microsoft.DotNet.ImageBuilder/tests/AnnotateEolDigestsCommandTests.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/12fe0a4610b7ad7bb860efd8b975f3f06980f23a

src/Microsoft.DotNet.ImageBuilder/tests/AnnotateEolDigestsCommandTests.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/12fe0a4610b7ad7bb860efd8b975f3f06980f23a

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
```diff
@@ -26,13 +28,18 @@ public bool IsDigestAnnotatedForEol(string digest, bool isDryRun)
             return true;
         }
 
-        public bool AnnotateEolDigest(string digest, DateOnly date, ILoggerService loggerService, bool isDryRun)
+        public bool AnnotateEolDigest(string digest, DateOnly? date, ILoggerService loggerService, bool isDryRun)
         {
+            if (date == null)
+            {
+                loggerService.WriteError($"EOL date is not specified for digest '{digest}'.");
+            }
```
mthalman, line 36: This doesn't stop execution of the rest of the method. So even though it's null, it'll still run `oras`. Also, it seems weird that an API will accept a nullable value only to reject it if it's null. The caller should do this check instead.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
```diff
@@ -0,0 +1,49 @@
...
+using static Microsoft.DotNet.ImageBuilder.Commands.CliHelper;
+
+#nullable enable
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class AnnotateEolDigestsOptions : DockerRegistryOptions
+    {
+        public string EolDigestsListPath { get; set; } = string.Empty;
+        public string? Subscription { get; set; }
+        public string? ResourceGroup { get; set; }
```
mthalman, line 18: These aren't used anywhere.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
```diff
@@ -0,0 +1,49 @@
...
+using System;
+using System.Collections.Generic;
+using System.CommandLine;
+using System.Linq;
+using static Microsoft.DotNet.ImageBuilder.Commands.CliHelper;
+
+#nullable enable
+namespace Microsoft.DotNet.ImageBuilder.Commands
+{
+    public class AnnotateEolDigestsOptions : DockerRegistryOptions
```
mthalman, line 14: I just realized that `DockerRegistryOptions` derives from `ManifestOptions` which is unfortunate because it causes the need to load Image Builder's manifest which isn't actually used for the annotation logic. There's no technical need for `DockerRegistryOptions` to derive from `ManifestOptions`. This is what causes the need for you to specify these options: https://github.com/dotnet/docker-tools/pull/1340/files#diff-d39079bb88b4127d62a9063364098bbb6d0fcd6e5ea70fba9ff9661f01cde625R25-R26.

I think `DockerRegistryCommand` could be removed entirely from the hierarchy and have its implementation separated out. This method could be moved to be an extension method of `IRegistryCredentialsProvider`:
https://github.com/dotnet/docker-tools/blob/94131bc813764ccc3464b018c12e461eea3b6912/src/Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryCommand.cs#L22

And `DockerRegistryOptions` could be changed to not derive from `ManifestOptions` and instead be a standalone options class that is composed into consuming commands in the same way that `ManifestFilterOptions` works.

Could you submit a separate PR to refactor things in this way? That'll clean things up and make things more correct.

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
NikolaMilosavljevic: Hmm, not sure how this got out. In local file I was returning early.

Will move the check to caller method.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
NikolaMilosavljevic: Hmm, interesting. I had to have them originally, but I see that it now works without them. Will remove.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
NikolaMilosavljevic: OK, so this would be a follow-up PR, right? For now we would need to keep Manifest specific options.

mthalman: My suggested changes should be done first and then merged into this PR so that the command doesn't need to derive from ManifestCommand (indirectly) anymore.

NikolaMilosavljevic: Working on this.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/107000f8be0c22ef4cdb74aabeb63e637f2f5a84

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/107000f8be0c22ef4cdb74aabeb63e637f2f5a84

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            if (!string.IsNullOrEmpty(stdOut) && stdOut.Contains("Discovered 0 artifact"))
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        public bool AnnotateEolDigest(string digest, DateOnly? date, ILoggerService loggerService, bool isDryRun)
```
mthalman, line 31:
```suggestion
        public bool AnnotateEolDigest(string digest, DateOnly date, ILoggerService loggerService, bool isDryRun)
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsOptions.cs
NikolaMilosavljevic: Prereq refactoring was merged with https://github.com/dotnet/docker-tools/pull/1342

Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/c2974135dd2894f75f8ce25a37ac35155eb76afb

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/881adf8e89e37f0abd325e01013a8ad07d188082

src/Microsoft.DotNet.ImageBuilder/src/OrasService.cs
NikolaMilosavljevic: ORAS version 1.2.0 changes the outputs of `discover` command and deprecates `output` option, besides other changes. It does add `format` option which can be used to produce the exact same output as before, with version 1.1.0.

Will update this shortly.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/fa106da9db77aec874013cc1663520a2471ceb30

mthalman: It would be better to output this as JSON and parse that rather than with unstructured text.

NikolaMilosavljevic: Updated the code to parse json output of `oras discover` command - https://github.com/dotnet/docker-tools/pull/1334/commits/4387c71365c5b1c114c799a5cb8c3fe2ca04783e

@mthalman @lbussell

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/EolAnnotationsData.cs
```diff
@@ -0,0 +1,29 @@
...
+            EolDate = eolDate;
+            EolDigests = eolDigests;
+        }
+
+        public DateOnly? EolDate { get; set; }
+
+        public List<EolDigestData>? EolDigests { get; set; }
+    }
+}
+#nullable disable
```
mthalman, line 29: Not needed

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/OciManifest.cs
```diff
@@ -0,0 +1,24 @@
...
+#nullable enable
+namespace Microsoft.DotNet.ImageBuilder.Models.EolAnnotations
+{
+    public class OciManifest
+    {
+        public string? Reference { get; set; }
+        public string? MediaType { get; set; }
+        public string? Digest { get; set; }
+        public int? Size { get; set; }
+        public IDictionary<string, string>? Annotations { get; set; }
```
mthalman, line 16: Since we only require `ArtifactType` I recommend excluding these other properties. Since the output of the CLI is still in flux, it would help to limit what can end up breaking us.

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/OciManifest.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+#nullable enable
+namespace Microsoft.DotNet.ImageBuilder.Models.EolAnnotations
```
mthalman, line 8: Namespace doesn't match file path.
```suggestion
namespace Microsoft.DotNet.ImageBuilder.Models.Annotations
```

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/EolAnnotationsData.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/3227d5c9b5121df01f5059ed7dc3fec4e13d4950

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/OciManifest.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/3227d5c9b5121df01f5059ed7dc3fec4e13d4950

src/Microsoft.DotNet.ImageBuilder/src/Models/Annotations/OciManifest.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1334/commits/3227d5c9b5121df01f5059ed7dc3fec4e13d4950

PR #1336, opened 2024-06-14
Fix staging server variable reference for PowerShell

mthalman:
The use of `$env:ACR-STAGING_SERVER` is incorrect syntax. It results in PowerShell looking for an environment variable named `ACR` and then just appends the string `-STAGING_SERVER` to the result because `-` is a variable delimiter. Since there is no `ACR` variable, the result of this is `-STAGING_SERVER` which obviously isn't what we want.

Fixed to use `{..}` syntax to include the entire variable name.

### DIFF
--- a/eng/common/templates/steps/test-images-windows-client.yml
+++ b/eng/common/templates/steps/test-images-windows-client.yml
@@ -23,7 +23,7 @@ steps:
 - ${{ parameters.customInitSteps }}
 - powershell: |
     if ("${{ variables['System.TeamProject'] }}" -eq "${{ parameters.internalProjectName }}" -and "${{ variables['Build.Reason'] }}" -ne "PullRequest") {
-      $optionalTestArgs="$optionalTestArgs -PullImages -Registry $env:ACR-STAGING_SERVER -RepoPrefix $env:STAGINGREPOPREFIX -ImageInfoPath $(artifactsPath)/image-info.json"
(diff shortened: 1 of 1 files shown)


lbussell (approved):

ellahathaway:
Thanks for fixing! Can you explain how you discovered this? I went back over [the test build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2462833&view=logs&j=ac422da1-c6cf-5423-8ede-22ab7060639a) I did with the initial (breaking) change, and I'm struggling to determine where the error occurred. I'd like to know what testing I could've done to catch this error earlier.

mthalman:
> Thanks for fixing! Can you explain how you discovered this? I went back over [the test build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2462833&view=logs&j=ac422da1-c6cf-5423-8ede-22ab7060639a) I did with the initial (breaking) change, and I'm struggling to determine where the error occurred. I'd like to know what testing I could've done to catch this error earlier.

I discovered it in this [build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2473599&view=logs&j=8fbd142f-3569-5a30-ef1a-e765423f275f&t=742f37bc-4355-57a9-a13d-16021ca96e22) (internal link). That build attempts to use the value of the staging server which then gets the error. In your test build, it doesn't pull any images so it doesn't run into the problem. The Image Builder pipeline is unique in that way. It just runs unit tests against the project and doesn't actually test any built images. That's my bad for not realizing this and letting you know.

PR #1337, opened 2024-06-14
Replace rendering tool w/ custom implementation

mthalman:
This ports the code that was originally in mcr.microsoft.com/mcr/renderingtool:1.0 so that it's contained within Image Builder itself. This is needed because the rendering tool is deprecated.

I've tested this logic against the readmes for both dotnet-docker and dotnet-framework-docker. It produces identical output.

Fixes dotnet/dotnet-docker#5561


lbussell (approved):
I didn't look very closely at the code but tested it locally and verified that it worked. So much faster!

PR #1340, opened 2024-06-17
EOL annotations pipeline

NikolaMilosavljevic:

This work requires new annotation command, being added with https://github.com/dotnet/docker-tools/pull/1334

New pipeline allows annotation of EOL digests as a standalone, one-off run, separate from regular image-building pipeline runs.

Digests are specified in a checked-in JSON file in this repo - in a separate branch, used for a specific run.

JSON file format (with sample dates):
```
{
  "eolDate": "2024-05-27",
  "eolDigests": [
    {
      "digest": "digest-value"
    },
    {
      "digest": "digest-value",
      "eolDate" : "2022-03-01"
    }
  ]
}
```

EOL data can be specified as a global value (optional) for all digests in the file. Each digest can have its own specific EOL date, which overrides the global value (if set).


eng/pipelines/annotate-eol-digests.yml
```diff
@@ -0,0 +1,46 @@
...
+    stages:
+    - stage: eolAnnotate
+      displayName: Annotate EOL images
+      jobs:
+      - job: AnnotateImages
+        displayName: Annotate EOL Images
+        steps:
+        - template: /eng/common/templates/steps/init-docker-linux.yml@self
+          parameters:
+            cleanupDocker: true
```
mthalman, line 37: Cleanup won't be necessary. That's just to ensure we have maximum disk space available which isn't a concern in this scenario.

eng/pipelines/annotate-eol-digests.yml
```diff
@@ -0,0 +1,46 @@
...
+        - template: /eng/common/templates/steps/annotate-eol-digests.yml@self
+          parameters:
+            internalProjectName: ${{ variables.internalProjectName }}
+            dataFile: ${{ parameters.dataFile }}
+            force: ${{ parameters.force }}
+            dockerRegistry: ${{ parameters.dockerRegistry }}
+            acrSubscription: ${{ parameters.acrSubscription }}
+            acrResourceGroup: ${{ parameters.acrResourceGroup }}
+        - template: /eng/common/templates/steps/cleanup-docker-linux.yml@self
```
mthalman, line 46: This isn't needed.

eng/common/templates/steps/annotate-eol-digests.yml
```diff
@@ -0,0 +1,31 @@
...
+      internalProjectName: ${{ parameters.internalProjectName }}
+      args: >
+        annotateEolDigests
+        /repo/${{ parameters.dataFile }}
+        --manifest $(manifest)
+        --var UniqueId=0000000
+        --acr-subscription '${{ parameters.acrSubscription }}'
+        --acr-resource-group '${{ parameters.acrResourceGroup }}'
+        --registry-override '${{ parameters.dockerRegistry }}'
+        --repo-prefix build-staging/9999999/
```
mthalman, line 30: None of these options are actually used.

eng/common/templates/steps/annotate-eol-digests.yml
```diff
@@ -0,0 +1,31 @@
...
+    parameters:
+      name: AnnotateEOLImages
+      displayName: Annotate EOL Images
+      serviceConnection: $(acr.serviceConnectionName)
+      internalProjectName: ${{ parameters.internalProjectName }}
+      args: >
+        annotateEolDigests
+        /repo/${{ parameters.dataFile }}
+        --manifest $(manifest)
+        --var UniqueId=0000000
```
mthalman, line 26: We shouldn't be providing these as they are not needed for this scenario. See https://github.com/dotnet/docker-tools/pull/1334#discussion_r1643001287.

eng/common/templates/steps/annotate-eol-digests.yml
NikolaMilosavljevic: Wil fix - thanks.

NikolaMilosavljevic: I think `--registry-override` option is needed, for log-in to registry, especially if we get rid of manifest options. We need to specify registry name in the call to `ExecuteWithCredentialsAsync`. Right now we use `Manifest.Registry`.

mthalman: This can be provided as a new argument for this command instead. Using `registry-override` is specific to the manifest but we don't need the manifest at all for this command.

eng/pipelines/annotate-eol-digests.yml
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1340/commits/8637ca7ffa28b700ced29eef4fdf76687326e83b

eng/common/templates/steps/annotate-eol-digests.yml
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1340/commits/8637ca7ffa28b700ced29eef4fdf76687326e83b

eng/common/templates/steps/annotate-eol-digests.yml
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1340/commits/8637ca7ffa28b700ced29eef4fdf76687326e83b

eng/common/templates/steps/annotate-eol-digests.yml
```diff
@@ -0,0 +1,24 @@
...
+  - template: /eng/common/templates/steps/run-imagebuilder.yml@self
+    parameters:
+      name: AnnotateEOLImages
+      displayName: Annotate EOL Images
+      serviceConnection: $(acr.serviceConnectionName)
+      internalProjectName: ${{ parameters.internalProjectName }}
+      args: >
+        annotateEolDigests
+        /repo/${{ parameters.dataFile }}
+        ${{ parameters.dockerRegistry }}
```
mthalman, line 23: This should just be `$(acr.server)`. It doesn't need to be parameterized.

NikolaMilosavljevic: We do need to specify server as a pipeline parameter though. How would it know which one to use?

NikolaMilosavljevic: Sample run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2477542&view=results

mthalman: There's only one it should be using `$(acr.server)`. That's the production ACR.

NikolaMilosavljevic: OK, that makes sense. Test ACRs are not in scope for this pipeline once the PR is merged.

I do need to update this PR, as I don't have a check-box for `dry run`, like other pipelines.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1340/commits/83de4c42a1cd1975da51f9d5c4cd8ef6b8887ea1

mthalman (approved):

PR #1342, opened 2024-06-18
Refactor DockerRegistry hierarchy and dependencies

NikolaMilosavljevic:
This PR fixes https://github.com/dotnet/docker-tools/pull/1334#discussion_r1643001287

- Removes `DockerRegistryCommand` from hierarchy, so `BuildCommand` and `PublishManifestCommand` inherit from `ManifestCommand`.
- Gets rid of `DockerRegistryOptions` altogether as it contained just `RegistryCredentialsOptions` which is independently composable.
- Moves `ExecuteWithCredentialsAsync` method to a new file for `RegistryCredentialsProvider` extensions.


src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -30,6 +30,8 @@ public class BuildCommand : DockerRegistryCommand<BuildOptions, BuildOptionsBuil
         private readonly List<TagInfo> _processedTags = new List<TagInfo>();
         private readonly HashSet<PlatformData> _builtPlatforms = new();
 
+        protected IRegistryCredentialsProvider RegistryCredentialsProvider { get; init; }
```
mthalman, line 33: This doesn't have any derived classes so there's no need to make it `protected`. I would recommend following the same pattern as the other DI components. Same comment applies to this pattern in `PublishManifestCommand`.

```suggestion
        private readonly IRegistryCredentialsProvider _registryCredentialsProvider;
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryCommand.cs
```diff
@@ -2,49 +2,19 @@
...
-using System;
-using System.Threading.Tasks;
-
 #nullable enable
 namespace Microsoft.DotNet.ImageBuilder.Commands
 {
-    public abstract class DockerRegistryCommand<TOptions, TOptionsBuilder> : ManifestCommand<TOptions, TOptionsBuilder>
-        where TOptions : DockerRegistryOptions, new()
-        where TOptionsBuilder : DockerRegistryOptionsBuilder, new()
+    public abstract class DockerRegistryCommand<TOptions, TOptionsBuilder> : Command<TOptions, TOptionsBuilder>
```
mthalman, line 8: This class should be deleted. It shouldn't be used anywhere now.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
NikolaMilosavljevic: Makes sense, I was on a fence making this a private field, but  thought it might be useful in the future - we could change it if needed later.

src/Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryCommand.cs
NikolaMilosavljevic: Yeah, it's not used anywhere - will remove.

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1342/commits/1bd3cfd6b0382fc3ac3b809e186aa794a2267f66

src/Microsoft.DotNet.ImageBuilder/src/Commands/DockerRegistryCommand.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1342/commits/1bd3cfd6b0382fc3ac3b809e186aa794a2267f66

mthalman (approved):

PR #1346, opened 2024-06-19
Use scenario-based service connections

mthalman:
Fixes https://dev.azure.com/dnceng/internal/_workitems/edit/6397. This updates the pipeline YAML to consume new service connections that are scenario-specific.


mthalman:
This will need to update https://github.com/dotnet/docker-tools/pull/1340/files#diff-d39079bb88b4127d62a9063364098bbb6d0fcd6e5ea70fba9ff9661f01cde625R18 as well once that gets merged.

PR #1358, opened 2024-07-01
Command for generation of EOL annotation data file

NikolaMilosavljevic:

Opening this PR as a draft. There are some areas that could use a little more cleanup and refactoring.

This work adds a new command, with related tests, for generation of EOL annotation data file.

The command is doing the following:
- Compares old and new image-info files to determine which digests should be marked for EOL
- Utilizes `Microsoft.Deployment.DotNet.Releases` library to determine EOL dates of .NET products and use them in annotations of images and platforms
- Queries Azure event log telemetry to detect any 'dangling' digests - i.e. those that are present in ACR but were never added to image-info files

Pipeline work, to enable annotation of EOL digests is not present in this PR. That work isn't complex, however it is incomplete and untested. Shared in https://github.com/NikolaMilosavljevic/docker-tools/pull/new/pipeline.eol.annotations


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        private void AnnotateDanglingDigests(string oldDigest, string tag, DateOnly? eolDate = null, string? newDigest = null)
+        {
+            // If newDigest is null, annotate all dangling digests starting with oldDigest.
+            // newDigest can be null if image or platform is being removed.
+            // If newDigest is not null, annotate all dangling digests starting with oldDigest, except newDigest.
+
+            string[] oldDigestParts = oldDigest.Split('@');
+            string repo = oldDigestParts[0].Replace("mcr.microsoft.com/", "public/");
```
mthalman, line 166: `public/` should be passed in as a command option. You can copy this option over: https://github.com/dotnet/docker-tools/blob/796f5119eba72bd2187994a4b9e0d64fd3b695f5/src/Microsoft.DotNet.ImageBuilder/src/Commands/ManifestOptions.cs#L54-L55

src/Microsoft.DotNet.ImageBuilder/src/AzureLogService.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.DotNet.ImageBuilder.Models.Annotations;
+using System.Linq;
+
+#nullable enable
+namespace Microsoft.DotNet.ImageBuilder
+{
+    [Export(typeof(IAzureLogService))]
+    public class AzureLogService : IAzureLogService
+    {
+        private string AcrLogsWorkspaceId = "4db2f537-e3cf-4a80-97ba-9e2e66717407";
```
mthalman, line 22: This should be configurable as a command option. We don't want this hardcoded.

src/Microsoft.DotNet.ImageBuilder/src/AzureLogService.cs
```diff
@@ -0,0 +1,58 @@
...
+        private const int TimespanDays = 7;
```
mthalman, line 23: This should be configurable as a command option as well, but defaulted to `7`.

src/Microsoft.DotNet.ImageBuilder/src/AzureLogService.cs
```diff
@@ -0,0 +1,58 @@
...
+            foreach (LogsTableRow row in logsTable.Rows)
+            {
+                entries.Add(new AcrEventEntry
+                {
+                    TimeGenerated = DateTime.Parse(row["TimeGenerated"].ToString()),
+                    Digest = row["Digest"].ToString()
+                });
+            }
+
+            entries.Sort((x, y) => x.TimeGenerated.CompareTo(y.TimeGenerated));
```
mthalman, line 42: Wouldn't it be better to sort directly in the log query?

mthalman (commented):
I haven't looked at the tests yet but this to be on track with what I was expecting.

mthalman:
@dagood - Be aware that we're adding a dependency on Azure Logs Workspace here that contains the ACR events. That would be necessary to configure in order to support EOL annotations.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -0,0 +1,272 @@
...
+                if (platform.SimpleTags?.Count > 0)
+                {
+                    tag = platform.SimpleTags.First();
+                }
+
+                AddDigestForAnnotation(platform.Digest, tag, eolDate, isNewDigest: isNewImage);
+            }
+        }
+
+        private void AddRepoForAnnotation(RepoData repo, DateOnly? eolDate = null, bool isNewRepo = false)
```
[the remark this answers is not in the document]
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1358/commits/13ce63d392b9598c79830c687f31e5a08f5a56ce

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -0,0 +1,272 @@
...
+
+            SerializeDigestDataJson();
+        }
+
+        private void SerializeDigestDataJson()
+        {
+            EolAnnotationsData eolAnnotations = new EolAnnotationsData { EolDate = _eolDate, EolDigests = [] };
+            foreach (KeyValuePair<string, DateOnly?> digest in _digestsToAnnotate)
+            {
+                eolAnnotations.EolDigests!.Add(new EolDigestData { Digest = digest.Key, EolDate = digest.Value });
```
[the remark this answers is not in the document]
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1358/commits/13ce63d392b9598c79830c687f31e5a08f5a56ce

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1358/commits/13ce63d392b9598c79830c687f31e5a08f5a56ce

src/Microsoft.DotNet.ImageBuilder/src/AzureLogService.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1358/commits/13ce63d392b9598c79830c687f31e5a08f5a56ce

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -0,0 +1,235 @@
...
+    private readonly IDotNetReleasesService _dotNetReleasesService;
+    private readonly ILoggerService _loggerService;
+    private readonly IContainerRegistryClientFactory _acrClientFactory;
+    private readonly IAzureTokenCredentialProvider _tokenCredentialProvider;
+    private readonly IOrasService _orasService;
+    private readonly DateOnly _eolDate;
+
+    private Dictionary<string, DateOnly?> _productEolDates = null!;
+    private ImageArtifactDetails _oldImageArtifactDetails = null!;
+    private ImageArtifactDetails _newImageArtifactDetails = null!;
```
lbussell, line 33: Prefer nullable annotations over the null-forgiving operator. If that's annoying we could consider passing these as arguments to `GetDigestsToAnnotate()` rather than keeping the class stateful.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+                    CreateOption<bool>("annotate-eol-products", nameof(GenerateEolAnnotationDataOptions.AnnotateEolProducts),
+                        "Annotate images of EOL products"),
+                ]
+            );
+
+    public override IEnumerable<Argument> GetCliArguments() =>
+        base.GetCliArguments()
+            .Concat(
+                [
+                    new Argument<string>(nameof(AnnotateEolDigestsOptions.EolDigestsListOutputPath),
```
lbussell, line 38:
```suggestion
                    new Argument<string>(nameof(GenerateEolAnnotationDataOptions.EolDigestsListPath),
```

I think this one is a bug - it causes an exception later on when you run the code.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        digestDataList = digestDataList.OrderBy(item => item.Digest).ToList();
+
+        return digestDataList;
+    }
+
+    /// <summary>
+    /// Finds all the digests that are in the registry but not in the supported digests list.
+    /// </summary>
+    private static IEnumerable<EolDigestData> GetUnsupportedDigests(IEnumerable<(string Digest, string? Tag)> registryDigests, IEnumerable<string> supportedDigests)
```
lbussell, line 128: This method can just be:

```cs
registryDigests
    .Where(registryDigest => !supportedDigests.Contains(registryDigest.Digest))
    .Select(registryDigest => new EolDigestData(registryDigest.Digest) { Tag = registryDigest.Tag });
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -0,0 +1,235 @@
...
+            registryIndex++;
+        }
+
+        return result;
+    }
+
+    private IEnumerable<string> GetSupportedDigests() =>
+        _newImageArtifactDetails.Repos
+            .SelectMany(repo => repo.Images)
+            .SelectMany(image => GetImageDigests(image))
```
lbussell, line 174:
```suggestion
            .SelectMany(GetImageDigests)
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -0,0 +1,235 @@
...
+        foreach (PlatformData platform in image.Platforms)
+        {
+            yield return (platform.Digest, GetLongestTag(platform.SimpleTags));
+        }
+    }
+
+    private static string? GetLongestTag(IEnumerable<string> tags) =>
+        tags.OrderByDescending(tag => tag.Length).FirstOrDefault();
+
+    private async Task<IEnumerable<(string Digest, string? Tag)>> GetAllDigestsFromRegistry(IEnumerable<string> repoNames)
```
lbussell, line 193: I have nothing to say besides I tried to speed up this method and I couldn't :)

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -0,0 +1,235 @@
...
+        if (image.ProductVersion != null)
+        {
+            string dotnetVersion = Version.Parse(image.ProductVersion).ToString(2);
+            if (_productEolDates != null && _productEolDates.TryGetValue(dotnetVersion, out DateOnly? date))
+            {
+                return GetImageDigests(image).Select(val => new EolDigestData(val.Digest) { Tag = val.Tag, EolDate = date });
+            }
+        }
+
+        return [];
```
lbussell, line 224: Consider inverting these conditions and returning early?

```suggestion
        if (image.ProductVersion == null)
        {
            return [];
        }

        string dotnetVersion = Version.Parse(image.ProductVersion).ToString(2);

        if (_productEolDates == null || !_productEolDates.TryGetValue(dotnetVersion, out DateOnly? date))
        {
            return [];
        }

        return GetImageDigests(image).Select(val => new EolDigestData(val.Digest) { Tag = val.Tag, EolDate = date });
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -0,0 +1,195 @@
...
+
+            if (Options.AnnotateEolProducts)
+            {
+                // Annotate images for eol products in new image info
+                foreach (ImageData image in newImageArtifactDetails.Repos.SelectMany(repo => repo.Images))
+                {
+                    digestDataList.AddRange(GetProductEolDigests(image, productEolDates));
+                }
+            }
+        }
```
lbussell, line 109: Sorry if this has already been discussed. But in what case would we decide to build and ship an image for a product that is already EOL, and mark it as EOL immediately?

lbussell (approved):
Just one question.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
mthalman: For .NET major version EOL. The last patch ships on the EOL date.

PR #1359, opened 2024-07-02
Close non-failure notification issues immediately after opening

### LINKED ISSUE #1305: Close "AutoBuilder Queued" and "Publish Success" notification issues immediately after opening, leave failures open
The Autobuilder notifications generate a tremendous amount of noise and prevent our internal repo issues page from being actionable.

I propose that the bot closes "AutoBuilder Queued" and "Publish Success" issues immediately after opening, since they require no action. Team members who have subscribed via email will still be able to see the notifications going by for informational purposes (and can filter them out to a separate folder if necessary).

"AutoBuilder Failed" notifications should be left open so that the issues page of our internal repo contains only actionable information. This also makes it so we don't have to go in and clean up hundreds of issues every couple of weeks.

lbussell:
Fixes #1305


lbussell:
Here's an example notification which was automatically closed by this code:

- https://github.com/dotnet/dotnet-docker-internal/issues/5840

src/Microsoft.DotNet.ImageBuilder/src/NotificationService.cs
```diff
@@ -68,7 +78,19 @@ public NotificationService(ILoggerService loggerService)
...
+            }
+
+            // Immediately close issues which aren't failures, since open issues should represent actionable items
+            if (!labels.Where(l => l.Contains(Commands.NotificationLabels.Failure)).Any())
+            {
+                _loggerService.WriteMessage("No failure label found in the notification labels.");
+                await github.Issue.Update(repoOwner, repoName, result.IssueId, new IssueUpdate { State = ItemState.Closed });
+            }
+
+            return result;
```
mthalman, line 93: It doesn't look like this result is ever consumed by the caller. If so, this should be changed to not return anything. That would mean you could also get rid of the `NotificationPublishResult` type.

mthalman (approved):

PR #1362, opened 2024-07-11
Upgrade System.Formats.Asn1 version

mthalman:
Addresses a CG issue with Image Builder for CVE-2024-38095 by explicitly upgrading System.Formats.Asn1 that's a dependency of Microsoft.DotNet.VersionTools.


mthalman:
Actually, I'm going to address this in arcade and then we'll just upgrade VersionTools here.

PR #1364, opened 2024-07-18
Remove force annotate option

mthalman:
The `annotateEolDigests` command provides a `--force` option that will always create an EOL annotation even if one already exists. This was intended to handle scenarios where we essentially want to overwrite an incorrectly defined annotation. However, this causes multiple such annotations to exist in the registry. We don't want obsolete annotations to continue to be around.

The correct thing to do here is to delete the existing annotation before pushing the new one. But it's not as simple as just applying those operations to the manifests in the ACR. That gets applied to the ACR, but the deletion does not propagate to MAR. So if you did that, you'd effectively still be left with multiple annotations in MAR. Deletions in MAR need to be done through a separate workflow that is not automated.

So there isn't a way to fully automate this scenario. For that reason, the command is being modified here to do away with the `--force` option entirely. In addition, it will report on image digests which have existing annotations. There is logic to check whether the existing annotation already matches the EOL date to be set. If so, it simply skips the annotation. If it differs, it will be skipped but logged as an error as it will need to be addressed through the manual steps described above.


src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
```diff
@@ -44,50 +39,82 @@ public class AnnotateEolDigestsCommand : Command<AnnotateEolDigestsOptions, Anno
...
+            DateOnly? globalEolDate = eolAnnotations.EolDate;
 
             await _registryCredentialsProvider.ExecuteWithCredentialsAsync(
                 Options.IsDryRun,
-                async () =>
+                () =>
                 {
-                    Parallel.ForEach(eolAnnotations.EolDigests, (a) =>
+                    Parallel.ForEach(eolAnnotations.EolDigests, digestData =>
                     {
```
lbussell, line 49: You can use expression body without the curly braces here.

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
```diff
@@ -44,50 +39,82 @@ public class AnnotateEolDigestsCommand : Command<AnnotateEolDigestsOptions, Anno
...
+                    $"Some digest annotations failed or were skipped due to existing non-matching EOL date annotations (failed: {_failedAnnotations.Count}, skipped: {_existingAnnotations.Count}).");
+            }
+        }
+
+        private void WriteNonEmptySummary(ConcurrentBag<EolDigestData> eolDigests, string message)
+        {
+            if (!eolDigests.IsEmpty)
+            {
+                _loggerService.WriteMessage(message);
                 _loggerService.WriteMessage("");
```
lbussell, line 80:
```suggestion
                _loggerService.WriteMessage();
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
```diff
@@ -44,50 +39,82 @@ public class AnnotateEolDigestsCommand : Command<AnnotateEolDigestsOptions, Anno
...
-                _loggerService.WriteMessage(JsonConvert.SerializeObject(new EolAnnotationsData(eolDigests: [.. _failedAnnotations])));
+                _loggerService.WriteMessage(JsonConvert.SerializeObject(new EolAnnotationsData(eolDigests: [.. eolDigests])));
                 _loggerService.WriteMessage("");
```
lbussell, line 82:
```suggestion
                _loggerService.WriteMessage();
```

src/Microsoft.DotNet.ImageBuilder/src/IOrasService.cs
```diff
@@ -3,13 +3,15 @@
...
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.DotNet.ImageBuilder.Models.Annotations;
 
 #nullable enable
 namespace Microsoft.DotNet.ImageBuilder
 {
     public interface IOrasService
     {
-        bool IsDigestAnnotatedForEol(string digest, ILoggerService loggerService, bool isDryRun);
+        bool IsDigestAnnotatedForEol(string digest, ILoggerService loggerService, bool isDryRun, [MaybeNullWhen(false)] out OciManifest lifecycleArtifactManifest);
```
lbussell, line 14:
```suggestion
        bool IsDigestAnnotatedForEol(string digest, ILoggerService loggerService, bool isDryRun, [NotNullWhen(true)] out OciManifest? lifecycleArtifactManifest);
```

lbussell (approved):

src/Microsoft.DotNet.ImageBuilder/src/Commands/AnnotateEolDigestsCommand.cs
```diff
@@ -4,37 +4,42 @@
...
-        private readonly IProcessService _processService;
         private readonly IOrasService _orasService;
         private readonly IRegistryCredentialsProvider _registryCredentialsProvider;
+        private readonly ConcurrentBag<EolDigestData> _failedAnnotationImageDigests = [];
+        private readonly ConcurrentBag<EolDigestData> _skippedAnnotationImageDigests = [];
+        private readonly ConcurrentBag<EolDigestData> _existingAnnotationImageDigests = [];
+        private readonly ConcurrentBag<string> _existingAnnotationDigests = [];
 
-        private ConcurrentBag<EolDigestData> _failedAnnotations = new ();
+        private static readonly JsonSerializerOptions _jsonSerializerOptions = new()
```
lbussell, line 30:
```suggestion
        private static readonly JsonSerializerOptions s_jsonSerializerOptions = new()
```

src/Microsoft.DotNet.ImageBuilder/src/Models/MarBulkDeletion/BulkDeletionDescription.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.ImageBuilder.Models.MarBulkDeletion;
+
+#nullable enable
+internal class BulkDeletionDescription
+{
+    public string RegistryType { get; set; } = "public";
+    public List<string> Digests { get; set;} = [];
+}
```
lbussell, line 13:
```suggestion
internal record BulkDeletionDescription
{
    public string RegistryType { get; init; } = "public";
    public List<string> Digests { get; init; } = [];
}
```

lbussell (approved):
Just a couple of nits

lbussell (approved):

PR #1367, opened 2024-07-23
Add command to wait for EOL annotation ingestion

mthalman:
This will enable the pipeline to validate that the EOL annotations that have been pushed are successfully ingested by MAR.

This refactors the existing image ingestion command so that it can reuse the logic defined there. See separate commits for refactoring.


src/Microsoft.DotNet.ImageBuilder/src/Commands/WaitForMcrImageIngestionOptions.cs
```diff
@@ -17,41 +17,34 @@ public class WaitForMcrImageIngestionOptions : ManifestOptions
...
+                .Concat(_ingestionOptionsBuilder.GetCliArguments())
                 .Concat(
-                    new Argument[]
-                    {
+                    [
                         new Argument<string>(nameof(WaitForMcrImageIngestionOptions.ImageInfoPath),
                             "Path to image info file")
-                    }
+                    ]
                 );
```
lbussell, line 38: If you are refactoring these you may as well go all-in on the collection expressions, like https://github.com/dotnet/docker-tools/blob/f4ec11104921b8f0d734e669338fb3d998154faf/src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyImagesOptions.cs#L30-L38

src/Microsoft.DotNet.ImageBuilder/src/Commands/WaitForMarAnnotationIngestionOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        public string EolDigestsListPath { get; set; } = string.Empty;
+
+        public MarIngestionOptions IngestionOptions { get; set; } = new();
+    }
+
+    public class WaitForMarAnnotationIngestionOptionsBuilder : CliOptionsBuilder
+    {
+        private static readonly TimeSpan DefaultWaitTimeout = TimeSpan.FromMinutes(20);
+        private static readonly TimeSpan DefaultRequeryDelay = TimeSpan.FromSeconds(10);
```
lbussell, line 23:
```suggestion
        private static readonly TimeSpan s_defaultWaitTimeout = TimeSpan.FromMinutes(20);
        private static readonly TimeSpan s_defaultRequeryDelay = TimeSpan.FromSeconds(10);
```

src/Microsoft.DotNet.ImageBuilder/src/MarImageIngestionReporter.cs
```diff
@@ -0,0 +1,282 @@
...
+            stringBuilder.AppendLine($"\tTag: {imageStatus.Tag}");
+            stringBuilder.AppendLine($"\tCommit digest: {result.CommitDigest}");
+            stringBuilder.AppendLine($"\tReason: {imageStatus.FailureReason}");
+            stringBuilder.Append(result.Substatus.ToString("\t"));
+            return stringBuilder.ToString();
+        }
+    }
+}
+
+internal class ImageResultInfo
```
lbussell, line 253: Should both of these be records and go under `Models/McrStatus`?

lbussell (commented):
Not totally done reading through this yet.

src/Microsoft.DotNet.ImageBuilder/src/MarImageIngestionReporter.cs
mthalman: I've moved `ImageResultInfo` as a nested type within `MarImageIngestionReporter` since it's only used from there.

The files in the Models directory are intended for data types that are [de]serialized from services or JSON files. So the `McrStatus` directory is meant for deserializing data from MCR's status service. In this case, `DigestInfo` is really just a supporting data type for `MarImageIngestionReporter`. So I prefer leaving it where it is.

src/Microsoft.DotNet.ImageBuilder/src/MarImageIngestionReporter.cs
```diff
@@ -0,0 +1,282 @@
...
+
+#nullable enable
+
+public interface IMarImageIngestionReporter
+{
+   Task ReportImageStatusesAsync(IEnumerable<DigestInfo> digestInfos, TimeSpan timeout, TimeSpan requeryDelay, DateTime? minimumQueueTime);
+}
+
+[Export(typeof(IMarImageIngestionReporter))]
+public class MarImageIngestionReporter : IMarImageIngestionReporter
```
lbussell, line 22: Would it be more accurate to call this `MarDigestIngestionReporter` given we are using it for more than just images now?

src/Microsoft.DotNet.ImageBuilder/src/MarImageIngestionReporter.cs
```diff
@@ -0,0 +1,282 @@
...
+        _loggerService = loggerService;
+        _statusClient = statusClient;
+        _environmentService = environmentService;
+    }
+
+    public Task ReportImageStatusesAsync(IEnumerable<DigestInfo> digestInfos, TimeSpan timeout, TimeSpan requeryDelay, DateTime? minimumQueueTime) =>
+        new ReporterImpl(_loggerService, _statusClient, _environmentService, timeout, requeryDelay, minimumQueueTime)
+            .ReportImageStatusesAsync(digestInfos);
+
+    private class ReporterImpl
```
lbussell, line 40: What's the purpose of putting this functionality into a private class? To make sure that `MarImageIngestionReporter` doesn't depend on any methods but `ReportImageStatusesAsync()`?

lbussell (approved):

src/Microsoft.DotNet.ImageBuilder/src/MarImageIngestionReporter.cs
mthalman: No, it's so we can have fields for the TimeSpan/DateTime values rather than needing to pass them around all over the stack.

src/Microsoft.DotNet.ImageBuilder/src/MarImageIngestionReporter.cs
mthalman: I debated with myself on that. I went with Image since that's the term used in the MAR status API we use (for both scenarios). I'm willing to change it if you want.

lbussell: I think it's fine then, taking `DigestInfo` as an argument communicates that use-case well enough.

lbussell (approved):

PR #1372, opened 2024-07-30
Clean up unused packages and update remaining packages

lbussell:
- Used VS's "Remove unused references" feature to remove packages that weren't being used
  - It mistakenly removed `Azure.Identity`, which we are using directly as well as referencing as a transient dependency through `Microsoft.Azure.Kusto.Ingest`, so I added it back manually
- Updated all packages that could be updated
- Fixed a new CG alert for CVE-2024-38095


eng/src/file-pusher/file-pusher.csproj
```diff
@@ -9,10 +9,13 @@
...
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.DotNet.VersionTools" Version="9.0.0-beta.24151.5" />
+    <PackageReference Include="Microsoft.DotNet.VersionTools" Version="9.0.0-beta.24379.1" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20574.7" />
     <PackageReference Include="System.Diagnostics.TextWriterTraceListener" Version="4.3.0" />
+
+    <!-- CVE-2024-38095: Upgrade version of System.Formats.Asn1 implicitly referenced by Microsoft.DotNet.VersionTools -->
+    <PackageReference Include="System.Formats.Asn1" Version="6.0.1" />
```
mthalman, line 18: I question whether this should be done here when the root fix should be made in Arcade.

src/Microsoft.DotNet.ImageBuilder/src/Microsoft.DotNet.ImageBuilder.csproj
mthalman, line 1: Does this also fix the other CG alert I see for Azure.Core?

eng/src/file-pusher/file-pusher.csproj
lbussell: Filed issue on arcade here: https://github.com/dotnet/arcade/issues/14983

src/Microsoft.DotNet.ImageBuilder/src/Microsoft.DotNet.ImageBuilder.csproj
lbussell: Azure.Core v1.40.0 has an issue with token retrieval - https://github.com/Azure/azure-sdk-for-net/issues/44817

We were actually resolving version 1.39.0. But I went ahead and pinned it anyways to 1.41.0 to have the newest version.

mthalman (approved):

PR #1379, opened 2024-08-06
Extend CleanAcrImagesCommand with support for cleanup of EOL images

NikolaMilosavljevic:
Allows cleanup of images that have EOL dates older than specified age parameter.

I have added the relevant test, but have not updated the pipeline YML. I'm unsure which parameters would be used for the new cleanup step and if we want to turn it on immediately.


src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -79,6 +83,10 @@ public override async Task ExecuteAsync()
                     await ProcessManifestsAsync(acrClient, acrContentClient, deletedImages, deletedRepos, repository,
                         manifest => !manifest.Tags.Any() && IsExpired(manifest.LastUpdatedOn, Options.Age));
                     break;
+                case CleanAcrImagesAction.PruneEol:
+                    await ProcessManifestsAsync(acrClient, acrContentClient, deletedImages, deletedRepos, repository,
+                        manifest => manifest.Tags.Any() && HasExpiredEol(manifest, Options.Age));
```
mthalman, line 88: We want _any_ image to be eligible for deletion, not just tagged images.

```suggestion
                        manifest => HasExpiredEol(manifest, Options.Age));
```

NikolaMilosavljevic: I was trying to avoid annotation manifests. Is there a better way of doing that?

Won't untagged images be deleted during cleanup of dangling images?

mthalman: > I was trying to avoid annotation manifests. Is there a better way of doing that?

See my changes in https://github.com/dotnet/docker-tools/pull/1378

> Won't untagged images be deleted during cleanup of dangling images?

No, we don't have a policy for cleaning dangling images in `public/*` repos, which is what this action type would be applied to.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1379/commits/38aa3f8b50aed5c3b85368216ddcbe1c9e4b5599

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -242,6 +242,14 @@ private async Task DeleteRepositoryAsync(IContainerRegistryClient acrClient, Lis
 
         private static bool IsExpired(DateTimeOffset dateTime, int expirationDays) => dateTime.AddDays(expirationDays) < DateTimeOffset.Now;
 
+        private bool IsAnnotationManifest(ArtifactManifestProperties manifest, IContainerRegistryContentClient acrContentClient)
+        {
+            ManifestQueryResult manifestResult = acrContentClient.GetManifestAsync(manifest.Digest).Result;
```
mthalman, line 247: This really should be using the async pattern here. `IsAnnotationManifest` should be an async method. So yes, this will have a ripple effect on other methods.

src/Microsoft.DotNet.ImageBuilder/tests/CleanAcrImagesCommandTest.cs
```diff
@@ -295,5 +293,95 @@ public async Task TestRepos()
...
+
+            const int age = 30;
+
+            ContainerRepository repo1 = CreateContainerRepository(
+                repo1Name,
+                CreateContainerRepositoryProperties(),
+                [
+                    CreateArtifactManifestProperties(repositoryName: repo1Name, digest: repo1Digest1, lastUpdatedOn: DateTimeOffset.Now.Subtract(TimeSpan.FromDays(8)), tags: ["latest"], registryLoginServer: AcrName),
+                    CreateArtifactManifestProperties(repositoryName: repo1Name, digest: repo1Digest2, lastUpdatedOn: DateTimeOffset.Now.Subtract(TimeSpan.FromDays(9)), tags: ["latest"], registryLoginServer: AcrName),
+                    CreateArtifactManifestProperties(repositoryName: repo1Name, digest: repo1Digest3, lastUpdatedOn: DateTimeOffset.Now.Subtract(TimeSpan.FromDays(10)), tags: ["latest"], registryLoginServer: AcrName)
```
mthalman, line 320: There should be another digest added that represents an annotation to demonstrate it's not getting cleaned up.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
mthalman: This would make the `canDeleteManifest` delegate an async function and it should be executed in parallel, not sequentially as it is today: https://github.com/dotnet/docker-tools/blob/f28f2506d48d560f806cf32ea629b1966c3e01ca/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs#L148-L150

That'd be a good thing. As I'm testing this, I see all these `oras discover` commands being run all sequentially and there are thousands to go through. So it's really slow.

NikolaMilosavljevic: Yeah, that was the reason I didn't use the `async` pattern. Let me see about modifying the delegate to be `async`.

src/Microsoft.DotNet.ImageBuilder/tests/CleanAcrImagesCommandTest.cs
NikolaMilosavljevic: Yes, will fix - thanks.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1379/commits/590cd16afb02c1c7ea5084de06dd442e37899c1a

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1379/commits/8893d7bb4c500e15e69fc9245aa50b72cf7cb4d2

mthalman: This still needs to be changed to use `await` instead of `Task.Result`.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -240,22 +247,22 @@ private async Task DeleteRepositoryAsync(IContainerRegistryClient acrClient, Lis
             }
         }
 
-        private static bool IsExpired(DateTimeOffset dateTime, int expirationDays) => dateTime.AddDays(expirationDays) < DateTimeOffset.Now;
+        private static async Task<bool> IsExpired(DateTimeOffset dateTime, int expirationDays) => dateTime.AddDays(expirationDays) < DateTimeOffset.Now;
 
-        private bool IsAnnotationManifest(ArtifactManifestProperties manifest, IContainerRegistryContentClient acrContentClient)
+        private async Task<bool> IsAnnotationManifest(ArtifactManifestProperties manifest, IContainerRegistryContentClient acrContentClient)
```
mthalman, line 252: Use an `Async` suffix for the method name.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -240,22 +247,22 @@ private async Task DeleteRepositoryAsync(IContainerRegistryClient acrClient, Lis
...
         {
             ManifestQueryResult manifestResult = acrContentClient.GetManifestAsync(manifest.Digest).Result;
 
             // An annotation is just a referrer and referrers are indicated by the presence of a subject field.
-            return manifestResult.Manifest["subject"] is null ? false : true;
+            return manifestResult.Manifest["subject"] is not null;
         }
 
-        private bool HasExpiredEol(ArtifactManifestProperties manifest, int expirationDays)
+        private async Task<bool> HasExpiredEol(ArtifactManifestProperties manifest, int expirationDays)
```
mthalman, line 260: Use an `Async` suffix for the method name.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -240,22 +247,22 @@ private async Task DeleteRepositoryAsync(IContainerRegistryClient acrClient, Lis
             }
         }
 
-        private static bool IsExpired(DateTimeOffset dateTime, int expirationDays) => dateTime.AddDays(expirationDays) < DateTimeOffset.Now;
+        private static async Task<bool> IsExpired(DateTimeOffset dateTime, int expirationDays) => dateTime.AddDays(expirationDays) < DateTimeOffset.Now;
```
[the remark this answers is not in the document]
NikolaMilosavljevic: OK - will fix.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
NikolaMilosavljevic: Will fix - thanks!

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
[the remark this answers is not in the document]
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1379/commits/1a49f5344dfce32187755f1501d4037ae26e9b51

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1379/commits/1a49f5344dfce32187755f1501d4037ae26e9b51

NikolaMilosavljevic: Also fixed naming of `IsAnnotationManifest` method.

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -77,11 +83,15 @@ public override async Task ExecuteAsync()
             {
                 case CleanAcrImagesAction.PruneDangling:
                     await ProcessManifestsAsync(acrClient, acrContentClient, deletedImages, deletedRepos, repository,
-                        manifest => !manifest.Tags.Any() && IsExpired(manifest.LastUpdatedOn, Options.Age));
+                        async manifest => !manifest.Tags.Any() && IsExpired(manifest.LastUpdatedOn, Options.Age));
```
mthalman, line 86:
```suggestion
                        manifest => Task.FromResult(!manifest.Tags.Any() && IsExpired(manifest.LastUpdatedOn, Options.Age)));
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -77,11 +83,15 @@ public override async Task ExecuteAsync()
...
+                    break;
+                case CleanAcrImagesAction.PruneEol:
+                    await ProcessManifestsAsync(acrClient, acrContentClient, deletedImages, deletedRepos, repository,
+                        async manifest => !(await IsAnnotationManifestAsync(manifest, acrContentClient)) && await HasExpiredEolAsync(manifest, Options.Age));
                     break;
                 case CleanAcrImagesAction.PruneAll:
                     await ProcessManifestsAsync(acrClient, acrContentClient, deletedImages, deletedRepos, repository,
-                        manifest => IsExpired(manifest.LastUpdatedOn, Options.Age));
+                        async manifest => IsExpired(manifest.LastUpdatedOn, Options.Age));
```
mthalman, line 94:
```suggestion
                        manifest => Task.FromResult(IsExpired(manifest.LastUpdatedOn, Options.Age)));
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -233,5 +248,24 @@ private async Task DeleteRepositoryAsync(IContainerRegistryClient acrClient, Lis
         }
 
         private static bool IsExpired(DateTimeOffset dateTime, int expirationDays) => dateTime.AddDays(expirationDays) < DateTimeOffset.Now;
+
+        private async Task<bool> IsAnnotationManifestAsync(ArtifactManifestProperties manifest, IContainerRegistryContentClient acrContentClient)
+        {
+            ManifestQueryResult manifestResult = acrContentClient.GetManifestAsync(manifest.Digest).Result;
```
mthalman:
```suggestion
            ManifestQueryResult manifestResult = await acrContentClient.GetManifestAsync(manifest.Digest);
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -233,5 +248,24 @@ private async Task DeleteRepositoryAsync(IContainerRegistryClient acrClient, Lis
...
+
+            // An annotation is just a referrer and referrers are indicated by the presence of a subject field.
+            return manifestResult.Manifest["subject"] is not null;
+        }
+
+        private async Task<bool> HasExpiredEolAsync(ArtifactManifestProperties manifest, int expirationDays)
```
mthalman, line 260: Now this method doesn't need to be async either.

```suggestion
        private bool HasExpiredEol(ArtifactManifestProperties manifest, int expirationDays)
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -77,11 +83,15 @@ public override async Task ExecuteAsync()
             {
                 case CleanAcrImagesAction.PruneDangling:
                     await ProcessManifestsAsync(acrClient, acrContentClient, deletedImages, deletedRepos, repository,
-                        manifest => !manifest.Tags.Any() && IsExpired(manifest.LastUpdatedOn, Options.Age));
+                        async manifest => !manifest.Tags.Any() && IsExpired(manifest.LastUpdatedOn, Options.Age));
+                    break;
+                case CleanAcrImagesAction.PruneEol:
+                    await ProcessManifestsAsync(acrClient, acrContentClient, deletedImages, deletedRepos, repository,
+                        async manifest => !(await IsAnnotationManifestAsync(manifest, acrContentClient)) && await HasExpiredEolAsync(manifest, Options.Age));
```
mthalman, line 90:
```suggestion
                        async manifest => !(await IsAnnotationManifestAsync(manifest, acrContentClient)) && HasExpiredEol(manifest, Options.Age));
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/docker-tools/pull/1379/commits/1a49f5344dfce32187755f1501d4037ae26e9b51

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -233,5 +248,24 @@ private async Task DeleteRepositoryAsync(IContainerRegistryClient acrClient, Lis
...
+        {
+            ManifestQueryResult manifestResult = acrContentClient.GetManifestAsync(manifest.Digest).Result;
+
+            // An annotation is just a referrer and referrers are indicated by the presence of a subject field.
+            return manifestResult.Manifest["subject"] is not null;
+        }
+
+        private async Task<bool> HasExpiredEolAsync(ArtifactManifestProperties manifest, int expirationDays)
+        {
+            if(_orasService.IsDigestAnnotatedForEol(manifest.RegistryLoginServer + "/" + manifest.RepositoryName + "@" + manifest.Digest, _loggerService, Options.IsDryRun, out OciManifest? lifecycleArtifactManifest) &&
```
mthalman, line 262: Explicitly set dry run to false here. The command has no side effects, so it's fine. It's necessary to do this because we want to allow using dry run for the clean command. But if we enable dry run for running the oras tool, the tool never gets invoked and this condition would never be true, giving incorrect results for the command to analyze.

```suggestion
            if(_orasService.IsDigestAnnotatedForEol(manifest.RegistryLoginServer + "/" + manifest.RepositoryName + "@" + manifest.Digest, _loggerService, isDryRun: false, out OciManifest? lifecycleArtifactManifest) &&
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
NikolaMilosavljevic: Yeah, missed this earlier - fixed with https://github.com/dotnet/docker-tools/pull/1379/commits/ae30e548513dd94eba4190ad4da567ec4f05978d

src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
NikolaMilosavljevic: This was fixed with https://github.com/dotnet/docker-tools/pull/1379/commits/ae30e548513dd94eba4190ad4da567ec4f05978d

mthalman (approved):

PR #1381, opened 2024-08-07
Add GenerateSigningPayloads command

lbussell:
Changes:
- New Command `GenerateSigningPayloads`
- Rename `OrasSerivce` -> `LifecycleMetadataService` - since it contains "business logic" I think those methods should live outside the ORAS client class. For some reason Git did not pick those up as renames.
- New `OrasClient` which only talks to ORAS CLI
- Removed need for nullable annotations in `ExecuteHelper` methods
- Rearranged models around Notary, ORAS, and OCI

Todo:
- [x] Authenticate
- [x] Validate payloads on-disk in tests
- [x] Handle registry overrides - ~~not sure how those manifest in the image info file - will this even be important if we only use digests in this command~~
- [ ] Write some additional tests
  - [x] Images without manifest lists
  - [ ] ...
- [x] Double-check dry-run functionality - is it even needed


lbussell:
Applying the `RegistryOverrideOptions` class to other commands is proving to be really involved and not really worth the ROI, so I'm going to stop that.

I am taking suggestions for other tests I could implement, but this command is pretty simple. So I think this is ready to review.

lbussell:
For some reason when running ImageBuilder's help command, this command shows up at the bottom while the others are in alphabetical order:

```
Commands:
  annotateEolDigests <EolDigestsListPath> <AcrName> <RepoPrefix> <AnnotationDigestsOutputPath>
  build
  cleanAcrImages <RepoName> <Subscription> <ResourceGroup> <RegistryName>
  <snip>
  waitForMcrDocIngestion <CommitDigest>
  waitForMcrImageIngestion <ImageInfoPath>
  generateSigningPayloads <ImageInfoPath> <PayloadOutputDirectory>
```

lbussell:
> For some reason when running ImageBuilder's help command, this command shows up at the bottom while the others are in alphabetical order:

Oh, it must be because I put it in a separate namespace...

src/Microsoft.DotNet.ImageBuilder/src/FileHelper.cs
```diff
@@ -38,5 +38,19 @@ public static void ForceDeleteDirectory(string path)
...
+        {
+            if (!Directory.Exists(path))
+            {
+                Directory.CreateDirectory(path);
+                return;
+            }
+
+            if (Directory.EnumerateFileSystemEntries(path).Any())
+            {
+                throw new IOException($"Directory '{path}' already exists and is not empty");
```
mthalman, line 52: I wouldn't expect this to be the default behavior of a method named `CreateDirectoryIfNotExists`. Consider adding a parameter named `throwIfNotEmpty` to enable this behavior.

src/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
```diff
@@ -90,6 +104,21 @@ public static ImageArtifactDetails LoadFromFile(string path, ManifestInfo manife
             return LoadFromContent(File.ReadAllText(path), manifest, skipManifestValidation, useFilteredManifest);
         }
 
+        /// <summary>
+        /// Loads an image info file as a parsed model directly with no validation or filtering.
+        /// </summary>
+        /// <param name="path">Path to the image info file.</param>
+        /// <exception cref="InvalidDataException"/>
+        public static ImageArtifactDetails LoadFromFile(string path, RegistryOverrideOptions registryOverrideOptions)
```
mthalman, line 112: I feel like this should be named differently than the others to differentiate that it's not correlating with the manifest content.

What do you think of just naming this `DeserializeImageArtifactDetails` and also remove the call to `ApplyRegistryOverride` to have that as opt-in behavior?

src/Microsoft.DotNet.ImageBuilder/src/Models/Oci/Manifest.cs
```diff
@@ -0,0 +1,25 @@
...
+    public string ArtifactType { get; init; } = string.Empty;
+
+    public string Reference { get; init; } = string.Empty;
+
+    public Dictionary<string, string> Annotations { get; init; } = [];
+
+    public static Manifest FromJson(string json)
+    {
+        return JsonConvert.DeserializeObject<Manifest>(json) ??
+            throw new InvalidOperationException("Unable to deserialize manifest");
```
mthalman, line 23: Is there a specific reason why Newtonsoft is being used here while your other model types use STJ?

src/Microsoft.DotNet.ImageBuilder/tests/Helpers/ImageInfoHelper.cs
```diff
@@ -4,13 +4,113 @@
...
     {
+        public static string WriteImageInfoToDisk(ImageArtifactDetails imageInfo, string directory)
+        {
+            string imageInfoJson = JsonConvert.SerializeObject(imageInfo);
+            string imageInfoPath = Path.Combine(directory, "image-info.json");
+            File.WriteAllText(imageInfoPath, imageInfoJson);
+            return imageInfoPath;
+        }
+
+        public static (ImageArtifactDetails, string) CreateImageInfoOnDisk(
```
mthalman, line 26:
```suggestion
        public static (ImageArtifactDetails ImageInfo, string FilePath) CreateImageInfoOnDisk(
```

src/Microsoft.DotNet.ImageBuilder/tests/Helpers/ImageInfoHelper.cs
```diff
@@ -46,5 +146,30 @@ public static PlatformData CreatePlatform(
...
         }
+
+        private static List<string> GetSharedTags(string productVersion, string os)
+        {
+            return
+            [
+                productVersion,
+                $"{productVersion}-{os}"
+            ];
+        }
```
mthalman, line 157:
```suggestion
        private static List<string> GetSharedTags(string productVersion, string os) =>
            [
                productVersion,
                $"{productVersion}-{os}"
            ];
```

src/Microsoft.DotNet.ImageBuilder/src/Models/Oci/Manifest.cs
lbussell: I adapted this directly from the existing Manifest code - just moved the type to a different namespace. I can switch it to STJ since it shouldn't make a difference here.

src/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
lbussell: That makes sense to me. I made this change.

src/Microsoft.DotNet.ImageBuilder/tests/Helpers/ImageInfoHelper.cs
lbussell: Just removed this whole method since it wasn't used.

mthalman (approved):

PR #1384, opened 2024-08-09
Clean EOL images

mthalman:
Updates the cleanup pipeline to target EOL images.

Related to https://github.com/dotnet/docker-tools/issues/1201

### DIFF
--- a/eng/pipelines/cleanup-acr-images.yml
+++ b/eng/pipelines/cleanup-acr-images.yml
@@ -30,12 +30,12 @@ jobs:
   - template: ../common/templates/steps/clean-acr-images.yml
     parameters:
       internalProjectName: ${{ variables.internalProjectName }}
-      repo: "public/dotnet/*nightly/*"
+      repo: "public/*"
       subscription: $(acr.subscription)
       resourceGroup: $(acr.resourceGroup)
       acr: $(acr.server)
-      action: pruneDangling
-      age: 30
+      action: pruneEol
+      age: 15
   - template: ../common/templates/steps/clean-acr-images.yml
     parameters:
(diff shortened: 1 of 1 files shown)


eng/pipelines/cleanup-acr-images.yml
```diff
@@ -30,12 +30,12 @@ jobs:
   - template: ../common/templates/steps/clean-acr-images.yml
     parameters:
       internalProjectName: ${{ variables.internalProjectName }}
-      repo: "public/dotnet/*nightly/*"
+      repo: "public/*"
       subscription: $(acr.subscription)
       resourceGroup: $(acr.resourceGroup)
       acr: $(acr.server)
-      action: pruneDangling
```
lbussell, line 37: Do we have any other plans to keep track of dangling images? I understand that they theoretically shouldn't exist after the EOL changes, but are we confident there won't be any more in the future?

lbussell (approved):
Just one question.

mthalman:
I've changed the configuration to be more targeted. This will be a multi-stage process of updating this pipeline as we move forward with annotating the historical images. It will eventually end up targeting all `public/*` repos. But we don't want that yet as it's going to be very inefficient by scanning a bunch or repos that won't end up getting cleaned in them. Since the oras tool is getting called for each manifest in the repo, it would be very wasteful and time consuming.

eng/pipelines/cleanup-acr-images.yml
mthalman: There are no plans to have that baked into a pipeline anywhere. The intention is that this would be manually verified as part of the rollout of the annotation work.

lbussell (approved):

Issue #1392, opened 2024-08-12
Update Image Builder container image to Full Mariner


ellahathaway:
Related to epic https://github.com/dotnet/docker-tools/issues/1060

The current Image Builder Dockerfile should be updated to be based first on full Mariner. This is the first step in completing https://github.com/dotnet/docker-tools/issues/1060 and is necessary so that we can find any issues and have a more iteration set of changes before moving to distroless Mariner.

lbussell:
I updated Mariner -> Azure Linux

PR #1393, opened 2024-08-12
Use azure linux for Image Builder container image

ellahathaway:
Closes https://github.com/dotnet/docker-tools/issues/1392

[Test build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2515837&view=results) (internal Microsoft link)

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
+++ b/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
@@ -3,13 +3,18 @@
 #     docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v <local path to build>:/repo -w /repo image-builder <image-build args>
 
 # build Microsoft.DotNet.ImageBuilder
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-alpine AS build-env
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/nightly/sdk:8.0-azurelinux3.0 AS build-env
 ARG TARGETARCH
+
+# install tooling
+RUN tdnf install -y \
+      tar \
+    && tdnf clean all
 
 # download oras package tarball
 WORKDIR /
 RUN oras_version=1.2.0 \
-    && wget https://github.com/oras-project/oras/releases/download/v${oras_version}/oras_${oras_version}_linux_${TARGETARCH}.tar.gz -O oras_linux.tar.gz \
+    && curl -sSL https://github.com/oras-project/oras/releases/download/v${oras_version}/oras_${oras_version}_linux_${TARGETARCH}.tar.gz -o oras_linux.tar.gz \
     && mkdir -p oras-install/ \
     && tar -zxf oras_linux.tar.gz -C oras-install/ \
     && rm -rf oras_linux.tar.gz
@@ -19,20 +24,23 @@ WORKDIR /image-builder
 # restore packages before copying entire source - provides optimizations when rebuilding
 COPY NuGet.config ./
 COPY src/Microsoft.DotNet.ImageBuilder.csproj ./src/
-RUN dotnet restore -r linux-musl-$TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj
+RUN dotnet restore -r linux-$TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj
 
 # copy everything else and publish
 COPY . ./
-RUN dotnet publish -r linux-musl-$TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj --self-contained=true --no-restore -o out
+RUN dotnet publish -r linux-$TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj --self-contained=true --no-restore -o out
 
 
 # build runtime image
-FROM mcr.microsoft.com/dotnet/runtime-deps:8.0-alpine
+FROM mcr.microsoft.com/dotnet/nightly/runtime-deps:8.0-azurelinux3.0
 
 # install tooling
-RUN apk add --no-cache \
-        docker \
-        git
+RUN tdnf install -y \
+      moby-engine \
+      moby-cli \
+      moby-buildx \
+      git \
+    && tdnf clean all
 
 # install oras tool
 COPY --from=build-env ["/oras-install/oras", "/usr/local/bin"]
--- a/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
+++ b/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
@@ -3,7 +3,7 @@
 #     docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v <local path to build>:/repo -w /repo image-builder <image-build args>
 
 # build Microsoft.DotNet.ImageBuilder
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/nightly/sdk:8.0-azurelinux3.0 AS build-env
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-azurelinux3.0 AS build-env
 ARG TARGETARCH
 
 # install tooling
@@ -32,7 +32,7 @@ RUN dotnet publish -r linux-$TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.cspr
 
 
 # build runtime image
-FROM mcr.microsoft.com/dotnet/nightly/runtime-deps:8.0-azurelinux3.0
+FROM mcr.microsoft.com/dotnet/runtime-deps:8.0-azurelinux3.0
 
 # install tooling
 RUN tdnf install -y \
--- a/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
+++ b/src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
@@ -6,15 +6,10 @@
 FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-azurelinux3.0 AS build-env
 ARG TARGETARCH
 
-# install tooling
-RUN tdnf install -y \
-      tar \
-    && tdnf clean all
-
 # download oras package tarball
 WORKDIR /
 RUN oras_version=1.2.0 \
-    && curl -sSL https://github.com/oras-project/oras/releases/download/v${oras_version}/oras_${oras_version}_linux_${TARGETARCH}.tar.gz -o oras_linux.tar.gz \
+    && curl -fSL --output oras_linux.tar.gz https://github.com/oras-project/oras/releases/download/v${oras_version}/oras_${oras_version}_linux_${TARGETARCH}.tar.gz \
     && mkdir -p oras-install/ \
     && tar -zxf oras_linux.tar.gz -C oras-install/ \
     && rm -rf oras_linux.tar.gz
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -3,13 +3,25 @@
 #     docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v <local path to build>:/repo -w /repo image-builder <image-build args>
 
 # build Microsoft.DotNet.ImageBuilder
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-alpine AS build-env
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/runtime:9.0-preview-azurelinux3.0 AS build-env
```
MichaelSimons, line 6: I don't think we should be using 9.0 during previews.  Perhaps once it reaches RC, but not before.

mthalman, line 6: > Because there are no distroless azure linux sdk images and the goal is ultimately to make this a distroless image, I chose to use the full azure linux image from runtime rather than the full azure linux image from sdk.

The ultimate goal is to use a distroless image for the _final_ image. We can use whatever we need for the intermediate stages, such as this one. So we should continue to use the `sdk` image for this stage.

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -19,20 +31,26 @@ WORKDIR /image-builder
...
 
 # install tooling
-RUN apk add --no-cache \
-        docker \
-        git
+RUN tdnf install -y \
+    moby-engine \
+    moby-cli \
+    moby-buildx \
+    git \
```
mthalman, line 49:
```suggestion
      moby-engine \
      moby-cli \
      moby-buildx \
      git \
```

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -19,20 +31,26 @@ WORKDIR /image-builder
...
+    && tdnf clean all
+
+# set docker api version
+ENV DOCKER_API_VERSION=1.43
```
mthalman, line 53: > There were some weird issues with the docker api version when I ran ImageBuilder in CI. I found the setting the ENV variable explicitly was the best fix. I'm looking into other solutions/seeing if this is even still needed after installing another necessary package.

If this is still necessary, it'd be good to document more about the need for it. What is the original issue that prompted the need for it? Why does this resolve it?

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -3,13 +3,25 @@
...
+    tar \
+    && tdnf clean all
+
+# install dotnet sdk
+RUN curl -sSL https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh \
+    && chmod +x ./dotnet-install.sh \
+    && ./dotnet-install.sh --channel 9.0 \
+    && rm -rf /dotnet-install.sh \
+    && ln -s /root/.dotnet/dotnet /usr/local/bin/dotnet
+
```
lbussell, line 20: It's expected to use the non-distroless SDK image, even with a distroless Runtime image, since the only outputs from the SDK build layer are DLLs/executables. You can see our [sample distroless Ubuntu Dockerfile](https://github.com/dotnet/dotnet-docker/blob/248742f8a701def93947ec1d2acdbccc4540170d/samples/aspnetapp/Dockerfile.chiseled) for reference. The images you most likely want to use for testing are:

- `mcr.microsoft.com/dotnet/nightly/sdk:8.0-azurelinux3.0`
- `mcr.microsoft.com/dotnet/nightly/runtime:8.0-azurelinux3.0`

lbussell (commented):
We should block this on https://github.com/dotnet/dotnet-docker/issues/5467, which should be completed today or tomorrow. Then, you'll be able to use the non-nightly images.

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
ellahathaway: This was something that came up when I was running the image builder image locally while trying to investigate failures in CI; the error message was complaining about the API version being too low (CI just showed red but didn't display this particular error). Turns out this local error was just a red herring because when I installed another moby package, the errors resolved in CI.

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
ellahathaway: I didn't see any azure linux images in main except for the 9.0-preview ones, so that's what led me to using those. But I see now that the non-preview images are in nightly and are expected to be added to main with https://github.com/dotnet/dotnet-docker/pull/5789. For now, I'll use the 8.0 nightly images and then switch to main once that PR gets merged.

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -3,13 +3,18 @@
 #     docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v <local path to build>:/repo -w /repo image-builder <image-build args>
 
 # build Microsoft.DotNet.ImageBuilder
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-alpine AS build-env
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-azurelinux3.0 AS build-env
 ARG TARGETARCH
 
+# install tooling
+RUN tdnf install -y \
+      tar \
```
lbussell, line 11: `tar` is installed by default in the SDK image:

https://github.com/dotnet/dotnet-docker/blob/a46d52ee1798fb37e050f52ca3fc802455ca229d/src/sdk/8.0/azurelinux3.0/amd64/Dockerfile#L18-L22

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -3,13 +3,18 @@
...
+    && tdnf clean all
+
 # download oras package tarball
 WORKDIR /
 RUN oras_version=1.2.0 \
-    && wget https://github.com/oras-project/oras/releases/download/v${oras_version}/oras_${oras_version}_linux_${TARGETARCH}.tar.gz -O oras_linux.tar.gz \
+    && curl -sSL https://github.com/oras-project/oras/releases/download/v${oras_version}/oras_${oras_version}_linux_${TARGETARCH}.tar.gz -o oras_linux.tar.gz \
```
lbussell, line 17: What's the reasoning for using `-sSL` instead of `-fSL` like our other images? https://github.com/dotnet/dotnet-docker/blob/a46d52ee1798fb37e050f52ca3fc802455ca229d/src/sdk/8.0/azurelinux3.0/amd64/Dockerfile#L25

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -19,20 +24,23 @@ WORKDIR /image-builder
 # restore packages before copying entire source - provides optimizations when rebuilding
 COPY NuGet.config ./
 COPY src/Microsoft.DotNet.ImageBuilder.csproj ./src/
-RUN dotnet restore -r linux-musl-$TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj
+RUN dotnet restore -r linux-$TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj
```
lbussell, line 27:
```suggestion
RUN dotnet restore -a $TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj
```

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -19,20 +24,23 @@ WORKDIR /image-builder
...
 
 # copy everything else and publish
 COPY . ./
-RUN dotnet publish -r linux-musl-$TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj --self-contained=true --no-restore -o out
+RUN dotnet publish -r linux-$TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj --self-contained=true --no-restore -o out
```
lbussell, line 31:
```suggestion
RUN dotnet publish -a $TARGETARCH ./src/Microsoft.DotNet.ImageBuilder.csproj --self-contained=true --no-restore -o out
```

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
```diff
@@ -19,20 +24,23 @@ WORKDIR /image-builder
...
+FROM mcr.microsoft.com/dotnet/runtime-deps:8.0-azurelinux3.0
 
 # install tooling
-RUN apk add --no-cache \
-        docker \
-        git
+RUN tdnf install -y \
+      moby-engine \
+      moby-cli \
+      moby-buildx \
```
lbussell, line 41: Please use `docker-*` versions where available. These packages were renamed: https://github.com/microsoft/azurelinux/blob/3.0/SPECS/docker-cli/docker-cli.spec

```
PS> docker run -it --rm mcr.microsoft.com/dotnet/sdk:8.0-azurelinux3.0 tdnf search docker
...
docker-buildx : A Docker CLI plugin for extended build capabilities with BuildKit
docker-cli : The open-source application container engine client.
moby-engine : The open-source application container engine
```

lbussell (changes requested):

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
ellahathaway: No reasoning - I'm just familiar with -sSL and didn't realize that we use -fSL in other areas. I'll update my PR with -fSL.

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
ellahathaway: Will do - looks like I can replace `moby-cli` and `moby-buildx`.

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
ellahathaway:
```
=> ERROR [build-env 8/9] RUN dotnet publish -a amd64 ./src/Microsoft.DotNet.ImageBuilder.csproj --self-contained=true --no-restore -o out                                                                       1.8s
------
 > [build-env 8/9] RUN dotnet publish -a amd64 ./src/Microsoft.DotNet.ImageBuilder.csproj --self-contained=true --no-restore -o out:
1.707 /usr/share/dotnet/sdk/8.0.400/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(491,5): error NETSDK1112: The runtime pack for Microsoft.NETCore.App.Runtime.linux-x64 was not downloaded. Try running a NuGet restore with the RuntimeIdentifier 'linux-x64'. [/image-builder/src/Microsoft.DotNet.ImageBuilder.csproj]
```

Looks like the runtime identifier is required.

src/Microsoft.DotNet.ImageBuilder/Dockerfile.linux
ellahathaway: See https://github.com/dotnet/docker-tools/pull/1393#discussion_r1717600858

mthalman (approved):

lbussell (approved):

PR #1397, opened 2024-08-14
Add custom steps to run after build

ellahathaway:
Needed for changes in https://github.com/dotnet/dotnet-docker/pull/5797 which runs `verify-secret-usage.yml` after building in CI and PRs.

For background, according to https://github.com/dotnet/dnceng/tree/main/src/SecretManager/Microsoft.DncEng.SecretManager#onboarding-a-new-repo, there are a series of steps which need to run after building in CI and PRs. In order to run these steps, the pipeline templates that we use need to be updated to accommodate for running custom steps in the build Stage after the build step.

Steps to do before publishing this PR:
- Test these changes via a pipeline run(s)

### DIFF
--- a/eng/common/templates/jobs/build-images.yml
+++ b/eng/common/templates/jobs/build-images.yml
@@ -5,6 +5,7 @@ parameters:
   dockerClientOS: null
   buildJobTimeout: 60
   customInitSteps: []
+  customPostSteps: []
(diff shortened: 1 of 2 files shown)


ellahathaway:
Closing since I don't this is no longer needed given what we're using secret manager for.

PR #1401, opened 2024-08-15
Use concurrency limiter when deleting repo manifests

mthalman:
This change adjusts how many delete requests that are sent to the ACR when cleaning up manifests. This avoids potential issues that have been seen when sending several thousand delete requests at once. It makes use of the concurrency limiter from Polly. That lets us queue up all our delete tasks at once and then it will only allow a specific number of them to execute concurrently.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
@@ -11,10 +11,10 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Azure.Containers.ContainerRegistry;
-using Microsoft.DotNet.ImageBuilder.Models.Annotations;
 using Microsoft.DotNet.ImageBuilder.Models.Oras;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesCommand.cs
```diff
@@ -189,10 +191,18 @@ await Parallel.ForEachAsync(allManifests, async (manifest, token) =>
...
+
+            List<Task> tasks = [];
             foreach (ArtifactManifestProperties manifest in manifests)
             {
-                tasks.Add(DeleteManifestAsync(acrContentClient, deletedImages, repository, manifest));
+                tasks.Add(pipeline.ExecuteAsync(async cancellationToken =>
+                {
+                    await DeleteManifestAsync(acrContentClient, deletedImages, repository, manifest);
+                }).AsTask());
             }
```
lbussell, line 206: This could be defined inline:
```cs
            IEnumerable<Task> tasks = manifests.Select(manifest => pipeline
                    .ExecuteAsync(async cancellationToken => 
                        await DeleteManifestAsync(acrContentClient, deletedImages, repository, manifest))
                    .AsTask());
```

lbussell (approved):

PR #1402, opened 2024-08-15
Fix file-pusher pipeline failure when pushing files that contain `##vso` tasks

### LINKED ISSUE #1388: push-common-updates pipeline failing with `BUILD_BUILDNUMBER` error
```
##[error]TF209010: The build number format $env:BUILD_BUILDNUMBER (Dry run)\"\n        Write-Host \"##vso[build.addbuildtag]dry-run\"\n        \n        if (\"$(officialBranches)\".Split(',').Contains(\"$(Build.SourceBranch)\"))\n        {\n          Write-Host \"##vso[task.logissue type=error]Cannot run a CG dry-run build from an official branch ($(officialBranches)).\"\n          Write-Host \"##vso[task.logissue type=error]Run the pipeline again from a different branch to avoid registering scan results.\"\n          exit 1\n        }\n      displayName: Update Build Number\n    - task: ComponentGovernanceComponentDetection@0\n      displayName: Component Detection (manually injected)\n      inputs:\n        # Running CG with `whatIf: true` or `scanType: LogOnly` outputs a list of detected components, but doesn't show\n        # which components would trigger an alert on a production build.\n        # As long as the build isn't ran on a tracked branch (typically main or nightly), it's OK to submit components\n        # to CG for analysis. Only the results for tracked branches matter.\n        scanType: Register\n        whatIf: false\n        alertWarningLevel: Low\n        failOnAlert: false\n        ignoreDirectories: $(Build.SourcesDirectory)/versions\n        showAlertLink: true\n        timeoutInMinutes: 10\n" contains invalid character(s), is too long, or ends with '.'. The maximum length of a build number is 255 characters. Characters which are not allowed include '"', '/', ':', '<', '>', '\', '|', '?', '@', and '*'.
```

[Example build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2512285&view=results) (internal link)

This is a regression from #1371

lbussell:
Fixes #1388

### DIFF
--- a/eng/src/file-pusher/AzDoSafeTraceListenerWrapper.cs
+++ b/eng/src/file-pusher/AzDoSafeTraceListenerWrapper.cs
@@ -1,0 +1,27 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+
+namespace FilePusher;
+
+public class AzDoSafeTraceListenerWrapper(TraceListener innerTraceListener) : TraceListener
+{
+    private readonly TraceListener _innerTraceListener = innerTraceListener;
+
+    public override void Write(string? message)
+    {
+        _innerTraceListener.Write(EscapeVsoDirectives(message));
+    }
+
+    public override void WriteLine(string? message)
+    {
+        _innerTraceListener.WriteLine(EscapeVsoDirectives(message));
+    }
+
+    private static string? EscapeVsoDirectives(string? message)
+    {
+        return message?.Replace("##vso", "#VSO_DIRECTIVE");
+    }
+}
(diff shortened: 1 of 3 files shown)


eng/src/file-pusher/AzDoSafeTraceListenerWrapper.cs
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public override void WriteLine(string? message)
+    {
+        _innerTraceListener.WriteLine(EscapeVsoDirectives(message));
+    }
+
+    private static string? EscapeVsoDirectives(string? message)
+    {
+        return message?.Replace("##vso", "#VSO_DIRECTIVE");
```
mthalman, line 25: Provide a link to https://github.com/dotnet/docker-tools/issues/1388 to make it easier to trace back to why this is needed. A comment explaining why would be good too.

lbussell: Done in https://github.com/dotnet/docker-tools/pull/1402/commits/3921909b087543679c377890af645dc3aebf4ba7

mthalman (approved):

Issue #1227, opened 2024-03-19
Factor out default build agent pools and image names into shared variables


lbussell:
Related: https://github.com/dotnet/dotnet-docker-internal/issues/4475

With the addition of 1ES pipeline templates, we have a few places where our default pool definitions are overridden or ignored. We should consolidate these into one set of variables for our pools and image names so we don't have any hardcoded values anywhere.

https://github.com/dotnet/docker-tools/blob/5a5564806a23862c5a8416477a59c89b30b66b4d/eng/common/templates/stages/dotnet/build-test-publish-repo.yml#L64-L73

https://github.com/dotnet/docker-tools/blob/5a5564806a23862c5a8416477a59c89b30b66b4d/eng/common/templates/variables/common.yml#L46-L57

https://github.com/dotnet/docker-tools/blob/5a5564806a23862c5a8416477a59c89b30b66b4d/eng/pipelines/dotnet-buildtools-image-builder-official.yml#L21-L27

https://github.com/lbussell/dotnet-docker/blob/9482b39dbd103f56ce42b62cbb14bda00f223c40/eng/pipelines/cg-images.yml#L25-L36

lbussell:
1. The `default*PoolImage` variables should stay .NET-independent, and we can continue to use the `/dotnet/build-test-publish-repo.yml` file to define the 1ES PT compliant images. We should however define variables for the SDL scanning pool so that they can be shared between all of our pipelines: https://github.com/dotnet/docker-tools/blob/5a5564806a23862c5a8416477a59c89b30b66b4d/eng/pipelines/dotnet-buildtools-image-builder-official.yml#L21-L27

lbussell:
Marking as wave1 since this is 1ES PT follow-up.

ellahathaway:
I'm continuing to work on this issue. Hoping to have a few PRs out for it soon.

ellahathaway:
https://github.com/microsoft/dotnet-framework-docker/pull/1171 is the last PR needed to close out this issue.

PR #1406, opened 2024-08-16
Use variables for default build agent pools and image names

ellahathaway:
Related to https://github.com/dotnet/docker-tools/issues/1227

[Image builder pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2520011&view=results)
[Pipeline run using 1es-unofficial template](https://dev.azure.com/dnceng/internal/_build/results?buildId=2520188&view=results)
[Pipeline run using 1es-official template](https://dev.azure.com/dnceng/internal/_build/results?buildId=2520191&view=logs&s=aeb66100-48e6-58f7-deed-7d3dc5316e8a)


eng/common/templates/stages/dotnet/build-test-publish-repo.yml
```diff
@@ -70,9 +70,9 @@ stages:
...
       ${{ elseif eq(variables['System.TeamProject'], parameters.internalProjectName) }}:
-        name: NetCore1ESPool-Internal
-        image: 1es-ubuntu-2204
-        os: linux
+        name: $(defaultInternalPoolName)
+        image: $(defaultInternalLinuxPoolImage)
+        os: $(defaultInternalLinuxOS)
 
     # Linux Arm64
     linuxArm64Pool:
```
lbussell, line 78: The rest of the hard-coded pool names and images in this file should also be put into variables.

The `default` qualifier seems redundant - these are variables meant to override the defaults set [here](https://github.com/ellahathaway/docker-tools/blob/62d276d0f338460f27f3bece9293888e2d6798a1/eng/common/templates/stages/build-test-publish-repo.yml#L28-L39). May I suggest a format? Something like:

```
linuxArm64InternalPoolName
linuxArm64InternalPoolImage
linuxArm64PublicPoolName
linuxArm64PublicPoolImage
...
```

eng/common/templates/variables/common.yml
```diff
@@ -56,6 +56,17 @@ variables:
...
+- name: defaultInternalPoolName
+  value: NetCore1ESPool-Internal
+- name: defaultInternalLinuxPoolImage
+  value: 1es-ubuntu-2204
+- name: defaultInternalWindowsPoolImage
+  value: 1es-windows-2022
+- name: defaultInternalLinuxOS
+  value: linux
+- name: defaultInternalWindowsOS
+  value: windows
```
lbussell, line 68: `defaultInternalLinuxOS` and `defaultInternalWindowsOS` should be OK to hard-code. I don't think those will change.

The dotnet-specific pool variables should go under variables/dotnet/common.yml, since the other (golang) team that uses these templates may want to choose different images.

The default pools for the non-dotnet-specific `1es-official` and `1es-unofficial` templates can remain in this file.

eng/common/templates/1es-official.yml
```diff
@@ -20,15 +20,15 @@ parameters:
...
+    os: $(defaultInternalLinuxOS)
 - name: sourceAnalysisPool
   type: object
   default:
-    name: NetCore1ESPool-Internal
-    image: 1es-windows-2022
-    os: windows
+    name: $(defaultInternalPoolName)
+    image: $(defaultInternalWindowsPoolImage)
+    os: $(defaultInternalWindowsOS)
```
lbussell, line 31: Consider making the source analysis pool a separate set of variables.

eng/common/templates/1es-official.yml
```diff
@@ -20,15 +20,15 @@ parameters:
 - name: pool
```
lbussell, line 20: I noticed that the `1es-official` template has separate `pool` and `sourceAnalysisPool` parameters while `1es-unofficial` does not. Can we make them symmetrical?

eng/common/templates/stages/dotnet/build-test-publish-repo.yml
ellahathaway: In https://github.com/dotnet/docker-tools/issues/1227#issuecomment-2013239789, you said "we can continue to use the `/dotnet/build-test-publish-repo.yml` file to define the 1ES PT compliant images...We should however define variables for the SDL scanning pool so that they can be shared between all of our pipelines." I took this to mean that we only wanted to factor out the SDL scanning pool (eg `NetCore1ESPool-Internal`)

Am I interpreting your comment incorrectly?

ellahathaway: Spoke offline - will follow up with fix.

eng/common/templates/stages/dotnet/build-test-publish-repo.yml
```diff
@@ -70,47 +70,47 @@ stages:
...
+        image: $(windowsServer2016InternalPoolImage)
 
     # Windows Server 2019 (1809)
     windows1809Pool:
       os: windows
-      name: Docker-1809-${{ variables['System.TeamProject'] }}
+      name: $(windowsServer2019PoolName)
       ${{ if eq(variables['System.TeamProject'], parameters.internalProjectName) }}:
-        image: Server2019-1809-NESDockerBuilds-1ESPT
+        image: $(windowsServer2019InternalPoolImage)
```
lbussell, line 109: It would be good to specify the public image as well (rather than relying on the pool's default). You can see which image the pool is resolving by default here, for example: https://dev.azure.com/dnceng-public/public/_build/results?buildId=781079&view=logs&j=21ee7c81-5e88-5422-0974-23d58b004dc6&t=dbda5bbe-9974-44b4-8b6a-b1aaafa0af22&l=7

eng/pipelines/dotnet-buildtools-image-builder-official.yml
```diff
@@ -10,6 +10,7 @@ pr: none
 
 variables:
 - template: /eng/pipelines/templates/variables/image-builder.yml@self
+- template: /eng/pipelines/templates/variables/common.yml@self
```
lbussell, line 13: This may have other side effects... we should put the common variables above image-builder.yml, or reference common at the top of image-builder.yml, in case we need to override any of the variables.

ellahathaway: This ended up being redundant since the other variable template eventually leads to the common variables.

eng/common/templates/1es-official.yml
```diff
@@ -20,14 +20,14 @@ parameters:
...
   type: object
   default:
-    name: NetCore1ESPool-Internal
-    image: 1es-ubuntu-2204
+    name: $(default1ESInternalPoolName)
```
ellahathaway, line 23: One thing I will say: I worry about repercussions that this might have on other repos. I suspect that a lot of the docker repos make use of this template (and the unofficial one), and not all of the pipelines may import the template where these variables are defined.

And because this template is used by other templates, I can't define variables it in. If I could, I would just import the common template in this yml, but that's not currently possible with Azdo pipeline templates.

ellahathaway: Spoke with Matt offline about this. The conclusion was to add these variables to the common vars template, and as these changes roll out to other repos, I will go in and open PRs to import the variables template where necessary. Most pipelines utilize other vars templates that import the common variables, but there may be a few pipelines that will require changes to add the common template.

eng/common/templates/variables/common.yml
```diff
@@ -56,6 +56,16 @@ variables:
...
+- name: default1ESInternalPoolName
+  value: NetCore1ESPool-Internal
+- name: default1ESInternalPoolImage
+  value: 1es-ubuntu-2204
+
+- name: defaultSourceAnalysisPoolName
+  value: NetCore1ESPool-Internal
+- name: defaultSourceAnalysisPoolImage
+  value: 1es-windows-2022
+
```
lbussell, line 68: FYI @dagood these new variables are being added to the common template - I don't think they conflict with any of yours but just checking in case they do.

dagood: Thanks, looks fine. We have the source analysis pool name/image directly in the pipeline for now.

(FWIW, we use `NetCore1ESPool-Svc-Internal`, but we only produce images for release branches, so it makes sense to me that the default here is `NetCore1ESPool-Internal`.)

lbussell (approved):

PR #1408, opened 2024-08-21
Use repo prefix for EOL data logic

mthalman:
There are a couple issues with the logic in `GenerateEolAnnotationDataCommand` related to image names that prevent it from working properly in the production environment:
* When computing the repo names to process, it doesn't prepend the repo prefix specified in the command options. This prevents it from querying the correct repos to find images that need to be annotated.
* The image digest names in the image info file are based on their public name (e.g. mcr.microsoft.com). These values need to be transformed to the ACR name (with repo prefix) in order to properly match against the names in the ACR.


lbussell:
I've just solved this same problem in #1381 - you may consider checking out/reviewing some of that code if it might be applicable: https://github.com/dotnet/docker-tools/commit/17c2fa8fd64a18eca74431e8338e74a62f1fa7ba

lbussell (approved):

Issue #1267, opened 2024-05-03
Remove project name parameters from pipeline templates


mthalman:
This PR (https://github.com/dotnet/docker-tools/pull/884) introduce a template pattern where `internalProjectName` and `publicProjectName` parameters were defined and passed around to templates rather than hardcoding the actual AzDO project names:

https://github.com/dotnet/docker-tools/blob/86df1f66c815ed8bc35c50147228d003aadcd9b7/eng/common/templates/stages/build-test-publish-repo.yml#L22-L23

This pattern doesn't provide much value. In fact, it has negative value because developers need to ensure that these values get passed around through the template graph. Omitting this can lead to unintended behavior. We should consider removing these parameters and simply hardcoding the project names where they are needed.

lbussell:
[Triage] This is something we can work on incrementally - as we make changes to existing pipelines we should feel free to replace these variables where appropriate.

PR #1413, opened 2024-08-22
Push EOL annotations in publish stage

mthalman:

Updates the publish stage of the common pipeline infrastructure to support pushing EOL annotations.

This is opt-in behavior for consuming pipelines. A pipeline will need to set the `publishEolAnnotations` variable to `true` to opt in.

### DIFF
--- a/eng/common/templates/jobs/publish.yml
+++ b/eng/common/templates/jobs/publish.yml
@@ -147,7 +147,7 @@ jobs:
     parameters:
       displayName: Generate EOL Annotation Data
       serviceConnection: $(publish.serviceConnectionName)
-      internalProjectName: ${{ parameters.internalProjectName }}
+      internalProjectName: internal
       condition: and(succeeded(), eq(variables['publishEolAnnotations'], 'true'))
       args: >
         generateEolAnnotationData
@@ -163,8 +163,8 @@ jobs:
       path: $(Build.ArtifactStagingDirectory)/eol-annotation-data
       artifactName: eol-annotation-data
       displayName: Publish EOL Annotation Data Artifact
-      internalProjectName: ${{ parameters.internalProjectName }}
-      publicProjectName: ${{ parameters.publicProjectName }}
+      internalProjectName: internal
+      publicProjectName: public
       condition: and(succeeded(), eq(variables['publishEolAnnotations'], 'true'))
   - template: /eng/common/templates/steps/annotate-eol-digests.yml@self
     parameters:
--- a/eng/common/templates/jobs/publish.yml
+++ b/eng/common/templates/jobs/publish.yml
@@ -7,6 +7,7 @@ parameters:
 jobs:
 - job: Publish
   pool: ${{ parameters.pool }}
+  timeoutInMinutes: 90
   variables:
   - name: imageBuilder.commonCmdArgs
     value: >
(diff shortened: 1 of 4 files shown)


eng/common/templates/jobs/publish.yml
```diff
@@ -7,6 +7,7 @@ parameters:
 jobs:
 - job: Publish
   pool: ${{ parameters.pool }}
+  timeoutInMinutes: 90
```
lbussell, line 10: What is the purpose of the timeout here? Wouldn't we want to put a timeout on individual legs instead? We wouldn't want things to get into an unknown state for example if we ever happened to time-out during an ACR import.

eng/common/templates/jobs/publish.yml
```diff
@@ -117,6 +124,8 @@ jobs:
       $(imageBuilder.commonCmdArgs)
     condition: and(succeeded(), eq(variables['publishImageInfo'], 'true'))
     displayName: Publish Image Info
+  - script: curl -fSL --output $(Build.ArtifactStagingDirectory)/eol-annotation-data/image-info-new.json $(imageInfoRawContentUrl)
```
lbussell, line 127: Just clarifying that the purpose of re-downloading this file is because we need to get the final version after it has been merged with the (incomplete) image-info file produced by the build?

eng/common/templates/jobs/publish.yml
```diff
@@ -134,6 +143,33 @@ jobs:
...
+        '$(acr.server)'
+        $(generateEolAnnotationDataExtraOptions)
+        $(dryRunArg)
+  - template: /eng/common/templates/steps/publish-artifact.yml@self
+    parameters:
+      path: $(Build.ArtifactStagingDirectory)/eol-annotation-data
+      artifactName: eol-annotation-data
+      displayName: Publish EOL Annotation Data Artifact
+      internalProjectName: ${{ parameters.internalProjectName }}
+      publicProjectName: ${{ parameters.publicProjectName }}
```
lbussell, line 167:
```suggestion
      internalProjectName: internal
      publicProjectName: public
```

eng/common/templates/jobs/publish.yml
```diff
@@ -134,6 +143,33 @@ jobs:
         --architecture '*'
         $(dryRunArg)
         $(imageBuilder.commonCmdArgs)
+  - template: /eng/common/templates/steps/run-imagebuilder.yml@self
+    parameters:
+      displayName: Generate EOL Annotation Data
+      serviceConnection: $(publish.serviceConnectionName)
+      internalProjectName: ${{ parameters.internalProjectName }}
```
lbussell, line 150:
```suggestion
      internalProjectName: internal
```

Feel free to disregard, but you could consider hard-coding these when adding new pipeline code. Related: https://github.com/dotnet/docker-tools/issues/1267

eng/common/templates/jobs/publish.yml
mthalman: What individual legs? There is just one job for publish. It has a default timeout of 60 minutes. I'm increasing it a bit to accommodate the extra time it'll take to handle all the images that currently exist in the repos as we transition to getting them cleaned up.

eng/common/templates/jobs/publish.yml
mthalman: Yes. Essentially we need to know the "before" and "after". This is really only used to determine if any repo has been deleted for a given publish. (Like if we decided to completely remove aspire-dashboard repo.)

NikolaMilosavljevic (approved):

eng/common/templates/jobs/publish.yml
lbussell: Sorry, I meant individual steps. I didn't realize there was a default 60 minute timeout. If that's the case then no problems increasing it.

lbussell (approved):

PR #1425, opened 2024-09-05
Defer Docker login until just before pulling cached images during build command

### LINKED ISSUE #1424: Auth error when pulling cached image
There's a [build in buildtools-prereqs](https://dev.azure.com/dnceng/internal/_build/results?buildId=2530258&view=logs&j=9aada89a-2e24-5acb-2807-6900158b9cce&t=6f3c41b4-0b5c-55eb-81af-364314aaa449) (internal link) that is failing with an unauthorized error when attempting to pull an image from the ACR. This scenario occurs when the build ends up building an image that takes a long time (in this case 3+ hrs) and then attempting to pull an image from the ACR due to a cache hit.

We probably need to change the logic so that it calls `IRegistryCredentialsProvider.ExecuteWithCredentialsAsync` for each Dockerfile that is processed rather than once for the whole batch.

lbussell:
Fixes #1424 

Instead of running `BuildImagesAsync()` under authentication the whole time, defer login to only when we need to pull cached images in the case of a cache hit. That should be the only place we need credentials during the actual building of images, since we already pulled the base images with `PullBaseImagesAsync()`.

This has the added benefit of running the rest of the build command without any Docker credentials, since they shouldn't be necessary (and should fail the build).

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/RegistryCredentialsProviderExtensions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/RegistryCredentialsProviderExtensions.cs
@@ -46,24 +46,16 @@ internal static class RegistryCredentialsProviderExtensions
         string registryName,
         string? ownedAcr)
     {
-        bool loggedIn = await LogInToRegistry(
-            credsProvider,
+        await credsProvider.ExecuteWithCredentialsAsync(
             isDryRun,
+            () => {
+                action();
+                return Task.CompletedTask;
+            },
             credentialsOptions,
             registryName,
-            ownedAcr);
-
-        try
-        {
-            action();
-        }
-        finally
-        {
-            if (loggedIn && !string.IsNullOrEmpty(registryName))
-            {
-                DockerHelper.Logout(registryName, isDryRun);
-            }
-        }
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/RegistryCredentialsProviderExtensions.cs
```diff
@@ -18,23 +18,44 @@ public static async Task ExecuteWithCredentialsAsync(
...
-            DockerHelper.Login(credentials, registryName, isDryRun);
-            loggedIn = true;
+            if (loggedIn && !string.IsNullOrEmpty(registryName))
+            {
+                DockerHelper.Logout(registryName, isDryRun);
+            }
         }
+    }
+
+    public static async Task ExecuteWithCredentialsAsync(
```
mthalman, line 41: This method's implementation can be simplified down to this:

```
await credsProvider.ExecuteWithCredentialsAsync(isDryRun,
  () =>
  {
      action();
      return Task.Completed;
  },
  credentialOptions,
  registryName,
  ownedAcr);
```

mthalman (approved):

PR #1429, opened 2024-09-10
Allow images to be excluded from cleaning

mthalman:
On rare occasions, there may be specific images that should be excluded from being cleaned by the `cleanAcrImages` command.

This adds a new `--exclude` option to the command that provides a way for the caller to provide an image name that will be excluded from being cleaned. Multiple `--exclude` options can be provided if necessary. This supports an image name that is referenced by tag or digest.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
@@ -19,6 +19,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         public string Subscription { get; set; }
         public string ResourceGroup { get; set; }
         public string RegistryName { get; set; }
+        public string[] ImagesToExclude { get; set; } = [];
     }
 
     public class CleanAcrImagesOptionsBuilder : CliOptionsBuilder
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/CleanAcrImagesOptions.cs
```diff
@@ -57,7 +58,9 @@ public override IEnumerable<Option> GetCliOptions() =>
                         CreateOption("action", nameof(CleanAcrImagesOptions.Action),
                             EnumHelper.GetHelpTextOptions(DefaultCleanAcrImagesAction), DefaultCleanAcrImagesAction),
                         CreateOption("age", nameof(CleanAcrImagesOptions.Age),
-                            $"Minimum age (days) of repo or images to be deleted (default: {DefaultAge})", DefaultAge)
+                            $"Minimum age (days) of repo or images to be deleted (default: {DefaultAge})", DefaultAge),
+                        CreateMultiOption<string>("exclude", nameof(CleanAcrImagesOptions.ImagesToExclude),
+                            $"Name of image to exclude from cleaning (does not apply when using the '{nameof(CleanAcrImagesAction.Delete)}' action)"),
```
lbussell, line 63: Consider throwing an error if we pass the `Delete` Action along with `ImagesToExclude`, if that's an invalid combination of options.

mthalman: fixed in https://github.com/dotnet/docker-tools/pull/1429/commits/e31f52c91399b61f1246ef779b4f9751a35d431f

lbussell (approved):

Issue #1433, opened 2024-09-11
EOL annotations are being applied to the newly supported images


mthalman:
After enabling EOL annotations in the dotnet-docker nightly branch, it was found that EOL annotations get applied to the images that are being newly published in that run. This happens because we're not getting the newly updated version of the image-info JSON file which defines what the currently supported images are. We're actually using the previous version of that file, which doesn't include the newly published images. So everything not in that file is considered to be EOL. Thus, the new images get EOL annotations applied to them.

The reason we're using the old version of the image info file is because the publish logic indicated below is not operating as expected:

https://github.com/dotnet/docker-tools/blob/597eb0f9628a77924f63c1e2683bf5f8cf314317/eng/common/templates/jobs/publish.yml#L109-L128

The intent is to do the following:
1. Download the current (old) version of the image info file from GitHub.
2. Run the Image Builder command which updates that file in GitHub.
3. Download the current (new) version of the image info file from GitHub.

The issue is that step 3 is still getting the old version of the file. I believe this is just a timing issue where use of the `raw.githubusercontent.com` URL is not retrieving the latest state of the file at the point we retrieve it. The commit has been made in the repo but `raw.githubusercontent.com` is not reflecting that state yet.

mthalman:
This could actually be caused by the GitHub server returning a cached version of the file. I will experiment to see if using a `Cache-Control` header on the request can avoid that.

mthalman:
Ok, there is significant latency with respect to raw.githubusercontent.com reflecting the latest content of the repo. I made a commit in a repo and then accessed that file via raw.githubusercontent.com, and it took several minutes for my change to show up there. This means raw.githubusercontent.com should not be used to download content for our pipeline as it does not reliably provide the latest content.

PR #1434, opened 2024-09-11
publishImageInfo: output original and updated content of image info

mthalman:
Contributes to https://github.com/dotnet/docker-tools/issues/1433

This will provide a way for the pipeline to get the content of both the original and updated versions of the image info file as part of the invocation of the `publishImageInfo` command. This will provide accurate content of the file that downloading from `raw.githubusercontent.com` does not provide.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoOptions.cs
@@ -4,7 +4,7 @@
 
 using System.Collections.Generic;
 using System.CommandLine;
-using System.Linq;
+using static Microsoft.DotNet.ImageBuilder.Commands.CliHelper;
 
 #nullable enable
 namespace Microsoft.DotNet.ImageBuilder.Commands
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoOptions.cs
```diff
@@ -4,27 +4,38 @@
...
 
 #nullable enable
 namespace Microsoft.DotNet.ImageBuilder.Commands
 {
     public class PublishImageInfoOptions : ImageInfoOptions, IGitOptionsHost
     {
         public GitOptions GitOptions { get; set; } = new GitOptions();
+
+        public string? OriginalImageInfoOutputPath { get; set; }
+        public string? UpdatedImageInfoOutputPath { get; set; }
```
lbussell, line 17: `ImageInfoOptions` has `ImageInfoPath`. Taking three image info paths as arguments is getting confusing so I'd like to make sure my interpretation is correct:

- `ImageInfoPath` should point to the build-specific image info file, AKA the one that contains only the images that were just built in a given pipeline run.
- `OriginalImageInfoOutputPath` will contain the file exactly as it is from GitHub? AKA we're going to remove the "Download Original Image Info File" step?
- `UpdatedImageInfoPath` will contain the new file that was pushed to GitHub, containing all the current links

If that's right, then the changes all look good to me - but the descriptions for the CLI options could be clearer.

mthalman: That's the correct interpretation. I'll include comments in the code to better clarify. The CLI descriptions should be kept fairly short for readability.

lbussell (approved):

PR #1435, opened 2024-09-12
Consume image info output of publishImageInfo command

mthalman:
Fixes https://github.com/dotnet/docker-tools/issues/1433

This uses the new options added in https://github.com/dotnet/docker-tools/pull/1434 to get the content of the image info file before and after it gets updated by the `publishImageInfo` command. This provides accurate content of these files that the existing pipeline steps which use `curl` to download the file do not. So those steps have been removed as well.

### DIFF
--- a/eng/common/templates/jobs/publish.yml
+++ b/eng/common/templates/jobs/publish.yml
@@ -106,10 +106,8 @@ jobs:
     parameters:
       dryRunArg: $(dryRunArg)
       condition: and(succeeded(), eq(variables['publishReadme'], 'true'))
-  - script: |
-      mkdir -p $(Build.ArtifactStagingDirectory)/eol-annotation-data
-      curl -fSL --output $(Build.ArtifactStagingDirectory)/eol-annotation-data/image-info-old.json $(imageInfoRawContentUrl)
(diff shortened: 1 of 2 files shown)


eng/common/templates/jobs/publish.yml
```diff
@@ -106,10 +106,8 @@ jobs:
     parameters:
       dryRunArg: $(dryRunArg)
       condition: and(succeeded(), eq(variables['publishReadme'], 'true'))
-  - script: |
-      mkdir -p $(Build.ArtifactStagingDirectory)/eol-annotation-data
-      curl -fSL --output $(Build.ArtifactStagingDirectory)/eol-annotation-data/image-info-old.json $(imageInfoRawContentUrl)
-    displayName: Download Original Image Info File
+  - script: mkdir -p $(Build.ArtifactStagingDirectory)/eol-annotation-data
+    displayName: Create EOL Annotation Data Directory
```
lbussell, line 110: Seems like the `imageInfoRawContentUrl` can be removed now?

mthalman: good catch. fixed

lbussell (approved):

PR #1436, opened 2024-09-12
Use pool vars in secret sync pipeline

ellahathaway:
Related to https://github.com/dotnet/docker-tools/issues/1227

I'm going back through the docker repos to ensure that there's no left-over image names and agent pools that need to use the shared variables.

[Pipeline run with changes](https://dev.azure.com/dnceng/internal/_build/results?buildId=2537265&view=logs&j=df0a3355-897e-53ae-288d-2f16ccab2822)

### DIFF
--- a/eng/pipelines/secret-management-weekly.yml
+++ b/eng/pipelines/secret-management-weekly.yml
@@ -8,15 +8,18 @@ schedules:
     - main
   always: true
 
+variables:
+- template: templates/variables/common.yml
+
 stages:
 - stage: SynchronizeSecrets
   jobs:
   - job: Synchronize
     displayName: Synchronize secrets
     pool:
-      name: NetCore1ESPool-Internal
-      demands: ImageOverride -equals 1es-windows-2019
-
+      name: $(default1ESInternalPoolName)
+      image: $(default1ESInternalPoolImage)
+      os: linux
     steps:
     - task: UseDotNet@2
       displayName: Install .NET 8.0 SDK


ellahathaway:
Drafting because I may have to add another variable.

eng/pipelines/secret-management-weekly.yml
```diff
@@ -8,15 +8,18 @@ schedules:
     - main
   always: true
 
+variables:
+- template: templates/variables/common.yml
+
 stages:
```
lbussell, line 14: Should this pipeline just extend the 1es-unofficial template?

ellahathaway: Makes sense, however I just tested with the pipeline with 1es template and I hit https://github.com/microsoft/azure-pipelines-tasks/issues/20396: https://dev.azure.com/dnceng/internal/_build/results?buildId=2537980&view=logs&j=df0a3355-897e-53ae-288d-2f16ccab2822&t=89f293a3-1646-5ba7-40e0-a78a72115a28&l=10.

This doesn't happen with the current implementation without the templates. I'm wary of switching until the linked issue is resolved.

eng/pipelines/secret-management-weekly.yml
```diff
@@ -8,40 +8,42 @@ schedules:
...
+      jobs:
+      - job: Synchronize
+        displayName: Synchronize secrets
+        steps:
+        - task: UseDotNet@2
+          displayName: Install .NET 8.0 SDK
+          inputs:
+            packageType: sdk
+            version: 8.0.x
+            installationPath: '$(Build.Repository.LocalPath)/.dotnet'
```
lbussell, line 28: One last thing, unrelated to the rest of the PR, but does secret manager only use .NET 6? Can we remove this install step for .NET 8? Or file a follow-up issue/PR for that.

lbussell (approved):

eng/pipelines/secret-management-weekly.yml
ellahathaway: It needs .NET 6 runtime and a .NET sdk. I suppose we could use .NET 6 for the sdk, but that'll be EOL soon so I'd rather not introduce a change now that'll have to be reverted soon anyways.

lbussell: The .NET 6 runtime will be EOL at the same time as the SDK. Is there a tracking issue for secret manager for updating to a supported .NET version?

ellahathaway: Just filed https://github.com/dotnet/dnceng/issues/4083

PR #1446, opened 2024-09-23
Enable relative links in Docker image tags tables

lbussell:
Context: https://github.com/dotnet/dotnet-docker/issues/5872#issuecomment-2369097600

This allows specifying `useRelativeLinks` in the manifest per each readme, and if it's set, then the link to the Dockerfiles in the tags table for that readme will be a relative link instead of a GitHub permalink.

This is where you would specify this in the manifest: https://github.com/dotnet/dotnet-docker/blob/08e0e7c3345a75af15ca0c1733d4f07676e0997d/manifest.json#L21-L24

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -171,7 +171,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
             foreach (PlatformData platform in processedPlatforms)
             {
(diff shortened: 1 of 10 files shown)


mthalman (commented):
In what scenario would we not use relative links?

lbussell:
> In what scenario would we not use relative links?

Good point. I'll edit this to make them all use relative links. Should simplify the code some.

lbussell:
I have updated the code to output relative links for readmes, but keep GitHub permalinks for MCR tags metadata, since we do push that metadata to MCR and they could decide to use those links again at any time.

Issue #1417, opened 2024-08-21
EOL annotation race condition


mthalman:
When [EOL annotation support](https://github.com/dotnet/docker-tools/issues/1201) is added to the [dotnet-buildtools-prereqs-docker](https://github.com/dotnet/dotnet-buildtools-prereqs-docker) repo, it can result in a race condition when determining which images should be annotated, potentially resulting in supported images being EOL-annotated.

This repo is unique in this aspect because of its use of multiple pipelines that all write to the same [image info file](https://github.com/dotnet/versions/blob/main/build-info/docker/image-info.dotnet-dotnet-buildtools-prereqs-docker-main.json). When there are multiple pipelines running concurrently, there can be the following race condition scenario:
1. Debian pipeline starts
2. Fedora pipeline starts
3. Debian pipeline pushes new images to ACR (but hasn't yet updated image info file)
4. Fedora pipeline pushes new images to ACR
5. Fedora pipeline updates image info file
6. Fedora pipeline queries to find images that need to be annotated. It finds the Debian images and sees they are not in the image info file.
7. Fedora pipeline sets EOL annotations, which includes newly published Debian images

Possible solutions:
* Have detection logic in the pipeline to see whether this scenario occurred and throw an error. This doesn't prevent the scenario but simply informs that this happened, allowing a team member to rebuild the effected images. In the example above, the Debian pipeline would check whether any of the images it published had been marked as EOL and it would run that check after updating the image info file.
* Prevent multiple pipelines from running concurrently. This may mean going back to a single pipeline.
* Define separate image info files for each pipeline. This would mean having separate container repos for each of them.
* Define some wait behavior in the pipeline so that it won't proceed with querying for images to be annotated if there have been recent pushes (within X minutes) to the registry outside the context of the images currently being published by that pipeline.
* Use pipeline tags as semaphores to indicate the pipeline status so that multiple concurrent runs can "communicate" with each other. Using the example above, Debian pipeline would add tags when it has started to push images and after it had updated the image info. For the Fedora pipeline, it would wait to proceed with querying for images to be annotated until after any other concurrently running pipeline had those two tags set.

mthalman:
> Another way to go back to a single build pipeline could be to compute which images will be cached/built ahead of build matrix generation, so that we avoid spinning up too much compute for each individual build leg.

I'm going with this approach.

PR #1449, opened 2024-09-26
Trim matrix based on image cache state

mthalman:
Contributes to #1417

This adds trimming logic to Image Builder's matrix generation. This trimming will optimize the number of build jobs that get created. It prevents scenarios of spinning up jobs that are essentially no-ops because the images they process may all be cached.

It trims the set of Dockerfiles based on whether cached images already exist for those Dockerfiles. Once that set of Dockerfiles is determined, then it proceeds with its usual logic, potentially expanding the set of Dockerfiles that get included. So the final matrix result may include images which are cached, but those are necessary to be included based on the build requirements.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Linq;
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
```diff
@@ -396,30 +405,85 @@ private static string FormatMatrixName(IEnumerable<string> parts)
             return allParts.First() + string.Join(string.Empty, allParts.Skip(1).Select(part => part.FirstCharToUpper()));
         }
 
-        private IEnumerable<PlatformInfo> GetPlatforms()
+        private async Task<IEnumerable<PlatformInfo>> GetPlatformsAsync()
         {
             if (_imageArtifactDetails.Value is null)
             {
                 return Manifest.GetFilteredPlatforms();
             }
```
lbussell, line 413: I debugged this with an "internal build" scenario to see how it works, and didn't notice a difference in behavior. The reason is because when we run an internal build, we're not going to have any `ImageArtifactDetails` to work with yet since that comes from the build output. So this condition here will trigger and return early from this method without trimming any platforms. It seems to me that the trimming needs to happen on the platforms directly from the manifest. And if we have `ImageArtifactDetails` to work with, then we shouldn't need to trim anything from that (i.e. we don't want to accidentally trim images off of our test matrices, right?)

lbussell (changes requested):

lbussell (approved):

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs
lbussell: I tested it again with the additional options. Works great.

PR #1450, opened 2024-09-26
Remove unnecessary Docker client version check

mthalman:
The `DockerHelper.Login` method had special logic to handle a legacy Docker client version (less than 17.7). That version is very old and has long been unsupported. We do not need to handle that version any longer. This removes all associated code.


lbussell (approved):
I double checked the Docker versions we run in all of our VM images. LGTM 👍

PR #1457, opened 2024-10-02
Enable tsa and policheck in pipelines

MichaelSimons:
These are required compliance steps.


lbussell:
@MichaelSimons have you tested these internally anywhere to make sure that they function correctly? We don't want to suddenly have failing pipeline runs if we have any violations.

MichaelSimons:
> @MichaelSimons have you tested these internally anywhere to make sure that they function correctly? We don't want to suddenly have failing pipeline runs if we have any violations.

I have been enabling these across our repos.  TSA and Policheck doesn't fail the build - they upload results to analyze separately.

lbussell (approved):

PR #1461, opened 2024-10-06
Remove InnerManifestService by using interface default implementations

### LINKED ISSUE #1296: Remove the need for ManifestClient wrapper for mocking extension methods
https://github.com/dotnet/docker-tools/pull/1293 introduced the `InnerManifestClient`/`ManifestClient` pairing of classes. `ManifestClient` wraps the `InnerManifestClient` (which was formerly just `ManifestClient`) so that we can mock its extension methods for testing.

The core issue is that GetImageDigestAsync makes a call to the Docker CLI, so mocking the `InnerManifestClient` by itself is not sufficient.

https://github.com/dotnet/docker-tools/blob/354896e65e7893aa79cfef70751e66db666077a6/src/Microsoft.DotNet.ImageBuilder/src/ManifestServiceExtensions.cs#L38-L65

We should re-architect this code so that we can mock only the `InnerManifestClient`'s methods directly, and then remove the wrapper and rename.

I was thinking that we could separate out `GetImageDigestAsync` into `ManifestServiceExtensions.GetImageDigestAsync` and `DockerService.GetLocalImageDigest`. Then, callers (like the build command) could define their own method for checking local images first or reaching out to the ACR when a local image isn't found (or define a static method for that, taking both services as arguments, to provide a shared implementation for all of the callers).

lbussell:
Fixes #1296 by using default method implementation on the interface rather than using extension methods.


src/Microsoft.DotNet.ImageBuilder/src/IManifestService.cs
```diff
@@ -11,7 +14,59 @@ namespace Microsoft.DotNet.ImageBuilder;
 public interface IManifestService
 {
     Task<ManifestQueryResult> GetManifestAsync(string image, bool isDryRun);
-    Task<IEnumerable<string>> GetImageLayersAsync(string tag, bool isDryRun);
-    Task<string?> GetLocalImageDigestAsync(string image, bool isDryRun);
-    Task<string> GetManifestDigestShaAsync(string tag, bool isDryRun);
+
+    public async Task<IEnumerable<string>> GetImageLayersAsync(string tag, bool isDryRun)
```
mthalman, line 18: How does this work for mocks? I still see these methods having mock implementations in the tests.

lbussell: By setting the `CallBase` parameter, Moq will actually call the default method implementation on the interface unless it has been overridden. And by putting these methods as default implementations on the interface, we still kind of get the same isolation from the `ManifestService` class, ensuring that they can't depend on anything that isn't defined in the interface. I suppose we could also seal those methods.

In order to mock a specific output from `GetImageLayersAsync`, then we can set up the mock for that method explicitly. Since the method is on the interface then everything works as expected.

Alternatively, if we set `CallBase = true`, we can set up the mock output for `GetManifest`. Then calling `GetImageLayersAsync` will actually run the code in `GetImageLayersAsync`, and it'll get the output from the mocked method `GetManifest`.

I found this to be the most flexible solution that removed the InnerManifest class and at the same time didn't require re-writing tons of test code.

lbussell: Well, you can't mock `sealed` methods. But I tried!

mthalman: Ok, seems like we have some flexibility here then. Sounds good.

mthalman (approved):

Issue #1470, opened 2024-10-16
EOL annotations not created on publish retry


mthalman:
In the scenario where the publish stage fails and it is retried, it is possible for no EOL annotations to be created on that retry (assuming it never got to the point of publishing annotations on the first attempt).

Example where this occurs: [build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2561042&view=logs&j=046815c1-7653-5dfd-0c09-c18aa9ed7f03&t=3f9f21c1-05cf-5f86-ae06-51569585cd33) (internal link). See `No digests to annotate because no image info files were provided.` in the output of the `Generate EOL Annotation Data (Authenticated)` step.

This occurs because the `Publish Image Info` step doesn't end up making any changes to the image info file of the versions repo. This is because it had already updated it in the first build attempt. Because it doesn't push any changes, it never [writes to the image info output files](https://github.com/dotnet/docker-tools/blob/10acdce4fd71b62bc9d4dce8f4e6f7075579b6d3/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs#L98-L108) because it [stops execution](https://github.com/dotnet/docker-tools/blob/10acdce4fd71b62bc9d4dce8f4e6f7075579b6d3/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs#L64-L68) before reaching that point. And since the image info output files are not written to, the `GenerateEolAnnotationCommand` [executes a no-op](https://github.com/dotnet/docker-tools/blob/10acdce4fd71b62bc9d4dce8f4e6f7075579b6d3/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs#L79-L83) because of https://github.com/dotnet/docker-tools/pull/1442. The result is that no attempt is made to find digests that need to be annotated and so nothing gets annotated.

A subsequent build (new, not retry) will correctly annotate all the images that need to be annotated.

mthalman:
The correct thing to do here is to change the workflow in the publish stage to defer merging the image info file into the versions repo until after annotations have been pushed. It will still require that the EOL computation have access to the fully resolved image info file; it's just that the image info file shouldn't be merged until later. This will ensure that the input to the `generateEolAnnotationData` command has access to the original "before" state of the image info which it relies upon.

[...]

PR #1488, opened 2024-11-05
Split merging of image info from publishing

mthalman:
Contributes to #1470 

As described in https://github.com/dotnet/docker-tools/issues/1470#issuecomment-2436157156, this updates Image Builder so that it splits apart the merging of image info files from the act of publishing the image info file to the versions repo. That logic has been removed from the `publishImageInfo` command. Instead, the already existing `mergeImageInfo` command has been updated to handle the publish scenario. So now `mergeImageInfo` command can be used for merging image info fragments from multiple build jobs (Post-Build stage) as well as merging the final image info with the image info from the versions repo (Publish stage).

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/MergeImageInfoCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/MergeImageInfoCommand.cs
@@ -9,7 +9,9 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.DotNet.ImageBuilder.Models.Image;
+using Microsoft.DotNet.ImageBuilder.ViewModel;
 
+#nullable enable
 namespace Microsoft.DotNet.ImageBuilder.Commands
 {
     [Export(typeof(ICommand))]
@@ -24,9 +26,9 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 "*.json",
                 SearchOption.AllDirectories);
 
-            List<ImageArtifactDetails> srcImageArtifactDetailsList = imageInfoFiles
+            List<(string Path, ImageArtifactDetails ImageArtifactDetails)> srcImageArtifactDetailsList = imageInfoFiles
                 .OrderBy(file => file) // Ensure the files are ordered for testing consistency between OS's.
-                .Select(imageDataPath => ImageInfoHelper.LoadFromFile(imageDataPath, Manifest))
+                .Select(imageDataPath => (imageDataPath, ImageInfoHelper.LoadFromFile(imageDataPath, Manifest, skipManifestValidation: Options.IsPublishScenario)))
                 .ToList();
 
             if (!srcImageArtifactDetailsList.Any())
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/MergeImageInfoCommand.cs
```diff
@@ -24,9 +26,9 @@ public override Task ExecuteAsync()
                 "*.json",
                 SearchOption.AllDirectories);
 
-            List<ImageArtifactDetails> srcImageArtifactDetailsList = imageInfoFiles
+            List<(string Path, ImageArtifactDetails ImageArtifactDetails)> srcImageArtifactDetailsList = imageInfoFiles
                 .OrderBy(file => file) // Ensure the files are ordered for testing consistency between OS's.
-                .Select(imageDataPath => ImageInfoHelper.LoadFromFile(imageDataPath, Manifest))
+                .Select(imageDataPath => (imageDataPath, ImageInfoHelper.LoadFromFile(imageDataPath, Manifest, skipManifestValidation: Options.IsPublishScenario)))
```
lbussell, line 31: I have an allergic reaction to lines over 120 chars long

src/Microsoft.DotNet.ImageBuilder/src/Commands/MergeImageInfoCommand.cs
```diff
@@ -35,16 +37,90 @@ public override Task ExecuteAsync()
...
+            ImageArtifactDetails targetImageArtifactDetails;
+            if (Options.InitialImageInfoPath != null)
+            {
+                targetImageArtifactDetails = srcImageArtifactDetailsList.First(item => item.Path == Options.InitialImageInfoPath).ImageArtifactDetails;
+
+                if (Options.IsPublishScenario)
+                {
+                    RemoveOutOfDateContent(targetImageArtifactDetails);
+                }
+            }
```
lbussell, line 54: Two comments here:
1. Are `IsPublishScenario` and `InitialImageInfoPath` always passed together? It seems like this could _almost_ be a separate command.
2. If `InitialImageInfoPath` is not in the `SourceImageInfoFolderPath`, then we'll fail here. Would it be better to just read it in explicitly here if it's not in the folder? Or will our pipelines never encounter that scenario? Seems weird to just have it technically passed in twice (in the folder, then by path).

mthalman: > Are IsPublishScenario and InitialImageInfoPath always passed together? It seems like this could almost be a separate command.

Yes, they will be. But passing `InitialImageInfoPath` doesn't _have_ to mean that it's a publish scenario.

> If InitialImageInfoPath is not in the SourceImageInfoFolderPath, then we'll fail here. Would it be better to just read it in explicitly here if it's not in the folder? Or will our pipelines never encounter that scenario? Seems weird to just have it technically passed in twice (in the folder, then by path).

To me, it would be weird to process a file that wasn't contained in the folder. The folder is the true input of what to process. The `InitialImageInfoPath` is just an optional reference to one of those files.

lbussell (approved):

PR #1494, opened 2024-11-11
Publish EOL annotations on delete-only scenario

### LINKED ISSUE #1484: Dockerfile deletion scenario won't trigger immediate EOL annotation publishing
In the scenario where an internal CI build gets triggered where the only update is a Dockerfile deletion, the expectation is that the build will publish an EOL annotation for the associated image. At least for pipelines that have matrix trimming enabled, that won't happen today. The pipeline will essentially no-op; no build jobs will be run due to the matrix trimming and the publish stage will be skipped. Further testing is needed to determine what would happen in a pipeline that isn't using matrix trimming.

But this is a scenario that we should probably support as it will commonly occur when deleting Dockerfiles for EOL distros/.NET versions.

mthalman:
Fixes #1484 

In a build where matrix trimming is enabled and the only change is to delete a Dockerfile, we need to ensure that an EOL annotation gets published for the associated image. Currently, the Publish stage gets skipped in this scenario because there is no image info produced by any build jobs.

This is fixed by allowing the Publish stage to run but have it skip certain steps when no image info file exists. This is indicated by having the Post-Build stage produce a special flag file as a pipeline artifact that is detected by the Publish stage.

### DIFF
--- a/eng/common/templates/jobs/post-build.yml
+++ b/eng/common/templates/jobs/post-build.yml
@@ -55,13 +55,17 @@ jobs:
     displayName: Prune Publish Artifacts
   - powershell: |
       $imageInfoFiles = Get-ChildItem "$(imageInfosHostDir)"
+      New-Item -ItemType Directory -Path $(imageInfosHostDir)$(imageInfosOutputSubDir) -Force
       if ($imageInfoFiles.Count -eq 0) {
         echo "No image info files found."
         echo "##vso[task.setvariable variable=noImageInfos;isOutput=true]true"
+
+        # Create a flag file to indicate that no image info files were found
+        Set-Content $(imageInfosHostDir)$(imageInfosOutputSubDir)/$(noImageInfoFlagFile) 1
+
         exit 0
       }
 
-      New-Item -ItemType Directory -Path $(imageInfosHostDir)$(imageInfosOutputSubDir) -Force
       $(runImageBuilderCmd) mergeImageInfo `
         --manifest $(manifest) `
         $(imageInfosContainerDir) `
@@ -79,7 +83,6 @@ jobs:
       publicProjectName: ${{ parameters.publicProjectName }}
   - template: /eng/common/templates/steps/publish-artifact.yml@self
     parameters:
-      condition: and(succeeded(), ne(variables['MergeImageInfoFiles.noImageInfos'], 'true'))
       path: $(imageInfosHostDir)$(imageInfosOutputSubDir)
       artifactName: image-info
       displayName: Publish Image Info File Artifact
--- a/eng/common/templates/jobs/post-build.yml
+++ b/eng/common/templates/jobs/post-build.yml
@@ -55,17 +55,13 @@ jobs:
     displayName: Prune Publish Artifacts
   - powershell: |
       $imageInfoFiles = Get-ChildItem "$(imageInfosHostDir)"
-      New-Item -ItemType Directory -Path $(imageInfosHostDir)$(imageInfosOutputSubDir) -Force
       if ($imageInfoFiles.Count -eq 0) {
         echo "No image info files found."
         echo "##vso[task.setvariable variable=noImageInfos;isOutput=true]true"
-
-        # Create a flag file to indicate that no image info files were found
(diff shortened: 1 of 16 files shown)


eng/common/templates/jobs/post-build.yml
```diff
@@ -55,13 +55,17 @@ jobs:
...
       $imageInfoFiles = Get-ChildItem "$(imageInfosHostDir)"
+      New-Item -ItemType Directory -Path $(imageInfosHostDir)$(imageInfosOutputSubDir) -Force
       if ($imageInfoFiles.Count -eq 0) {
         echo "No image info files found."
         echo "##vso[task.setvariable variable=noImageInfos;isOutput=true]true"
+
+        # Create a flag file to indicate that no image info files were found
+        Set-Content $(imageInfosHostDir)$(imageInfosOutputSubDir)/$(noImageInfoFlagFile) 1
+
         exit 0
```
lbussell, line 66: Does using a file as a flag here accomplish anything that an azure pipelines variable can't? [You can use variables across stages](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#use-outputs-in-a-different-stage).

eng/common/templates/jobs/publish.yml
```diff
@@ -94,12 +106,13 @@ jobs:
...
       internalProjectName: ${{ parameters.internalProjectName }}
       publicProjectName: ${{ parameters.publicProjectName }}
+      condition: ne(variables.noImagesToPublish, 'true')
   - template: /eng/common/templates/steps/wait-for-mcr-image-ingestion.yml@self
     parameters:
       imageInfoPath: '$(imageinfoContainerDir)/image-info.json'
       minQueueTime: $(imageQueueTime)
       dryRunArg: $(dryRunArg)
-      condition: succeeded()
+      condition: ne(variables.noImagesToPublish, 'true')
```
lbussell, line 115: Should this be 
```suggestion
      condition: and(succeeded(), ne(variables.noImagesToPublish, 'true'))
```

eng/common/templates/jobs/publish.yml
```diff
@@ -54,11 +63,13 @@ jobs:
...
       '$(imageInfoContainerDir)/image-info.json'
     displayName: Trim Unchanged Images
+    condition: ne(variables.noImagesToPublish, 'true')
   - template: /eng/common/templates/steps/run-imagebuilder.yml@self
     parameters:
       displayName: Copy Images
       serviceConnection: $(publish.serviceConnectionName)
       internalProjectName: ${{ parameters.internalProjectName }}
+      condition: ne(variables.noImagesToPublish, 'true')
       args: >
```
lbussell, line 73: My opinion is that instead of skipping specific stages based on pipeline logic, that the ImageBuilder commands themselves should ideally be able to recognize when they don't need to run and just no-op instead. The more special logic that we keep out of pipelines the better. Are the conditions an attempt at making the pipeline faster?

eng/common/templates/jobs/post-build.yml
mthalman: Yes. Cross-stage variables are of no use here in scenarios where you execute the publish stage by itself.

eng/common/templates/jobs/publish.yml
mthalman: That's fair. I'll experiment with that to see what changes might be necessary in Image Builder.

eng/common/templates/jobs/post-build.yml
lbussell: I see. Because we output it as an artifact, then it'll be downloaded by any pipelines that reference that build ID. Makes sense now.

eng/common/templates/jobs/publish.yml
lbussell: [Gracefully handle missing image info in publish commands (#1501)](https://github.com/dotnet/docker-tools/pull/1501) was merged. Do you plan to re-visit this, or is this PR unnecessary now?

mthalman: Yep, working on that this morning.

lbussell (approved):

PR #1503, opened 2024-11-19
Reduce concurrent delete requests

mthalman:
Ran into another issue with the ACR getting in a bad state from image deletion. Related to https://github.com/dotnet/docker-tools/pull/1401. Reducing the concurrent request count down some more to try to prevent this from happening.


lbussell:
Will this fix the recent failure in the cleanup pipeline?

https://dev.azure.com/dnceng/internal/_build/results?buildId=2584658&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=550011cb-b399-5286-851b-873bc5e1386f&l=194642

```
Unhandled exception: Azure.RequestFailedException: manifest sha256:ab799474005e3b968ee6303109e3d17cc75be81258308039b2f164d79a4ad32b is not found
```

It has happened on the same digest multiple times. Are manual steps required to get it back into a good state?

lbussell (approved):

Issue #1428, opened 2024-09-09
ImageBuilder requires Docker to run all commands, even if they don't use it


lbussell:
This is needed for https://github.com/dotnet/dotnet-docker/issues/5654.

## Steps to reproduce the issue
1. Run ImageBuilder in a scenario where Docker is not installed

## Additional information (e.g. issue happens only occasionally)
There are two issues here.

One is that this CLI middleware runs for every command, outputting `docker version` and `docker info` before the command starts:

https://github.com/dotnet/docker-tools/blob/eb53e75aa4157d922b8c30d9132f8d11b79e8183/src/Microsoft.DotNet.ImageBuilder/src/ImageBuilder.cs#L55-L63

This is only useful for CI scenarios, not local dev scenarios, so it could be hidden behind a common CLI argument (like `--ci` or `--no-ci`) that we only use in pipelines.

The second is that, for commands with a `ManifestFilterOptions`, the default values for those options rely on output from the Docker CLI:

https://github.com/dotnet/docker-tools/blob/eb53e75aa4157d922b8c30d9132f8d11b79e8183/src/Microsoft.DotNet.ImageBuilder/src/Commands/ManifestFilterOptions.cs#L29-L37

For cases like generating readmes and Dockerfiles, it doesn't make sense to filter by arch or OS type. We could split `ManifestFilterOptions` into `DockerfileFilterOptions` and `PlatformFilterOptions`. `DockerfileFilterOptions` would let you filter by version, OS, or path. The other would let you filter by architecture and OS type (Linux/Windows). We could combine both of those for some commands as needed.

I would probably argue that it doesn't make sense for readme and Dockerfile generation to take path filters at all - our goal should be for generating those to be as quick and frictionless as possible.

PR #1508, opened 2024-11-27
Remove default dependency on Docker CLI for generating Dockerfiles

lbussell:
This PR contains some changes needed to accomplish [Consider using Docker custom build outputs (`--output`) for generated Dockerfiles and Readmes (dotnet/dotnet-docker#5654)](https://github.com/dotnet/dotnet-docker/issues/5654), by fixing [ImageBuilder requires Docker to run all commands, even if they don't use it (#1428)](https://github.com/dotnet/docker-tools/issues/1428).

These changes shouldn't cause breaks for any features used by any commands. It does remove some filtering options from `GenerateDockerfiles`, but those options were unused.

Commits are separated into buildable chunks that each pass all tests, if that makes reviewing easier.

## The problem

- https://github.com/dotnet/dotnet-docker/issues/5654 requires commands be used in a Dockerfile build and not a running container.
- In order to use ImageBuilder inside a Dockerfile build, we can't use any `docker` CLI commands because we aren't going to bind-mount the Docker socket.
- `docker version` and `docker info` are run for every command.
- Some commands call the `docker` CLI to figure out the default value for some options (arch and OS version).

## Solution

- `docker version` and `docker info` are run for every command.
  - This is turned into a new option, `--ci` that is allowed for every command, that we can use in shared pipeline templates
- Some commands call the `docker` CLI to figure out the default value for some options (arch and OS version).
  - The arch and OS version filters are not useful for all commands. So, we can split `ManifestFilterOptions` into two separate sets of options: `PlatformFilterOptions` and `DockerfileFilterOptions`. Then, for `GenerateDockerfiles`, we use only `DockerfileFilterOptions` so that it won't call the `docker` CLI anymore.

## Using the new ImageBuilder

Here's an example Dockerfile that we would use to generate Readmes and Dockerfiles. (additional work is needed to pass in `branch` via an `ARG`, also should really use `COPY --link`)

```Dockerfile
FROM mcr.microsoft.com/dotnet-buildtools/image-builder:linux-amd64 AS imagebuilder
WORKDIR /repo
COPY . .

# generate-dockerfiles generates Dockerfiles from templates
FROM imagebuilder AS generate-dockerfiles
RUN [ \
    "/image-builder/Microsoft.DotNet.ImageBuilder", \
    "generateDockerfiles", \
    "--var", "branch=nightly" ]

# generate-readmes generates Readmes from templates
FROM imagebuilder AS generate-readmes
RUN [ \
    "/image-builder/Microsoft.DotNet.ImageBuilder", \
    "generateReadmes", \
    "--manifest", "manifest.json", \
    "--var", "branch=nightly", \
    "'https://github.com/dotnet/dotnet-docker'" ]
RUN [ \
    "/image-builder/Microsoft.DotNet.ImageBuilder", \
    "generateReadmes", \
    "--manifest", "manifest.samples.json", \
    "--var", "branch=main", \
    "'https://github.com/dotnet/dotnet-docker'" ]

# generate-dockerfiles-output contains generated Dockerfiles
FROM scratch AS generate-dockerfiles-output
COPY --from=generate-dockerfiles /repo/src src

# generate-readmes-output contains generated Readmes
FROM scratch AS generate-readmes-output
COPY --from=generate-readmes /repo/*.md ./
COPY --from=generate-readmes /repo/.portal-docs ./.portal-docs
```

## Results

`--no-cache` used for all Docker build commands, including those in PowerShell scripts. Measured with `Measure-Command` in PowerShell.

| Command | Execution time |
| --- | --- |
| `.\eng\dockerfile-templates\get-generateddockerfiles.ps1` | 7.78 seconds |
| `.\eng\readme-templates\get-generatedreadmes.ps1` | 16.28 seconds |
| `docker build --no-cache --target=generate-dockerfiles-output -f .\Dockerfile -o . .` | **5.23** seconds |
| `docker build --no-cache --target=generate-readmes-output -f .\Dockerfile -o . .` | **5.13** seconds |

### Caveats

Doesn't work on Windows since BuildKit doesn't support WCOW yet.

## Try it out

(cloning not required)

```pwsh
# Build ImageBuilder
docker build -t imagebuilder -f ./Dockerfile.linux https://github.com/lbussell/docker-tools.git#ci-mode-2:src/Microsoft.DotNet.ImageBuilder/

$dockerfile="https://raw.githubusercontent.com/lbussell/docker-tools/refs/heads/ci-demo/src/Microsoft.DotNet.ImageBuilder/Dockerfile.demo"

# See available options...
docker build -q --call=targets -f $dockerfile .

# Use the new Dockerfile to generate contents
docker build --no-cache --target=generate-dockerfiles-output -f $dockerfile -o . .
docker build --no-cache --target=generate-readmes-output -f $dockerfile -o . .
```

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/Options.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/Options.cs
@@ -16,6 +16,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     {
         public bool IsDryRun { get; set; }
         public bool IsVerbose { get; set; }
+        public bool CIMode { get; set; }
     }
 
     public class CliOptionsBuilder
@@ -39,7 +40,11 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 CreateOption<bool>(
                     alias: "verbose",
                     propertyName: nameof(Options.IsVerbose),
-                    description: "Show details about the tasks run")
+                    description: "Show details about the tasks run"),
+                CreateOption<bool>(
+                    alias: "ci",
+                    propertyName: nameof(Options.CIMode),
+                    description: "Enable additional logging useful for CI scenarios")
             ];
     }
 }
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/Options.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/Options.cs
@@ -16,7 +16,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     {
         public bool IsDryRun { get; set; }
         public bool IsVerbose { get; set; }
-        public bool CIMode { get; set; }
+        public bool NoVersionLogging { get; set; }
     }
 
     public class CliOptionsBuilder
@@ -42,9 +42,9 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                     propertyName: nameof(Options.IsVerbose),
                     description: "Show details about the tasks run"),
                 CreateOption<bool>(
-                    alias: "ci",
-                    propertyName: nameof(Options.CIMode),
-                    description: "Enable additional logging useful for CI scenarios")
+                    alias: "no-version-logging",
+                    propertyName: nameof(Options.NoVersionLogging),
+                    description: "Disable automatic logging of Docker version information")
             ];
     }
 }
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/Command.TOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/Command.TOptions.cs
@@ -42,6 +42,11 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
             cmd.Handler = CommandHandler.Create<TOptions>(options =>
             {
+                if (Options.CIMode)
+                {
+                    LogDockerVersions();
+                }
+
                 Initialize(options);
                 return ExecuteAsync();
             });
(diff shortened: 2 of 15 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/Options.cs
```diff
@@ -39,7 +40,11 @@ public virtual IEnumerable<Option> GetCliOptions() =>
                 CreateOption<bool>(
                     alias: "verbose",
                     propertyName: nameof(Options.IsVerbose),
-                    description: "Show details about the tasks run")
+                    description: "Show details about the tasks run"),
+                CreateOption<bool>(
+                    alias: "ci",
+                    propertyName: nameof(Options.CIMode),
+                    description: "Enable additional logging useful for CI scenarios")
```
mthalman, line 47: I'm reluctant to have such an option. Having this option will require consistent configuration across all pipelines that use Image Builder to set this option. In my view, Image Builder is primarily a CI tool so it would seem better for the CI behavior to be the default. For the few dev scenarios where we don't want that, there could be an option to opt out.

lbussell: Makes complete sense. I did choose this way because we like to have `false` by default for options, and we can set this flag directly in our shared [`run-imagebuilder` template](https://github.com/dotnet/docker-tools/blob/main/eng/common/templates/steps/run-imagebuilder.yml).

However, we can just change the option to `--no-ci` or even something like `--no-version-logging` to be more specific.

lbussell: I went with the latter approach.

mthalman (approved):

PR #1512, opened 2024-12-06
Exclude missing EOL digests from the EOL data output

### LINKED ISSUE #1507: EOL annotation fails when EOL Dockerfiles exist but images do not exist in ACR
There can be circumstances where the EOL annotation logic will fail when attempting to query the metadata about an image in the ACR, where the image does not exist.

This occurs when the following conditions occur:
* A .NET version has become EOL (such as 6.0) and the build has marked all of those images as EOL.
* Those Dockerfiles are not yet removed from the repo.
* ACR cleanup has run which deletes those EOL images from the ACR.
* A subsequent build runs which builds Dockerfiles other than the EOL ones.

The issue here is in the `generateEolAnnotationData` command. It will see that the EOL Dockerfiles exist and that they should be added to the list of digests to be annotated here: https://github.com/dotnet/docker-tools/blob/6c08c3f30f8a88e21e5539d37ce5133a61aaa049/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs#L204-L229. The problem is that the images shouldn't have been added in the first place, or at least checked to see whether they exist before attempting to get their metadata.

[Example failing build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2589683&view=results) (internal link)

mthalman:
Fixes #1507 by updating the `generateEolAnnotationData` command to not output digests which do not already exist in the registry.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
@@ -130,15 +130,13 @@ public class GenerateEolAnnotationDataCommand : Command<GenerateEolAnnotationDat
                 // Annotate images for EOL products in new image info
                 // Only do so for those digests that actually exist in the registry (they may have been cleaned up
                 // because they are EOL).
-                foreach (EolDigestData eolDigest in
+                IEnumerable<EolDigestData> eolDigests =
                     newImageArtifactDetails.Repos
                         .SelectMany(repo =>
                             repo.Images
                                 .SelectMany(image => GetProductEolDigests(image, productEolDates)))
-                        .Where(digestData => registryTagsByDigest.ContainsKey(digestData.Digest)))
-                {
-                    digestDataList.Add(eolDigest);
-                }
+                        .Where(digestData => registryTagsByDigest.ContainsKey(digestData.Digest));
+                digestDataList.AddRange(eolDigests);
             }
         }
         catch (Exception e)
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
```diff
@@ -127,10 +127,17 @@ private async Task<List<EolDigestData>> GetDigestsToAnnotate()
...
+                foreach (EolDigestData eolDigest in
+                    newImageArtifactDetails.Repos
+                        .SelectMany(repo =>
+                            repo.Images
+                                .SelectMany(image => GetProductEolDigests(image, productEolDates)))
+                        .Where(digestData => registryTagsByDigest.ContainsKey(digestData.Digest)))
                 {
-                    digestDataList.AddRange(GetProductEolDigests(image, productEolDates));
+                    digestDataList.Add(eolDigest);
                 }
```
lbussell, line 141: Is the for loop necessary here? Nothing is async here so this would be easier to read and debug I think:

```cs
IEnumerable<EolDigestData> eolDigests =
    newImageArtifactDetails.Repos
        .SelectMany(repo =>
            repo.Images
                .SelectMany(image => GetProductEolDigests(image, productEolDates)))
        .Where(digestData => registryTagsByDigest.ContainsKey(digestData.Digest)))

digestDataList.AddRange(eolDigests);
```

lbussell (approved):
Approved with one small suggestion.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommand.cs
mthalman: fixed

lbussell (approved):
:shipit:

Issue #1516, opened 2024-12-11
Dependent images are cached and not re-built when their base image's Dockerfile is changed


lbussell:
See https://github.com/dotnet/dotnet-docker/pull/6110#issuecomment-2536965183

## Steps to reproduce the issue

Queue a PR validation build of dotnet-docker repo where some runtime deps Dockerfiles are changed but not any dependent runtime or aspnet Dockerfiles.

## Expected behavior

All images dependent on the changed runtime-deps images will be re-built.

## Actual behavior

No images dependent on the changed runtime-deps images are re-built. This results in test failures when there is an expected behavior change like in https://github.com/dotnet/dotnet-docker/pull/6110.

mthalman:
This is a regression that was caused by https://github.com/dotnet/docker-tools/pull/1427. That change explicitly prevented it from accounting for locally built base images which is problematic for this scenario.

Issue #1517, opened 2024-12-12
Build command doesn't work in some scenarios with new Docker version


mthalman:
While attempting to fix https://github.com/dotnet/docker-tools/issues/1516 I was trying to run my fix locally with the configuration that would verify it was working (i.e. image caching was enabled). It ended up not working correctly because some of the logic in Image Builder was not getting the expected results from the locally built image.

This logic expects that a locally built image will not have a digest associated with it from the `docker inspect` command: https://github.com/dotnet/docker-tools/blob/008a7cf2b73cf7af715f9e692cfd84911da693b0/src/Microsoft.DotNet.ImageBuilder/src/IManifestService.cs#L41-L47

This turns out not to be the case for my version of Docker Desktop. In other words, the image _does_ have a digest for the locally built image. This causes issues later on in that method because it will attempt to verify the digest reflects what is the registry (because it assumes it is not a locally built image at this point):

https://github.com/dotnet/docker-tools/blob/008a7cf2b73cf7af715f9e692cfd84911da693b0/src/Microsoft.DotNet.ImageBuilder/src/IManifestService.cs#L49-L62

This then throws the exception because the digests don't match which makes sense.

So at some point, a new version of Docker (not sure if this is in the client or server) caused the digest to be created. You can see this in the build output: 

![Image](https://github.com/user-attachments/assets/ced8d872-ee6e-4b17-a382-448da60556dd)

This differs from the behavior of the version of Docker installed on the build agents. But this will become an issue whenever they get updated to whichever version this behavior was introduced.

For reference, here's my `docker info` output for my Docker Desktop installation:

```
Client:
 Version:    27.3.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Ask Gordon - Docker Agent (Docker Inc.)
    Version:  v0.1.0
    Path:     C:\Users\mthalman\.docker\cli-plugins\docker-ai.exe
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.18.0-desktop.2
    Path:     C:\Users\mthalman\.docker\cli-plugins\docker-buildx.exe
  compose: Docker Compose (Docker Inc.)
    Version:  v2.30.3-desktop.1
    Path:     C:\Users\mthalman\.docker\cli-plugins\docker-compose.exe
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.37
    Path:     C:\Users\mthalman\.docker\cli-plugins\docker-debug.exe
[...]

PR #1518, opened 2024-12-12
Image caching fix for locally built images

mthalman:
Fixes https://github.com/dotnet/docker-tools/issues/1516

In the scenario where this issue was discovered, it was the runtime image whose base runtime-deps image was being inspected to determine whether it had changed. So it was querying the registry the runtime-deps digest instead of accounting for the new locally built version of runtime-deps. Since runtime-deps was rebuilt then runtime should also be rebuilt. But since it had the digest from the registry and that matched what was in the image info, then it just used the cached image for runtime.

This fixes the logic by accounting for two different scenarios. In the build scenario, all base images will have either been built locally (if they are .NET base images) or pre-emptively pulled. Prior to this fix, the logic was not accounting for a locally built image at all and would always query the registry. So for build, it should check for the local image. But for the matrix generation command, there are no builds happening so it must query the registry all the time to get the digest.

Related: https://github.com/dotnet/docker-tools/issues/1517

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/tests/BuildCommandTests.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/tests/BuildCommandTests.cs
@@ -1085,6 +1085,13 @@ namespace Microsoft.DotNet.ImageBuilder.Tests
             null, "runtimeDepsCommitSha-1",
             null, "runtimeCommitSha-1",
             false, false)]
+        [InlineData(
+            "All previously published, commit diff for runtime-deps",
+            "sha256:baseImageSha", "sha256:baseImageSha",
+            "sha256:runtimeDepsImageSha-1", "sha256:runtimeDepsImageSha-1",
+            "runtimeDepsCommitSha-1", "runtimeDepsCommitSha-2",
+            "runtimeCommitSha", "runtimeCommitSha",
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.ImageBuilder/tests/BuildCommandTests.cs
```diff
@@ -1125,7 +1132,7 @@ public async Task BuildCommand_Caching(
                 localImageDigestResults:
                 [
                     new($"{runtimeDepsRepoQualified}:{tag}", runtimeDepsDigest),
-                    new($"{overridePrefix}{runtimeDepsRepo}:{tag}", runtimeDepsDigest),
+                    new($"{overridePrefix}{runtimeDepsRepo}:{tag}", runtimeDepsDigest, OnCallCount: 2),
```
lbussell, line 1135: What's the reason for only returning this on the second call now?

mthalman: The first call should return null to simulate it's locally built and doesn't have a digest yet. Later during the build, after a push, it calls it again and now it should have the digest.

lbussell (approved):

PR #1522, opened 2024-12-19
Make PublishManifest command skip unchanged images

### LINKED ISSUE #1510: publishManifest command should skip unchanged tags
The [`publishManifest` command](https://github.com/dotnet/docker-tools/blob/3a078d892227e6446a5bcaf13536b8e37934685d/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs#L19) is processing shared tags that don't necessarily need to be updated.

For example, the [build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2594275&view=results) (internal link) that ran for https://github.com/dotnet/dotnet-docker/pull/6086 had no path filter set. The PR only updated 9.0 Dockerfiles. So all the .NET versions were cached, resulting in unchanged images for those versions. That was correctly indicated, via the `isUnchanged` field, in the image info file that was produced. But when `publishManifest` ran, it ignored that data and processed all shared tags, such as `6.0`. This was unnecessary processing since the `6.0` manifest list remained identical to the current version. It also was problematic because it relied on those associated images existing within the ACR in order to produce the manifest list. In this scenario, those images didn't actually exist anymore because they were EOL and had been removed from the ACR but the Dockerfiles hadn't yet been removed. So that actually broke the build because of this behavior, which was worked around.

The solution to this is for `publishManifest` to look at the `isUnchanged` field of the images associated with the shared tag. If all those images have `isUnchanged` set to `true`, then the shared tag should be skipped.

lbussell:
Fixes #1510

I don't love the null-forgiving operator I had to use in PublishManifestCommand, so I'm trying to think of a different solution for that. But it isn't the worst.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
@@ -80,8 +80,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                                 .Where(image => image.AllPlatforms
                                     .Select(platform =>
                                         ImageInfoHelper.GetMatchingPlatformData(platform, repo, imageArtifactDetails))
-                                    .Where(platformMapping => platformMapping.Platform != null)
-                                    .Any(t => !t.Platform!.IsUnchanged))
+                                    .Where(platformMapping => platformMapping != null)
+                                    .Any(platformMapping => !platformMapping?.Platform.IsUnchanged ?? false))
                                 .Select(image => (repo, image)))
                         .ToList();
 
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
@@ -77,7 +77,13 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                         .SelectMany(repo =>
                             repo.FilteredImages
                                 .Where(image => image.SharedTags.Any())
-                                .Select(image => (repo, image)));
+                                .Where(image => image.AllPlatforms
+                                    .Select(platform =>
+                                        ImageInfoHelper.GetMatchingPlatformData(platform, repo, imageArtifactDetails))
+                                    .Where(platformMapping => platformMapping.Platform != null)
+                                    .Any(t => !t.Platform!.IsUnchanged))
+                                .Select(image => (repo, image)))
+                        .ToList();
 
                     Parallel.ForEach(manifests, ((RepoInfo Repo, ImageInfo Image) repoImage) =>
                     {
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
```diff
@@ -200,12 +200,13 @@ public static ImageArtifactDetails LoadFromFile(string path, ManifestInfo manife
         /// <param name="platform">Platform being searched.</param>
         /// <param name="repo">Repo that corresponds to the platform.</param>
         /// <param name="imageArtifactDetails">Image info content.</param>
-        public static (PlatformData Platform, ImageData Image)? GetMatchingPlatformData(PlatformInfo platform, RepoInfo repo, ImageArtifactDetails imageArtifactDetails)
+        #nullable enable
+        public static (PlatformData? Platform, ImageData? Image) GetMatchingPlatformData(PlatformInfo platform, RepoInfo repo, ImageArtifactDetails imageArtifactDetails)
```
mthalman, line 204: I don't understand why this change was made. I can understand if there was a case where one of the returned items was null and not the other. But that's not the case here. They're either both set, or nothing is returned. So the original implementation fits that better.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishManifestCommand.cs
```diff
@@ -77,7 +77,13 @@ await _registryCredentialsProvider.ExecuteWithCredentialsAsync(
                         .SelectMany(repo =>
                             repo.FilteredImages
                                 .Where(image => image.SharedTags.Any())
-                                .Select(image => (repo, image)));
+                                .Where(image => image.AllPlatforms
+                                    .Select(platform =>
+                                        ImageInfoHelper.GetMatchingPlatformData(platform, repo, imageArtifactDetails))
+                                    .Where(platformMapping => platformMapping.Platform != null)
+                                    .Any(t => !t.Platform!.IsUnchanged))
```
mthalman, line 84: This is another way to do this. I'm fine with either.

```suggestion
                                    .Any(t => t.Platform?.IsUnchanged == false))
```

src/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
lbussell: I made the change because I was having some issues with the nullability of the tuple and unwrapping it. But I figured it out, so I reverted the change (3058b39).

src/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
```diff
@@ -160,7 +160,7 @@ public static ImageArtifactDetails LoadFromContent(string imageInfoContent, Mani
                                 {
                                     imageData.ManifestImage = manifestImage;
                                 }
-
+
```
mthalman: nit: this can be reverted

mthalman (approved):

src/Microsoft.DotNet.ImageBuilder/src/ImageInfoHelper.cs
lbussell: Revert the revert? i.e. remove the whitespace?

lbussell: I'm just going to merge since there are no changes to this file in the current diff.

PR #1532, opened 2025-01-14
Add Dependabot config

lbussell:
Part of https://github.com/dotnet/dotnet-docker-internal/issues/6178.

I tested this out on my fork. It generated lots of updates initially:

https://github.com/lbussell/docker-tools/pulls

We can mess with the configuration further.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,14 +4,12 @@
 version: 2
 updates:
   - package-ecosystem: "nuget"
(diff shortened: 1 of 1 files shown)


lbussell:
Seems to be working, so this is ready for review. It'll submit lots of PRs at first until we tell dependabot which specific updates we want to ignore (for example System.CommandLine).

mthalman (approved):

PR #1575, opened 2025-01-29
Pipeline to import a custom image into an ACR

mthalman:
Pipeline that will allow us to import any specified image into the public mirror ACR.

### DIFF
--- a/eng/pipelines/import-image.yml
+++ b/eng/pipelines/import-image.yml
@@ -14,6 +14,16 @@ parameters:
 
 variables:
 - template: /eng/pipelines/templates/variables/common.yml@self
+- ${{ if eq(parameters.isDockerHubImage, 'true') }}:
+  - name: normalizedImageName
+    value: ${{ format('docker.io/{0}', parameters.imageName) }}
+  - name: extraImportOptions
+    value: --username $(dotnetDockerHubBot.userName) --password $(BotAccount-dotnet-dockerhub-bot-PAT)
+- ${{ else }}:
+  - name: normalizedImageName
+    value: ${{ parameters.imageName }}
+  - name: extraImportOptions
+    value: ''
 
 extends:
   template: /eng/common/templates/1es-official.yml@self
@@ -23,13 +33,6 @@ extends:
       jobs:
       - job: Import
         steps:
-        - pwsh: |
-            $normalizedImageName = '${{ parameters.imageName }}'
-            if ('${{ parameters.isDockerHubImage }}' -eq 'true') {
-              $normalizedImageName = 'docker.io/' + $normalizedImageName
-            }
-            Write-Output "##vso[task.setvariable variable=normalizedImageName]$normalizedImageName"
-          displayName: Normalize image name
         - template: /eng/common/templates/steps/run-pwsh-with-auth.yml@self
           parameters:
             displayName: Import image
@@ -41,3 +44,4 @@ extends:
               --name $(public-mirror.server)
               --source $(normalizedImageName)
               --image ${{ parameters.imageName }}
+              $(extraImportOptions)


src/Microsoft.DotNet.ImageBuilder/src/Commands/CopyCustomImageOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.CommandLine;
+
+#nullable enable
+namespace Microsoft.DotNet.ImageBuilder.Commands;
+
+public class CopyCustomImageOptions : Options
+{
+    public RegistryCredentialsOptions CredentialsOptions { get; set; } = new();
+
+    public string ImageName { get; set; } = string.Empty;
```
lbussell, line 14: Would it be valuable to support importing multiple images?

lbussell (commented):
Is there any functionality that this command will give us that we couldn't accomplish using `az acr import ...` directly in an AzDO pipeline?

mthalman:
> Is there any functionality that this command will give us that we couldn't accomplish using `az acr import ...` directly in an AzDO pipeline?

Excellent point. My answer is that I was too myopic and didn't consider that. I like that better.

lbussell:
> > Is there any functionality that this command will give us that we couldn't accomplish using `az acr import ...` directly in an AzDO pipeline?
> 
> Excellent point. My answer is that I was too myopic and didn't consider that. I like that better.

I can't say I blame you for wanting to write C# instead of yaml!

The AzDO task for Az CLI should handle all of the auth with managed identity, I think.

eng/pipelines/import-image.yml
```diff
@@ -0,0 +1,43 @@
+# Imports an image from a 3rd party registry to the public mirror registry
+
+trigger: none
+pr: none
+
+parameters:
+- name: imageName
+  displayName: Name of the image to be imported from 3rd party registry
```
lbussell, line 8: Can you give an example or two of acceptable inputs here, to make it easier to use? Like `anchore/syft:v1.XX` or `quay.io/centos/centos:stream10`. "Name" doesn't make it clear if we would accept images with or without registry, tag, etc.

For reference, similar pipeline definition for NuGet packages: https://dev.azure.com/dnceng/internal/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=931&branch=main

mthalman: Updated display name to clarify

eng/common/templates/variables/dotnet/common.yml
```diff
@@ -54,6 +54,6 @@ variables:
 - name: windowsServer2025PoolName
   value: Docker-2025-${{ variables['System.TeamProject'] }}
 
-- group: DotNet-Docker-Common
+- group: DotNet-Docker-Common-2
```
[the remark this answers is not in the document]
mthalman: This was used for testing. Reverted it

eng/pipelines/import-image.yml
```diff
@@ -0,0 +1,43 @@
...
+          parameters:
+            displayName: Import image
+            serviceConnection: $(public-mirror.serviceConnectionName)
+            command: >
+              az acr import
+              --subscription $(public-mirror.subscription)
+              --resource-group $(public-mirror.resourceGroup)
+              --name $(public-mirror.server)
+              --source $(normalizedImageName)
+              --image ${{ parameters.imageName }}
```
lbussell, line 43: This will also need the Docker Hub credentials or else it will be rate limited. Since it doesn't go through the local/CI machine we have no control over this either. When I use `acr import` locally I always get rate limited unless I use a DockerHub PAT.

mthalman: fixed

lbussell (approved):

PR #1601, opened 2025-02-19
Add devcontainer config

lbussell:
Related: https://github.com/dotnet/dotnet-docker/pull/6238

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/dotnet/docker-tools/pull/1601?shareId=b61153af-875c-470e-a214-3834000d7c56).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,11 @@
 {
     "image": "mcr.microsoft.com/dotnet/sdk:9.0-noble",
     "features": {
+        "ghcr.io/devcontainers/features/common-utils": {
+            "username": "app",
+            "userUid": 1654,
+            "userGid": 1654
+        },
         "ghcr.io/devcontainers/features/docker-in-docker:2": {}
     },
     "customizations": {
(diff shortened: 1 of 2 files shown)


mthalman (approved):

.devcontainer/devcontainer.json
```diff
@@ -14,11 +19,10 @@
...
                 "remote.otherPortsAttributes": {
                     "onAutoForward": "ignore"
-                },
-                "terminal.integrated.defaultProfile.linux": "bash"
+                }
             }
         }
     },
     "remoteUser": "app",
-    "updateContentCommand": "docker pull mcr.microsoft.com/dotnet-buildtools/image-builder:latest"
```
mthalman: Why get rid of this?

lbussell: I didn't think it would be particularly useful since most work in this repo is building ImageBuilder and running tests outside of Docker. But we can still add it back.

mthalman: That's true. And it would get pulled down automatically when needed anyway.

mthalman (approved):

.devcontainer/devcontainer.json
lbussell: I had debated removing it from dotnet-docker too. In both repos it only ends up being useful if the ImageBuilder image is on the latest version and not stuck waiting in a PR. But it's more useful when it works in that repo.

PR #1604, opened 2025-02-21
Add image/layer size to image-info and ingest in Kusto

### LINKED ISSUE #745: Ingest image/layer size in Kusto
Tracking the image/layer size information is useful for metric tracking purposes.  The wire/compressed size is the most useful to track.

lbussell:
Fixes #745

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/IngestKustoImageInfoCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/IngestKustoImageInfoCommand.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -99,9 +99,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
                         for (int i = 0; i < platform.Layers.Count; i++)
                         {
-                            // TODO: Track layer size (currently set to 0) https://github.com/dotnet/docker-tools/issues/745
                             layerInfo.AppendLine(FormatLayerCsv(
-                                platform.Layers[i], 0, platform.Layers.Count - i, sha, platform, image, repo.Repo, timestamp));
+                                platform.Layers[i].Digest, platform.Layers[i].Size, platform.Layers.Count - i, sha, platform, image, repo.Repo, timestamp));
                         }
                     }
                 }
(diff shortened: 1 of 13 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/IngestKustoImageInfoCommand.cs
```diff
@@ -117,7 +116,7 @@ private static string FormatImageCsv(string imageId, PlatformData platform, Imag
 
         private static string FormatLayerCsv(
             string layerDigest,
-            int size,
+            long size,
```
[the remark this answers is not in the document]
lbussell: The size in Kusto is already `long`.

mthalman (approved):

lbussell:
It has occurred to me that we might need to bootstrap this somehow. Since we changed the image-info model, the new version of ImageBuilder probably won't be able to deserialize any current image-info.json files.

lbussell:
At the same time, we'd need to update all image-infos and ImageBuilder versions at once, since the docker-tools repo handles queueing rebuilds and downstream repos handle publishing the image infos.

It might be better to update the model to support reading either the old or new model (list of string vs. list of Layer objects).

lbussell:
I think the correct approach for this would be to increment the SchemaVersion on ImageArtifactDetails:

https://github.com/dotnet/docker-tools/blob/71f2d6afc7f014edce35f9a0562c127116eedde7/src/Microsoft.DotNet.ImageBuilder/src/Models/Image/ImageArtifactDetails.cs#L9-L15

Then I think we'd implement a custom [JsonConverter<T>](https://www.newtonsoft.com/json/help/html/CustomJsonConverterGeneric.htm) for either `ImageArtifactDetails` or `PlatformInfo`. Need to think about it some more!

src/Microsoft.DotNet.ImageBuilder/src/Models/Image/ImageArtifactDetails.cs
```diff
@@ -2,18 +2,89 @@
...
+            JToken token = JToken.Load(reader);
+            return token.Type switch
+            {
+                // If the token is an object, proceed as normal
+                JTokenType.Object =>
+                    new Layer(
+                        Digest: token["digest"]?.Value<string>()
+                            ?? throw new JsonSerializationException(
+                                $"Unable to serialize Layer digest from '{token}'"),
+                        Size: token["size"]?.Value<long>() ?? 0),
```
mthalman, line 66: Instead of doing all this, can't we just let the deserializer do this work for us? Take the token string representation and run it through the deserializer to give us the `Layer` object.

mthalman: Yes, you'd need a separate instance without a converter.

PR #1617, opened 2025-03-04
Add back support for DockerfileGitCommitSha variable

sbomer:
We'd like to use this for https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1375.

@MichaelSimons PTAL

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
@@ -113,10 +113,6 @@ namespace Microsoft.DotNet.ImageBuilder.ViewModel
                 {
                     variableValue = getContextBasedSystemValue(variableType, variableName);
                 }
-                else
-                {
-                    variableValue = Options.GetOption(variableName);
-                }
             }
             else if (string.Equals(variableType, RepoVariableTypeId, StringComparison.Ordinal))
             {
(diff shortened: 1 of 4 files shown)


MichaelSimons:
>@MichaelSimons PTAL

Deferring to @mthalman and @lbussell as they are the current maintainers of the repo.

src/Microsoft.DotNet.ImageBuilder/src/ViewModel/VariableHelper.cs
```diff
@@ -105,7 +107,18 @@ public VariableHelper(Manifest manifest, IManifestOptionsInfo options, Func<stri
...
+            if (string.Equals(variableType, SystemVariableTypeId, StringComparison.Ordinal))
+            {
+                if (getContextBasedSystemValue != null)
+                {
+                    variableValue = getContextBasedSystemValue(variableType, variableName);
+                }
+                else
+                {
+                    variableValue = Options.GetOption(variableName);
+                }
```
lbussell, line 119: Correct me if I'm wrong but I don't think `Options.GetOption()` needs to be used if you only want to use `$(System:DockerfileGitCommitSha)`.

When getting a tag's value, if you're referencing a variable `$(System:foo)` then we'd like that to either be empty or throw. It wasn't being used before and I'd prefer not to add this extra option back to all commands. If we do need it, it should be scoped to a more specific sub-class of `Options`.

lbussell (approved):

Issue #1656, opened 2025-03-28
Use GitHub App authentication for PublishMcrDocs command


lbussell:
The Microsoft GitHub org is restricting PAT lifetimes and will eventually require the usage of fine-grained PATs. Current guidance is to migrate all interactions with the Microsoft GitHub org to use a GitHub App by the end of FY25.
 
### How to create the app

- [x] Create the app on personal github account
  - Permissions are defined at app creation time - this app probably just needs repo content write access
  - Record the App's Client ID somewhere - probably a pipeline variable group (it can be found on the App's management page)
- [ ] Transfer ownership to the `dotnet` org
- [ ] Add .NET containers team members as managers/admins of the App
- [ ] Get the app installed on the `microsoft` org
- [ ] Create a private key for the app and store it in our KeyVault
 
### How to authenticate as an app

1. Use the app's private key and Client ID to generate a JWT
    - Example: https://github.com/lbussell/gh-app-tester/blob/3455cf66bc08890ac8229fafbcd7c334d7bf6eb1/GitHubAppTester/GitHubHelper.cs#L51-L75
2. Exchange the JWT for an installation-specific token
3. Use the token to authenticate to GitHub as the app

Step 1 is not handled by OctoKit or Dotnet.VersionTools.Automation. For step 2, we could add support for the [installation token endpoint](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app) to VersionTools, or use OctoKit as a helper to exchange tokens. For step 3, we should be able to use the app installation token just like any other token with the VersionTools client. Migrating entirely to OctoKit is not necessary at this time.

### Related

- https://github.com/dotnet/docker-tools/issues/1637
- https://github.com/dotnet/dotnet-docker-internal/issues/7468

mthalman:
How is VersionTools involved here? AFAIK, ImageBuilder has no dependency on VersionTools. I see it has a package reference but it looks to be unused.

lbussell:
> How is VersionTools involved here? AFAIK, ImageBuilder has no dependency on VersionTools. I see it has a package reference but it looks to be unused.

It's used here, in PublishMcrDocs command: https://github.com/dotnet/docker-tools/blob/3e6ff3e63b45a28263a1b596b2a0e6f71c588f4d/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishMcrDocsCommand.cs#L62

[...]

PR #1660, opened 2025-04-03
Move GitHub authentication options to a separate class

lbussell:
This PR refactors how GitHub authentication options are handled in `ImageBuilder`. The goal of this refactor is to make changes for [Use GitHub App authentication for PublishMcrDocs command (#1656)](https://github.com/dotnet/docker-tools/issues/1656) easier to implement.

* [`src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs`](diffhunk://#diff-e25301fcace036c72944c530eab77fbde1f9d65978e13e53046644b312a460eaL17-R34): Replaced the use of `AuthToken` with `GitHubAuthOptions` to support both personal access tokens and private key files for GitHub authentication. [[1]](diffhunk://#diff-e25301fcace036c72944c530eab77fbde1f9d65978e13e53046644b312a460eaL17-R34) [[2]](diffhunk://#diff-e25301fcace036c72944c530eab77fbde1f9d65978e13e53046644b312a460eaL50-R52) [[3]](diffhunk://#diff-e25301fcace036c72944c530eab77fbde1f9d65978e13e53046644b312a460eaR94-L117)
* Added test project to Microsoft.DotNet.DockerTools solution (it was missing before)

### Token auth

```
> ImageBuilder <command> --github-auth token=abcdef123456
```

### Private key auth

Currently throws a NotImplementedException.

```
> ImageBuilder <command> --github-auth private-key-file=/path/to/my-key.pem
```

### Errors

What happens if you provide both private-key and auth-token:
```
Option '--github-auth' expects a single argument but 2 were provided.
```

What happens if auth is required but you provide no input:
```
Option '--github-auth' is required.
```

What happens if auth is required but you provide the wrong input:
```
> dotnet run -- myCommand --github-auth foo=bar

System.ArgumentException: GitHub token or private key file must be provided.
   at Microsoft.DotNet.ImageBuilder.Commands.GitOptionsBuilder
```


src/Microsoft.DotNet.ImageBuilder/src/StringHelper.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.ImageBuilder;
+
+public static class StringHelper
+{
+    public static string Format<T>(params IEnumerable<T> value)
```
MichaelSimons, line 7: This method strikes me as a little odd.  The name doesn't match what I expected it to do.  In my cursory search, I don't see it referenced.

MichaelSimons:
> Replaced AuthToken in GitOptions with a new GitHubAuthOptions class, which includes properties for AuthToken and PrivateKeyFilePath.

What is the benefit of doing this versus keeping the auth options as part of git options?  Is not everyplace that requires git options going to require auth?

src/Microsoft.DotNet.ImageBuilder/src/StringHelper.cs
lbussell: It was left over from a previous iteration of the changes - removed in https://github.com/dotnet/docker-tools/pull/1660/commits/92f2a279f0436c8e7598d647db99878da2992d52.

lbussell:
> > Replaced AuthToken in GitOptions with a new GitHubAuthOptions class, which includes properties for AuthToken and PrivateKeyFilePath.
> 
> What is the benefit of doing this versus keeping the auth options as part of git options? Is not everyplace that requires git options going to require auth?

It was an attempt at separation of concerns. My reasoning was that GitOptions need not know anything specific about how GitHub wants you to authenticate. But you are right that they occur together in each command except some cases where they are taken as an argument to a method.

Thinking about it again, another way that adding .pem key support could be achieved is by adding additional configuration options on GitOptionsBuilder.

Just spitballing here...

```cs
class GitHubAuth
{
    string? AccessToken { get; set; } = null;
    string? PrivateKeyFile { get; set; } = null;
    bool IsPrivateKeyAuth => PrivateKeyFile is not null;
}

var gitOptions =
    GitOptionsBuilder()
      .Defaults()
      .WithGitHubAuth(isRequired: true);

// Do the right thing based on what's in GitHubAuth
var patClient = OctokitClientFactory.Create(gitOptions.GitHubAuth);

// Some future undefined scenario...
var azdoOptions =
    new GitOptionsBuilder()
      .Defaults()
      .WithAzureDevOpsAuth();
```

lbussell:
> Your spitballing is very close to what I had in mind. The one thing I was pondering is if it would make sense to expose an auth option of sorts on the Default method (e.g. Defaults(AuthMode)). My thinking was that this could help in requiring an auth mode is defined.

OK, I rewrote it mostly as described in https://github.com/dotnet/docker-tools/pull/1660#issuecomment-2779017257. It is now more of a breaking change. We can update downstream pipeline code to use the new CLI option when the new ImageBuilder image is built.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
```diff
@@ -95,23 +91,65 @@ public GitOptionsBuilder WithRepo(
...
+                    string token = dictionary.GetValueOrDefault("token", "");
+                    string privateKeyFile = dictionary.GetValueOrDefault("private-key-file", "");
+
+                    if (isRequired)
+                    {
+                        if (string.IsNullOrEmpty(token) && string.IsNullOrEmpty(privateKeyFile))
+                        {
+                            throw new ArgumentException("GitHub token or private key file must be provided.");
+                        }
+                    }
```
mthalman, line 118:
```suggestion
                    if (isRequired && string.IsNullOrEmpty(token) && string.IsNullOrEmpty(privateKeyFile))
                    {
                        throw new ArgumentException("GitHub token or private key file must be provided.");
                    }
```

mthalman (approved):

MichaelSimons (approved):

Issue #1670, opened 2025-04-14
GitHub App auth CreateGitHubTokenAsync arbitrarily picks the first installation, potentially causing auth failure


dagood:
https://github.com/dotnet/docker-tools/blob/c9761380c45fec98dc6b355096fe20703354be03/src/Microsoft.DotNet.ImageBuilder/src/OctokitClientFactory.cs#L77-L78

In our Go tools, we always just take the installation ID as another auth-related parameter. We aren't using any installation listing API, so at the time we thought that we'd have to have our own mapping of org/repo -> installation ID, so we simply provide the installation ID to our tools.

I believe I encountered an error once when I used the wrong installation ID, because that installation didn't have permission for the target repository. (This might be something we're hitting with our attempt to use this feature, but I'm not set up to easily debug it.)

A few possible suggestions:

* Use `Single` rather than `First` (or something like that) to fail if there are too many installations rather than picking one arbitrarily. (But this might not support us--I'm not sure if we (still) have multiple installations.)
* Cross-reference the input target of the current operation to determine which installation to use.
* (Optionally?) Take the installation ID as an arg.
* Log the encountered installations to aid debugging when working with "real" secrets through the gloves of AzDO pipelines.

/cc @gdams

lbussell:
Thanks for the issue, I overlooked this. I meant to address this, potentially as a follow-up, but forgot. We should probably take installation ID as a required argument.

Installations should be logged as part of the InvalidOperationException in the same line of code you linked - are you not finding that to be the case?

dagood:
We don't actually get an `InvalidOperationException`--my hypothesis is that we actually have more than one installation and the wrong one is getting picked. More info (although it's just very vague exception text and a build link) in the internal issue: https://github.com/dotnet/dotnet-docker-internal/issues/7468#issuecomment-2802509376

dagood:
In any case, adding an arg for the installation ID would let me eliminate this possibility, so that sounds like a good step. 🙂

PR #1671, opened 2025-04-15
Add installation ID parameter for GitHub App auth

gdams:
fixes: https://github.com/dotnet/docker-tools/issues/1670

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
@@ -130,7 +130,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
             var installationIdOption = CreateOption<long?>(
                 InstallationIdAlias,
                 nameof(GitHubAuthOptions.InstallationId),
-                "GitHub App installation ID to use (optional; if not provided, it will be discovered automatically)");
+                "GitHub App installation ID to use (only required if app has more than one installation)");
 
             _options.AddRange([tokenOption, privateKeyOption, clientIdOption, installationIdOption]);
 
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
@@ -28,7 +28,8 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     public record GitHubAuthOptions(
         string AuthToken = "",
         string PrivateKey = "",
-        string ClientId = "")
+        string ClientId = "",
+        long? InstallationId = null)
     {
         public bool IsGitHubAppAuth =>
             !string.IsNullOrEmpty(PrivateKey) &&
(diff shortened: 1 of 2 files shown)


gdams:
@dotnet-policy-service agree company="Microsoft"

src/Microsoft.DotNet.ImageBuilder/src/OctokitClientFactory.cs
```diff
@@ -74,14 +74,15 @@ information about the App itself.
...
-                        {appInfo}
-                        """);
+                Installation? appInstallation;
+
+                if (authOptions.InstallationId is long providedId) {
+                    appInstallation = appInfo.Installations.FirstOrDefault(i => i.Id == providedId)
+                        ?? throw new InvalidOperationException($"No installation found with ID {providedId}.");
+                } else {
+                    appInstallation = appInfo.Installations.SingleOrDefault()
+                        ?? throw new InvalidOperationException("Expected exactly one installation for GitHub App but found none or multiple. Provide an installation ID explicitly.");
```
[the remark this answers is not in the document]
lbussell: Not helpful, let's ignore this one 😄

src/Microsoft.DotNet.ImageBuilder/src/OctokitClientFactory.cs
```diff
@@ -74,14 +74,15 @@ information about the App itself.
...
 
-                Installation? appInstallation = appInfo.Installations.FirstOrDefault()
-                    ?? throw new InvalidOperationException(
-                        $"""
-                        No installations found for GitHub App.
-                        Ensure the App is installed on the organization you are trying to access.
-
-                        {appInfo}
-                        """);
+                Installation? appInstallation;
```
lbussell, line 77: We can remove the nullable annotation here:

```suggestion
                Installation appInstallation;
```

src/Microsoft.DotNet.ImageBuilder/src/OctokitClientFactory.cs
lbussell, line 84: Nit: line over 120 chars
```suggestion
                        ?? throw new InvalidOperationException("Expected exactly one installation for GitHub App but "
                            + "found none or multiple. Provide an installation ID explicitly.");
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/GitOptions.cs
```diff
@@ -125,7 +126,13 @@ public GitOptionsBuilder WithGitHubAuth(string? description = null, bool isRequi
                 nameof(GitHubAuthOptions.ClientId),
                 "GitHub Client ID for GitHub App authentication");
 
-            _options.AddRange([tokenOption, privateKeyOption, clientIdOption]);
+            const string InstallationIdAlias = "gh-app-installation-id";
+            var installationIdOption = CreateOption<long?>(
+                InstallationIdAlias,
+                nameof(GitHubAuthOptions.InstallationId),
+                "GitHub App installation ID to use (optional; if not provided, it will be discovered automatically)");
```
lbussell, line 133: This could describe how the option is required if the app has more than one installation.

```suggestion
                "GitHub App installation ID to use (only required if app has more than one installation)");
```

src/Microsoft.DotNet.ImageBuilder/src/OctokitClientFactory.cs
```diff
@@ -74,14 +74,16 @@ information about the App itself.
...
-                        """);
+                Installation appInstallation;
+
+                if (authOptions.InstallationId is long providedId) {
+                    appInstallation = appInfo.Installations.FirstOrDefault(i => i.Id == providedId)
+                        ?? throw new InvalidOperationException($"No installation found with ID {providedId}.");
+                } else {
+                    appInstallation = appInfo.Installations.SingleOrDefault()
+                        ?? throw new InvalidOperationException("Expected exactly one installation for GitHub App but "
+                            + "found none or multiple. Provide an installation ID explicitly.");
```
dagood, line 85: I think that copilot review was actually right... `SingleOrDefault` doesn't return `null` when there is more than one, it actually throws its own `InvalidOperationException` exception. (`OrDefault` only applies to the 0 case.) So the exception message is accurate and the stack trace will lead you here if you follow it, but it won't show up in the >1 case.

But, clunky to fix--the .NET standard library doesn't have an expressive way to handle it IMO.

PR #1672, opened 2025-04-15
Fix GitHub App auth for Git push in PublishImageInfoCommand

gdams:
fixes: https://github.com/dotnet/dotnet-docker-internal/issues/7806

When using GitHub App credentials, the image builder correctly generates an installation access token but was not using it during Git operations. Instead, the push relied on GitHubAuthOptions.AuthToken, which is empty in GitHub App scenarios. This caused LibGit2Sharp to fail with repeated redirects due to missing or invalid credentials.

This change ensures the installation token is retrieved via CreateGitHubTokenAsync and injected into both CloneOptions and PushOptions as a UsernamePasswordCredentials with the required "x-access-token" format.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
@@ -47,13 +47,9 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
                 string authToken = await _octokitClientFactory.CreateGitHubTokenAsync(Options.GitOptions.GitHubAuthOptions);
                 CredentialsHandler credentials = GetCredentials(authToken);
 
-                CloneOptions cloneOptions = new()
+                CloneOptions cloneOptions = new(new() { CredentialsProvider = credentials })
                 {
                     BranchName = Options.GitOptions.Branch,
-                    FetchOptions = new FetchOptions
-                    {
-                        CredentialsProvider = credentials
-                    }
                 };
 
                 using IRepository repo = _gitService.CloneRepository(
@@ -112,7 +108,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         private static CredentialsHandler GetCredentials(string token) =>
             (_, _, _) => new UsernamePasswordCredentials
             {
-                Username = "x-access-token",
+                Username = "_", // A placeholder is required, but GitHub doesn't care what it is.
                 Password = token
             };
     }
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
@@ -16,19 +16,21 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
     public class PublishImageInfoCommand : ManifestCommand<PublishImageInfoOptions, PublishImageInfoOptionsBuilder>
     {
         private readonly IGitService _gitService;
+        private readonly IOctokitClientFactory _octokitClientFactory;
         private readonly ILoggerService _loggerService;
         private const string CommitMessage = "Merging Docker image info updates from build";
 
         [ImportingConstructor]
-        public PublishImageInfoCommand(IGitService gitService, ILoggerService loggerService)
+        public PublishImageInfoCommand(IGitService gitService, IOctokitClientFactory octokitClientFactory, ILoggerService loggerService)
         {
             _gitService = gitService ?? throw new ArgumentNullException(nameof(gitService));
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
```diff
@@ -42,17 +44,26 @@ public override Task ExecuteAsync()
...
 
-                using IRepository repo =_gitService.CloneRepository(
+                CloneOptions cloneOptions = new()
+                {
+                    BranchName = Options.GitOptions.Branch,
+                    FetchOptions = new FetchOptions
+                    {
+                        CredentialsProvider = credentials
+                    }
+                };
```
dagood, line 57: CI hit: `/image-builder/src/Commands/PublishImageInfoCommand.cs(53,21): error CS0200: Property or indexer 'CloneOptions.FetchOptions' cannot be assigned to -- it is read only [/image-builder/src/Microsoft.DotNet.ImageBuilder.csproj]`

I don't have this open in an IDE, but I think this is the general approach given https://github.com/libgit2/libgit2sharp/blob/50d697881e857ea0c78d63cc7f23fece3f526852/LibGit2Sharp/CloneOptions.cs#L15:

```suggestion
                CloneOptions cloneOptions = new(new() { CredentialsProvider = credentials })
                {
                    BranchName = Options.GitOptions.Branch,
                };
```

gdams:
I was able to build and test this locally which created the following commit in my fork: https://github.com/gdams/go-images-versions/commit/b097e0604d7398080f1784c2f8413c560375b19f

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
```diff
@@ -42,17 +44,23 @@ public override Task ExecuteAsync()
             {
                 _loggerService.WriteSubheading("Cloning GitHub repo");
 
-                CloneOptions cloneOptions = new() { BranchName = Options.GitOptions.Branch };
-                cloneOptions.FetchOptions.CredentialsProvider = GetCredentials(Options.GitOptions.GitHubAuthOptions);
+                string authToken = await _octokitClientFactory.CreateGitHubTokenAsync(Options.GitOptions.GitHubAuthOptions);
+                CredentialsHandler credentials = GetCredentials(authToken);
```
lbussell, line 48: Can we make `GetCredentials` non-static and put the call to `CreateGitHubTokenAsync` in there?

I also don't see a reason to re-format this code, so just changing `GetCredentials` could make this diff a fair bit smaller.

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
```diff
@@ -41,9 +43,9 @@ public override Task ExecuteAsync()
             try
             {
                 _loggerService.WriteSubheading("Cloning GitHub repo");
-
                 CloneOptions cloneOptions = new() { BranchName = Options.GitOptions.Branch };
-                cloneOptions.FetchOptions.CredentialsProvider = GetCredentials(Options.GitOptions.GitHubAuthOptions);
+                CredentialsHandler credentials = GetCredentials(authToken);
+                cloneOptions.FetchOptions.CredentialsProvider = credentials;
```
lbussell, line 48:
```suggestion
                cloneOptions.FetchOptions.CredentialsProvider = GetCredentials();
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
```diff
@@ -93,18 +94,21 @@ private void UpdateGitRepos(string repoPath, IRepository repo)
...
+                    CredentialsProvider = credentials
                 });
 
             Uri gitHubCommitUrl = GitHelper.GetCommitUrl(Options.GitOptions, commit.Sha);
             _loggerService.WriteMessage($"The '{Options.GitOptions.Path}' file was updated: {gitHubCommitUrl}");
         }
 
-        private static CredentialsHandler GetCredentials(GitHubAuthOptions gitHubAuthOptions) =>
-            (url, user, cred) => new UsernamePasswordCredentials
+        private async Task<CredentialsHandler> GetCredentials()
```
lbussell, line 104: Nit: async naming convention

```suggestion
        private async Task<CredentialsHandler> GetCredentialsAsync()
```

lbussell (approved):
Approved w/ one nit

src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs
gdams: updated PTAL

PR #1675, opened 2025-04-16
Add gitHubVersionsRepoInfo.authArgs configurable variable

dagood:
Add `gitHubVersionsRepoInfo.authArgs` that we can change in microsoft/go-images to use an app rather than a token.

Override tested with:

```yml
- name: gitHubVersionsRepoInfo.authArgs
  value: >-
    --gh-private-key '$(BotAccount-bot-for-go-private-key)'
    --gh-app-client-id '$(BotAccount-bot-for-go-client-id)'
    --gh-app-installation-id '$(BotAccount-bot-for-go-installation)'
```

https://dev.azure.com/dnceng/internal/_build/results?buildId=2688959&view=results

### DIFF
--- a/eng/common/templates/variables/dotnet/build-test-publish.yml
+++ b/eng/common/templates/variables/dotnet/build-test-publish.yml
@@ -14,8 +14,8 @@ variables:
 - name: officialRepoPrefixes
   value: public/,internal/private/,unlisted/
 
-- name: mcrDocsRepoInfo.accessToken
-  value: $(BotAccount-dotnet-docker-bot-PAT)
+- name: mcrDocsRepoInfo.authArgs
+  value: --gh-token '$(BotAccount-dotnet-docker-bot-PAT)'
 - name: mcrDocsRepoInfo.userName
   value: $(dotnetDockerBot.userName)
 - name: mcrDocsRepoInfo.email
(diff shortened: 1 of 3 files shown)


eng/common/templates/variables/dotnet/build-test-publish.yml
```diff
@@ -38,8 +38,8 @@ variables:
   value: main
 - name: gitHubVersionsRepoInfo.path
   value: ${{ variables.commonVersionsImageInfoPath }}
-- name: gitHubVersionsRepoInfo.accessToken
-  value: $(BotAccount-dotnet-docker-bot-PAT)
+- name: gitHubVersionsRepoInfo.authArgs
+  value: --gh-token '$(BotAccount-dotnet-docker-bot-PAT)'
```
lbussell, line 42: Do you plan to change the other options besides just `githubVersionsRepoInfo`?
e.g.

https://github.com/dotnet/docker-tools/blob/b6f70ca4d333ea28dcff877fce01a1d1430813ff/eng/common/templates/variables/dotnet/build-test-publish.yml#L17

https://github.com/dotnet/docker-tools/blob/b6f70ca4d333ea28dcff877fce01a1d1430813ff/eng/common/templates/variables/dotnet/build-test-publish.yml#L30

dagood: No, we don't use either of those. Happy to add indirects for those to this PR too, though, if it helps.

lbussell: That would be preferable! Thanks.

dagood: Done! The only `.accessToken`s in the codebase now are either `$(System.AccessToken)` or in unrelated C# code.

lbussell (approved):

PR #1702, opened 2025-05-07
Use Azure DevOps logger formatting commands

lbussell:
This is an attempt at making ImageBuilder's logging output easier to parse by using Azure Pipelines' built-in log formatting commands: https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#logging-command-format

The big one is the collapsible groups. If this makes any individual commands harder to read, we can adjust the output of those commands as necessary in follow-ups.

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Logger.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Logger.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#nullable enable
+
 using System;
 
 namespace Microsoft.DotNet.ImageBuilder
@@ -20,12 +22,7 @@ namespace Microsoft.DotNet.ImageBuilder
             Console.WriteLine(new string('-', heading.Length));
         }
 
-        public static void WriteMessage()
-        {
-            Console.WriteLine();
-        }
-
-        public static void WriteMessage(string message)
+        public static void WriteMessage(string? message = null)
         {
             Console.WriteLine(message);
         }
--- a/src/Microsoft.DotNet.ImageBuilder/src/Logger.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Logger.cs
@@ -10,7 +10,7 @@ namespace Microsoft.DotNet.ImageBuilder
     {
         public static void WriteError(string error)
         {
-            Console.Error.WriteLine(error);
+            Console.Error.WriteLine($"##[error]{error}");
         }
 
         public static void WriteHeading(string heading)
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.ImageBuilder/src/Logger.cs
```diff
@@ -2,37 +2,86 @@
...
+            WriteLine();
         }
+
+        /// <summary>
+        /// Ends a group in the log output with the specified message.
+        /// </summary>
+        /// <remarks>
+        /// This should only be called after WriteGroup to close the current group.
+        /// </remarks>
+        public static void EndGroup() => WriteLine($"##[endgroup]");
```
mthalman, line 67: I don't see any usage of this method. Isn't that necessary to have the group defined properly?

lbussell: I was making an assumption that starting a new group would automatically end the current group. Let me run a test to validate that and get back to you.

lbussell: If you do not end the group then it doesn't collapse. Also, groups start out collapsed. It would be good to put things with lots of verbose output in there (logging files, command line outputs, etc.).

https://dev.azure.com/dnceng/internal/_build/results?buildId=2704898&view=logs&j=792982d5-3bb1-5d82-222e-228148b73448&t=1f7e9bd0-3cdf-5b91-30b0-87c9b2af71b7

lbussell: (Review comment is out of date)

src/Microsoft.DotNet.ImageBuilder/src/LoggingGroup.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// <summary>
+    /// Creates a new collapsible logging group. When this object is created,
+    /// all subsequent logging output will be inside this group until this
+    /// object is disposed.
+    /// </summary>
+    /// <param name="groupName">The name of the logging group.</param>
+    public LoggingGroup(string name)
+    {
+        Console.WriteLine($"##[group]{name}");
```
mthalman, line 28: I expected that this would make use of `ILoggerService`. So probably need a method for `LoggerService` like the following and expose it via `ILoggerService` as well:

```c#
IDisposable UseLoggingGroup(string name) => new LoggingGroup(name, (ILoggerService)this);
```

lbussell: Sure, I think that makes sense.

lbussell:
Back to the drawing board with this one.

PR #1704, opened 2025-05-08
Add support for standalone publish pipeline

lbussell:
Part 1/2 of: https://github.com/dotnet/dotnet-docker-internal/issues/7414

- Make source build pipeline ID and pipeline run ID parameterized (this is the pipeline that we will be publishing images from)
- Add parameter for standalone publishes that results in the publish stage having no dependencies on 
other stages
- Added some extra command logging to some script steps
- Tried to make publish job source code easier to read by adding whitespace.


lbussell:
The changes to print the ImageBuilder commands before running them ended up resolving some variables too early which failed in the public where those variables aren't defined. Since those changes are unrelated to the goal of this PR, I reverted the change.

eng/common/templates/jobs/publish.yml
```diff
@@ -3,11 +3,14 @@ parameters:
   internalProjectName: null
   customInitSteps: []
   customPublishVariables: []
+  sourceBuildPipelineDefinitionId: $(System.DefinitionId)
+  sourceBuildPipelineRunId: $(sourceBuildId)
```
mthalman, line 7: Referencing variables for default values of parameters isn't supported.

eng/common/templates/steps/init-common.yml
```diff
@@ -3,7 +3,9 @@ parameters:
 
 steps:
 - powershell: |
-    $sourceBranch=$Env:BUILD_SOURCEBRANCH -replace "refs/heads/","" -replace "refs/tags/","" -replace "refs/pull/","" 
+    echo "Env:BUILD_SOURCEBRANCH: $Env:BUILD_SOURCEBRANCH"
+    $sourceBranch=$Env:BUILD_SOURCEBRANCH -replace "refs/heads/","" -replace "refs/tags/","" -replace "refs/pull/",""
+    echo "Setting sourceBranch to $sourceBranch"
```
mthalman, line 8: It may be more helpful to just enable diagnostic/debug (or whatever it's called) logging for the pipeline. That spits out the `##vso` echos so that these extra logging commands wouldn't be needed.

eng/common/templates/jobs/publish.yml
lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/1704/commits/38ef140281bb5320c6d4ffee984c0a5565fe1704 by parameterizing `sourceBuildId`

eng/common/templates/steps/init-common.yml
lbussell: During non debug builds I would still want to keep _some_ output for steps that only set variables. Are you suggesting that I use that for debugging or are you suggesting we turn it on for all pipeline runs?

mthalman: I'm suggesting it be turned on for all runs. This is currently done for the nightly-playground pipeline.

mthalman (approved):

Issue #1698, opened 2025-05-06
PostPublishNotification command should not depend on the displayName of individual stages


lbussell:
This is needed before [Add a common prefix to all pipeline stages that aren't injected by 1ES pipeline templates (#1696)](https://github.com/dotnet/docker-tools/issues/1696) can be implemented.

The PostPublishNotification command depends on the `displayName` of individual stages:

https://github.com/dotnet/docker-tools/blob/f4c1d804fe90a050da4e087481687d00d829fa93/eng/common/templates/jobs/publish.yml#L199-L208

Instead, it should depend on the `name` of the stages which is independent of the `displayName`.

PostPublishNotification gets the build timeline from the Azure DevOps API:

https://github.com/dotnet/docker-tools/blob/41c28c39060023a762fc3ee2d2b8bd399274b4ed/src/Microsoft.DotNet.ImageBuilder/src/Commands/PostPublishNotificationCommand.cs#L151

Here is an example timeline (internal link): https://dev.azure.com/dnceng/internal/_apis/build/builds/2703129/timeline

An individual task might look like this:

```jsonc
    {
      "previousAttempts": [],
      "id": "<guid>",
      "parentId": "<another guid>",
      "type": "Task",
      "name": "Publish Readmes", // this is the displayName
      "refName": "PublishReadmes", // this is the name
      "startTime": "2025-05-06T17:55:34.6033333Z",
      "finishTime": "2025-05-06T17:55:39.9666667Z",
      "currentOperation": null,
      "percentComplete": null,
      "state": "completed",
      "result": "succeeded",
      "resultCode": null,
      // ...bunch of other properties
    },
```

If we could get at the `refName` property, then we could use that instead of the `displayName`/`name`.

PR #1708, opened 2025-05-12
Add common prefix to all pipeline steps

### LINKED ISSUE #1696: Add a common prefix to all pipeline stages that aren't injected by 1ES pipeline templates
When reviewing a pipeline's output, it's difficult to determine which stages we control and which we don't. Lots of additional stages are injected by required pipeline templates.

https://github.com/dotnet/docker-tools/blob/f4c1d804fe90a050da4e087481687d00d829fa93/eng/common/templates/1es-official.yml#L40-L60

We should add some logic to inject a common prefix to each task/step's `displayName`. Something simple that stands out is an emoji like 🟪. (purple because that's .NET's color).

One problem that I see with this approach is that the `postPublishNotification` ImageBuilder command relies on the `displayName` of tasks to check if they were successful.

https://github.com/dotnet/docker-tools/blob/f4c1d804fe90a050da4e087481687d00d829fa93/eng/common/templates/jobs/publish.yml#L199-L208

That behavior should be based on the `name` of the steps, not the `displayName`. That way we can do whatever we want with `displayNames` without worrying about breaking publish notifications.

lbussell:
Fixes #1696

### DIFF
--- a/eng/common/templates/1es-official.yml
+++ b/eng/common/templates/1es-official.yml
@@ -38,48 +38,25 @@ resources:
     ref: refs/tags/release
 
 extends:
-  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+  template: /eng/common/templates/task-prefix-decorator.yml@self
   parameters:
-    pool: ${{ parameters.pool }}
-    sdl:
-      binskim:
-        enabled: true
-      componentgovernance:
-        ignoreDirectories: $(Build.SourcesDirectory)/versions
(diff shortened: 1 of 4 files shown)


lbussell:
> Doesn't this require updates to the names referenced here

Fixed in https://github.com/dotnet/docker-tools/pull/1708/commits/50e9bd9c0e88eeaca63e2b9a32ea29735f7fe996

Related: https://github.com/dotnet/docker-tools/issues/1698

mthalman (approved):

PR #1744, opened 2025-07-18
Use service connection to pull latest image info during publish

### LINKED ISSUE #1737: During publishing, "Download Latest Image Info" step can fail with 403 unauthorized error
https://github.com/dotnet/docker-tools/blob/fc626a41b1f991e761735cd36ac5c733ee74d6ff/eng/common/templates/jobs/publish.yml#L140-L147

This pipeline step is just a raw `curl` command which can fail sometimes causing our publish pipeline to be unreliable. In the example below it fails with a 403 unauthorized. Things to consider:

- At a minimum we should wrap this request in an exponential backoff retry
- Do we need to authenticate for this request? As our GitHub bot account?

Example pipeline where this is failing: https://dev.azure.com/dnceng/internal/_build/results?buildId=2746286&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=bf6a5b93-88b0-5310-04fe-dec22ee32d0e&s=380ab8c6-4139-55ff-15c4-fb48260f4dca

lbussell:
Fixes #1737 - prevents rate limiting for raw.githubusercontent.com by using a service connection to clone the repo and pull the latest changes instead. This should make publishing more reliable.

**Slight breaking change**: All publishing pipelines which publish image info now need to specify a repo resource. That's the cost of avoiding rate limiting.

Also simplified `internalVersionsRepoRef` and `publicVersionsRepoRef` to just `versionsRepoRef`. I couldn't see any reason they needed to be separate.

### DIFF
--- a/eng/pipelines/dotnet-buildtools-image-builder-official.yml
+++ b/eng/pipelines/dotnet-buildtools-image-builder-official.yml
@@ -18,19 +18,20 @@ parameters:
   type: string
   default: $(Build.BuildId)
 
-resources:
-  repositories:
-  - repository: VersionsRepo
-    type: github
-    endpoint: dotnet
-    name: dotnet/versions
-
 variables:
 - template: /eng/pipelines/templates/variables/image-builder.yml@self
   parameters:
     sourceBuildPipelineRunId: ${{ parameters.sourceBuildPipelineRunId }}
 - name: publishEolAnnotations
   value: true
+
+resources:
+  repositories:
+  - repository: VersionsRepo
+    type: github
+    endpoint: dotnet
+    name: dotnet/versions
+    ref: ${{ variables['gitHubVersionsRepoInfo.branch'] }}
 
 extends:
   template: /eng/common/templates/1es-official.yml@self
--- a/eng/pipelines/dotnet-buildtools-image-builder-pr.yml
+++ b/eng/pipelines/dotnet-buildtools-image-builder-pr.yml
@@ -9,15 +9,16 @@ pr:
 
 trigger: none
 
+variables:
+- template: templates/variables/image-builder.yml
+
 resources:
   repositories:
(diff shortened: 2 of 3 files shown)


eng/common/templates/1es-official.yml
```diff
@@ -59,6 +59,7 @@ extends:
           exclude:
           - repository: InternalVersionsRepo
           - repository: PublicVersionsRepo
```
mthalman: Why are these still needed?

eng/common/templates/steps/init-matrix-build-publish.yml
```diff
@@ -1,17 +1,21 @@
...
 
 steps:
 - checkout: self
-- ${{ if and(eq(variables['System.TeamProject'], 'public'), eq(parameters.noCache, false)) }}:
-  - checkout: ${{ parameters.publicVersionsRepoRef }}
-    path: s/${{ parameters.versionsRepoPath }}
-- ${{ if and(eq(variables['System.TeamProject'], 'internal'), eq(parameters.noCache, false)) }}:
-  - checkout: ${{ parameters.internalVersionsRepoRef }}
+- ${{ if ne(parameters.versionsRepoRef, '') }}:
+  - checkout: ${{ parameters.versionsRepoRef }}
```
mthalman, line 14: Can we target the necessary branch here, `$(gitHubVersionsRepoInfo.branch)`?

lbussell: That's done in the repository resource, but it can be done. You can only use template expressions to pick the `ref` though, unfortunately.

eng/common/templates/1es-official.yml
lbussell: Removed

mthalman (approved):

PR #1745, opened 2025-07-22
Remove duplicate manifest directory in Readme path in PublishMcrDocs command

lbussell:
This was a bug. It was causing issues when publishing readmes with the changes from https://github.com/dotnet/docker-tools/pull/1744. The Manifest directory is already prepended to the readme path in [ManifestInfo.Create](https://github.com/dotnet/docker-tools/blob/4617aa8afdeb9f27267e3f80dd7fc3f5cba1662e/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/ManifestInfo.cs#L85-L89). It was applied a second time in PublishMcrDocsCommand, causing publishing to fail because it couldn't find the readme file.

Related: https://github.com/dotnet/dotnet-docker/pull/6563


lbussell:
Successful pipeline test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2722469&view=results

Issue #1592, opened 2025-02-12
Add a `buildContext` option to Platform in Manifest


lbussell:
By default, ImageBuilder uses an explicit build context of the Dockerfile's location:

https://github.com/dotnet/docker-tools/blob/14b1496d071432e9776c93ad637479371bfec6c9/src/Microsoft.DotNet.ImageBuilder/src/ViewModel/PlatformInfo.cs#L59-L62

In some cases it's desirable to access content from outside the Dockerfile's directory. For example, when building .NET projects, you may need to access a `Directory.Build.Props` or `NuGet.config` located in a parent directory.

There are two ideas I came up with for implementing this:

## Option 1: explicit path

This option would allow you to set the explicit path to the build context. The path would be relative to the manifest.

```jsonc
{
  "dockerfile": "src/ImageBuilder/Dockerfile.linux",
  "buildContext": "path/to/my/preferred/context",
  "os": "linux",
  "osVersion": "azurelinux",
  "tags": {
    "linux-amd64": {},
    "linux-amd64-$(UniqueId)": {}
  }
},
```

## Option 2: limited choices

The other option would allow you to choose between having the build context be the Dockerfile's parent directory (the default, today) or the manifest's parent directory. This could be via an enum:

```jsonc
{
  "dockerfile": "src/ImageBuilder/Dockerfile.linux",
  "buildContext": "Manifest", // OR "Dockerfile"
  "os": "linux",
  "osVersion": "azurelinux",
  "tags": {
    "linux-amd64": {},
    "linux-amd64-$(UniqueId)": {}
  }
},
```

Or a boolean:

```jsonc
{
  "dockerfile": "src/ImageBuilder/Dockerfile.linux",
  "manifestRelativeBuildContext": "true", // "false" is the default
  "os": "linux",
  "osVersion": "azurelinux",
  "tags": {
    "linux-amd64": {},
    "linux-amd64-$(UniqueId)": {}
  }
},
```

PR #1750, opened 2025-07-24
Fix nullable reference warnings

lbussell:
This PR fixes all of the existing build warnings.

### DIFF
--- a/eng/src/yaml-updater/YamlUpdater.cs
+++ b/eng/src/yaml-updater/YamlUpdater.cs
@@ -41,7 +41,8 @@ namespace YamlUpdater
             Trace.Listeners.Add(safeTraceListener);
 
             string configJson = File.ReadAllText(options.ConfigPath);
-            FilePusher.Models.Config config = JsonConvert.DeserializeObject<FilePusher.Models.Config>(configJson);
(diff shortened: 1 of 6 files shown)


eng/src/file-pusher/file-pusher.csproj
```diff
@@ -6,6 +6,8 @@
     <RootNamespace>FilePusher</RootNamespace>
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
+    <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
mthalman, line 10: Add these in a `Directory.Build.props` directory at the repo root?

lbussell: It wouldn't be loaded in Dockerfile builds, because they're outside the scope of the Dockerfile build context for the individual projects. We could do this if we implemented https://github.com/dotnet/docker-tools/issues/1592.

mthalman: Oh yes, forgot about that.

mthalman (approved):

PR #1768, opened 2025-08-18
Set EOL annotation for staged images and mirrored base images

mthalman:
This sets an EOL annotation for all images pushed to staging and the mirrored base images. This ensures that won't be scanned by the container scanning system. It's not necessary to scan these images as they do not represent the latest supported set of images that are on MAR which is all we care about.

I've verified that importing an image from a staging repo to the public repo does not include the EOL annotation.

I did a bit of small refactoring too. So things are divided into separate commits.

Fixes https://github.com/dotnet/dotnet-docker-internal/issues/7673

### DIFF
--- a/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
+++ b/src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
@@ -28,6 +28,7 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
         private readonly IRegistryCredentialsProvider _registryCredentialsProvider;
         private readonly IAzureTokenCredentialProvider _tokenCredentialProvider;
         private readonly IImageCacheService _imageCacheService;
+        private readonly ILifecycleMetadataService _lifecycleMetadataService;
         private readonly ImageDigestCache _imageDigestCache;
         private readonly List<TagInfo> _processedTags = new List<TagInfo>();
         private readonly HashSet<PlatformData> _builtPlatforms = new();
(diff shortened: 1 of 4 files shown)


lbussell:
> I've verified that importing an image from a staging repo to the public repo does not include the EOL annotation.

@mthalman this is because the EOL annotations have separate manifests, right?

mthalman:
> > I've verified that importing an image from a staging repo to the public repo does not include the EOL annotation.
> 
> @mthalman this is because the EOL annotations have separate manifests, right?

Yes

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
```diff
@@ -735,6 +738,15 @@ private void PushImages()
...
                     _dockerService.PushImage(tag.FullyQualifiedName, Options.IsDryRun);
+
+                    // Immediately set the staged image as EOL to prevent it from being picked up by scanning.
+                    // We only care about scanning for the images hosted by MAR. Not all staged images make it to MAR,
+                    // and, for those that do, they may be replaced by more up-to-date images later on.
+                    _lifecycleMetadataService.AnnotateEolDigest(
+                        tag.FullyQualifiedName,
+                        DateOnly.FromDateTime(DateTime.UtcNow),
+                        Options.IsDryRun,
+                        out _);
```
lbussell, line 749: Do you have any concerns about not putting this behind a config option?

Otherwise, LGTM.

lbussell (approved):

src/Microsoft.DotNet.ImageBuilder/src/Commands/BuildCommand.cs
mthalman: I can't think of any scenarios where we wouldn't want it. You?

lbussell: Not off the top of my head, just figured I'd ask.

Issue #1747, opened 2025-07-23
ImageBuilder should consume a configuration file for build/publishing operations


lbussell:
## The problem

Configuring build, publishing, and pipeline settings for ImageBuilder is complex and needlessly complicated.

There are lots of inputs to the build/publishing process. This includes things like which pipeline image to run on as well as which registry to publish to.

In our system, pipeline inputs and build/publishing inputs are intertwined. Everything is configured through azure pipelines parameters and variables. This has a number of downsides:

- It is impossible to separate the publishing from the pipeline environment that it runs in
- It's difficult to track down where/how/why a particular pipeline variable was assigned
- We spend lots and lots of pipeline steps setting up variables inside PowerShell scripts
- Pipeline variables are [inherently insecure](https://learn.microsoft.com/en-us/azure/devops/pipelines/security/inputs?view=azure-devops#variables). Any compromised step can override variables by printing to stdout

For example, look at [`eng/common/templates/variables/common.yml`](https://github.com/dotnet/docker-tools/blob/29d405ebd61fe701c1a23713d1281636caf3c842/eng/common/templates/variables/common.yml). There is no logical separation between pipeline considerations (e.g. what pool we are running on) vs. publishing considerations (which ACR should we publish to? under which repo path?). Furthermore, Azure pipelines do not give us an easy way to separate those concerns.

Here's an example of a `build` command that ran recently:

```
build \
    --manifest dotnet-dotnet-docker/manifest.json \
    --os-version alpine3.21 \
    --os-type linux \
    --architecture amd64 \
    --retry \
    --digests-out-var 'builtImages' \
    --acr-subscription '00000000-0000-0000-0000-000000000000' \
    --acr-resource-group 'DotnetContainers' \
    --image-info-output-path /artifacts/imageInfo/linuxamd64src-runtime-deps-10.0-alpine3.21-graph-image-info.json \
    --source-repo https://github.com/dotnet/dotnet-docker \
    --source-repo-prefix mirror/ \
    --registry-override somerandomacr.azurecr.io \
    --repo-prefix build-staging/9999999/ \
    --push \
    --image-info-source-path versions/build-info/docker/image-info.dotnet-dotnet-docker-nightly.json
```

Hopefully it's obvious that lots of those arguments are a function of which publishing environment we're targeting (main, nightly, or testing). Ideally there would be one place for the entirety of those configurations to live.

## Config file

[...]

PR #1771, opened 2025-08-20
Add unofficial pipeline for image builder

lbussell:
I have settled on a pattern for sharing pipeline definitions between official, unofficial, and PR validation pipelines.

Official pipelines publish to production resources. Unofficial pipelines publish to non-production resources. This exercises much more of the pipeline than previously when we ran most publish steps as a dry-run. Both pipelines can share the same steps.

Some additional work is needed to port the official pipeline and PR validation pipeline to the newer format and ensure that PR validation works properly.


eng/common/templates/dotnet/base-pipeline-nonprod.yml
```diff
@@ -0,0 +1,64 @@
...
+
+  parameters:
+    ${{ insert }}: ${{ parameters.baseTemplate.parameters }}
+
+    internalProjectName: "internal"
+    publicProjectName: "public"
+
+    publishConfig:
+      internalMirrorAcr:
+        server: $(acr-staging-test.server)
```
mthalman: Is there any isolation of variables between prod and non-prod?

eng/common/templates/jobs/test-images-linux-client.yml
```diff
@@ -20,9 +21,14 @@ jobs:
   pool: ${{ parameters.pool }}
   timeoutInMinutes: ${{ parameters.testJobTimeout }}
   steps:
+  - powershell: |-
+      Write-Host 'Parameters for eng/common/templates/jobs/test-images-linux-client.yml'
+      Write-Host '${{ convertToJson(parameters) }}'
+    displayName: Print parameters
```
mthalman: Delete?

eng/common/templates/stages/build-and-test.yml
```diff
@@ -58,21 +59,27 @@ stages:
...
-        - powershell: |
-            echo "##vso[task.setvariable variable=productVersion]"
-            echo "##vso[task.setvariable variable=imageBuilderPaths]"
-            echo "##vso[task.setvariable variable=osVersions]"
-            echo "##vso[task.setvariable variable=architecture]"
-          displayName: Initialize Test Variables
+      - powershell: |-
+          Write-Host 'Parameters for eng/common/templates/stages/build-and-test.yml'
+          Write-Host '${{ convertToJson(parameters) }}'
+        displayName: Print parameters
```
mthalman: Delete?

eng/common/templates/steps/test-images-windows-client.yml
```diff
@@ -49,7 +50,7 @@ steps:
   displayName: Test Images
   condition: and(succeeded(), ${{ parameters.condition }})
 - ${{ if and(eq(variables['System.TeamProject'], parameters.internalProjectName), ne(variables['Build.Reason'], 'PullRequest')) }}:
-  - script: docker logout $(acr-staging.server)
+  - script: docker logout ${{ parameters.publishconfig.buildacr.server }}
```
mthalman:
```suggestion
  - script: docker logout ${{ parameters.publishConfig.buildAcr.server }}
```

lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/1771/commits/832401055d3c98f2645e93e046f7b7f155f8009f

eng/common/templates/jobs/test-images-linux-client.yml
lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/1771/commits/8d3eb2283c06ac8e6bbd671181c1edb7865d7c97

eng/common/templates/stages/build-and-test.yml
lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/1771/commits/8d3eb2283c06ac8e6bbd671181c1edb7865d7c97

eng/common/templates/dotnet/base-pipeline-nonprod.yml
lbussell: Not currently. But since variable groups are referenced in the pipeline definition only, nothing is stopping us from doing this after the fact.

eng/pipelines/dotnet-buildtools-image-builder-unofficial.yml
```diff
@@ -32,13 +45,13 @@ resources:
...
+    baseTemplate:
+      template: /eng/pipelines/templates/pipelines/dotnet-buildtools-image-builder.yml@self
       parameters:
-        noCache: true
-        internalProjectName: ${{ variables.internalProjectName }}
-        publicProjectName: ${{ variables.publicProjectName }}
         sourceBuildPipelineRunId: ${{ parameters.sourceBuildPipelineRunId }}
         versionsRepoRef: VersionsRepo
+        baseTemplate:
+          template: /eng/common/templates/1es.yml@self
```
lbussell: This "layering" or "injection" approach is the most recent addition to these changes. I'm not that attached to it. Before this the `publishConfig` was defined inline at the root pipeline level.

One other option I considered was something like this:

```yaml
extends:
  template: /eng/common/templates/dotnet/base-pipeline-nonprod.yml@self
  parameters:
    sourceBuildPipelineRunId: ${{ parameters.sourceBuildPipelineRunId }}
    stages:
    - template: /eng/pipelines/templates/stages/dotnet-buildtools-image-builder.yml@self
      parameters:
        sourceBuildPipelineRunId: ${{ parameters.sourceBuildPipelineRunId }}
```

Then base-pipeline-nonprod PR would need to iterate through all of the stages defined in dotnet-buildtools-image-builder and pass in the `publishConfig` parameter to each stage manually. But that requires us to assume that `publishConfig` is accepted a as a parameter for all those stages, which might not be the case for all pipelines that want to use this pattern. So I chose this direction because it allows the core pipeline to do whatever it wants with the PublishConfig.

Downside of this approach is, `dotnet-buildtools-image-builder.yml` kind of needs to assume that it's extending the 1es template. Hence the condition in that file for declaring service connections.

Perhaps `base-pipeline-nonprod.yml` could take a stage template as a parameter instead of an extends template as a parameter. I'll hack that up really quick and see if it works.

lbussell: This has been refactored to use stage templates instead of extends templates. I think it is much easier to follow the logic as a result.

eng/pipelines/dotnet-buildtools-image-builder-official.yml
```diff
@@ -17,6 +17,11 @@ parameters:
     The default value should be left alone if you want to build new images.
   type: string
   default: $(Build.BuildId)
+- name: publishRepoPrefix
+  displayName: >
+    Publish repo prefix. Final images will be pushed under this prefix in the
+    production ACR. Examples: `public/`, `test/`
+  default: "public/"
```
mthalman: Why do we want this to be a parameter for official? Shouldn't it always be `public/`?

lbussell: I added it so that I could do a test run. But I suppose you won't be able to do that in the near future. So I'm fine with removing the parameter.

mthalman (approved):

lbussell:
One small change (https://github.com/dotnet/docker-tools/pull/1771/commits/4818c33d6d0229b71ff644a7beb973d4547cf9b5) was needed to facilitate unofficial pipelines that run without `noCache: true`. Even if we don't plan to publish image info anywhere, we still want to checkout the Versions repo since we need it to properly simulate caching images.

lbussell:
/cc @dagood, fairly large breaking change here. We are slowly moving more and more pipeline configuration to parameters. ACRs and service connections are now defined in this publishConfig object: https://github.com/dotnet/docker-tools/blob/main/eng/common/templates/stages/dotnet/publish-config-prod.yml. The idea is that most of the things that don't change between builds can all be defined in one place. With some templating magic you can easily swap between production/testing environments as well, which makes implementing unofficial pipelines easier.

The publishConfig schema/structure isn't officially defined anywhere at the moment, but you can treat the `dotnet/` implementations of it as the source of truth.

Related/follow-up work:
- #1747

dagood:
Nice, yeah, we added a `ctx` parameter that seems similar, to help with reuse while splitting. I wrote down some notes about my recent work on the non-images Go pipelines:

https://github.com/microsoft/go-infra/blob/main/docs/pipeline-yml-style.md#compile-time-variable-usage

https://github.com/microsoft/go-infra/blob/main/docs/pipeline-yml-style.md#pipeline-templates-to-reduce-duplication-with-officialunofficial-split

I actually also added some `variable`s to our own pipelines when I realized that I can make them compile-time safe by passing them through to `ctx`. [Used an `each`](https://github.com/microsoft/go/blob/ec38a984f92a8a9f3b526e0616b8462b04fb3ba1/eng/pipeline/rolling-internal-pipeline.yml#L109-L122) to avoid a duplicate-list problem.

I haven't poked around to see if any of that would be useful here (or what's already been done 🙂) but thought it might be worth sharing.

PR #1778, opened 2025-08-25
Command to mark non-prod images as EOL

mthalman:
Contributes to https://github.com/dotnet/dotnet-docker-internal/issues/7673

This adds a `generateEolAnnotationDataForNonProd` command that will annotate all images in a registry as EOL. This is intended for the staging and cache registries. This prevents images in those registries from showing up in vulnerability reports. A separate pipeline that runs on a schedule will be used to execute this command.


src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommandBase.cs
```diff
@@ -0,0 +1,134 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands;
+
+#nullable enable
+
+public abstract class GenerateEolAnnotationDataCommandBase<TOptions, TOptionsBuilder>
+    : Command<TOptions, TOptionsBuilder>
+    where TOptions : GenerateEolAnnotationDataOptions, new()
+    where TOptionsBuilder : GenerateEolAnnotationDataOptionsBuilder, new()
```
lbussell, line 21: Did you consider a composition approach instead of an inheritance approach for this problem? e.g. pseudocode below.

It doesn't seem like we're getting a lot from the inheritance in this case, besides creating two nearly identical commands.

```C#
class GenerateEolAnnotationDataCommand
{
    Execute()
    {
        // Instantiate rule based on Options
        // Then call rule.GetDigestsToAnnotate()
    }
}

class GenerateEolAnnotationDataOptions
{
    EolAnnotationRule Mode { get; set; }
}

interface IEolAnnotationRule
{
    IEnumerable<EolDigestData> GetDigestsToAnnotate()
}

class AllImagesEolRule : IEolAnnotationRule { ... }
class UnsupportedImagesEolRule : IEolAnnotationRule { ... }
enum EolAnnotationRule { AllImages, UnsupportedImages }
```

mthalman: I had thought about it but didn't go with that approach since the command options are different. For the publish scenario, there are required args for registry creds, old image info and new image info. But those don't apply to the non-prod scenario.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForPublishCommand.cs
```diff
@@ -0,0 +1,98 @@
...
+    public override async Task ExecuteAsync() =>
+        await _registryCredentialsProvider.ExecuteWithCredentialsAsync(Options.IsDryRun, async () =>
+            {
+                IEnumerable<EolDigestData> digestsToAnnotate = await GetDigestsToAnnotate();
+                WriteDigestDataJson(digestsToAnnotate);
+            },
+            Options.CredentialsOptions,
+            registryName: Options.RegistryOptions.Registry,
+            ownedAcr: Options.RegistryOptions.Registry,
+            serviceConnection: Options.AcrServiceConnection);
```
lbussell: We don't need credentials in order to run `WriteDigestDataJson(...)`, right? Only to get the digests (and potentially enumerate them?).

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForPublishCommand.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    private async Task<IEnumerable<EolDigestData>> GetDigestsToAnnotate()
```
lbussell: I know you didn't change the name of these methods from the original class, but since we're modifying them, can we fix the names that don't follow the async naming convention?

```suggestion
    private async Task<IEnumerable<EolDigestData>> GetDigestsToAnnotateAsync()
```

Applies to several methods in this PR.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForNonProdCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    protected override string Description => "Generate EOL annotation data for non-prod images";
+
+    public override async Task ExecuteAsync()
+    {
+        // All images in all repos of the registry are marked as unsupported.
+        Dictionary<string, string?> registryTagsByDigest = await GetAllImageDigestsFromRegistry();
+        IEnumerable<EolDigestData> eolDigests = GetUnsupportedDigests(registryTagsByDigest, []);
+        WriteDigestDataJson(eolDigests);
+    }
```
lbussell: If we're sticking with inheritance, then I think `ExecuteAsync()` should be a sealed method on `GenerateEolAnnotationDataCommandBase` which calls an abstract method `GetDigestsToAnnotate()` that's defined by the subclasses. That way, the responsibility of the base class vs. subclasses is more clearly defined.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForPublishCommand.cs
```diff
@@ -0,0 +1,98 @@
...
+            loggerService,
+            tokenCredentialProvider,
+            acrContentClientFactory,
+            acrClientFactory,
+            lifecycleMetadataService)
+    {
+        _registryCredentialsProvider = registryCredentialsProvider ?? throw new ArgumentNullException(nameof(registryCredentialsProvider));
+    }
+
+    protected override string Description => "Generate EOL annotation data for publish stage";
```
lbussell: Is this a more accurate description?

```suggestion
    protected override string Description => "Generate EOL annotation data for all images that exist in OldImageInfoPath and not in NewImageInfoPath";
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForNonProdCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+        : base(
+            loggerService,
+            tokenCredentialProvider,
+            acrContentClientFactory,
+            acrClientFactory,
+            lifecycleMetadataService)
+    {
+    }
+
+    protected override string Description => "Generate EOL annotation data for non-prod images";
```
lbussell:
```suggestion
    protected override string Description => "Generate EOL annotation data for all images in a registry";
```

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommandBase.cs
lbussell: Fine with me - resolving this conversation then.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForPublishCommand.cs
mthalman: Fixed with https://github.com/dotnet/docker-tools/pull/1778/commits/2ee44fa4e12e97778fa22de01f1d6b86cffbb69c

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForPublishCommand.cs
mthalman: Fixed with https://github.com/dotnet/docker-tools/pull/1778/commits/b3cd7c551164111c22eaff57e97a54276631f672

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForNonProdCommand.cs
mthalman: Fixed with https://github.com/dotnet/docker-tools/pull/1778/commits/1570ea9e7b6ed0370c412a57c93429a7d64ff504

I also determined that the registry options are required for both commands so I included that with this commit.

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForPublishCommand.cs
mthalman: Fixed with https://github.com/dotnet/docker-tools/pull/1778/commits/77dd5b9f9f1461a271ee1bb550039f1e3e8df938

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForNonProdCommand.cs
mthalman: Fixed with https://github.com/dotnet/docker-tools/pull/1778/commits/77dd5b9f9f1461a271ee1bb550039f1e3e8df938

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataCommandBase.cs
```diff
@@ -0,0 +1,152 @@
...
+
+    public sealed override async Task ExecuteAsync()
+    {
+        IEnumerable<EolDigestData> digestsToAnnotate = [];
+        await _registryCredentialsProvider.ExecuteWithCredentialsAsync(Options.IsDryRun,
+            async () => digestsToAnnotate = await GetDigestsToAnnotateAsync(),
+            Options.CredentialsOptions,
+            registryName: Options.RegistryOptions.Registry,
+            ownedAcr: Options.RegistryOptions.Registry,
+            serviceConnection: Options.AcrServiceConnection);
```
lbussell:
```suggestion
        await _registryCredentialsProvider.ExecuteWithCredentialsAsync(
            Options.IsDryRun,
            async () => digestsToAnnotate = await GetDigestsToAnnotateAsync(),
            Options.CredentialsOptions,
            registryName: Options.RegistryOptions.Registry,
            ownedAcr: Options.RegistryOptions.Registry,
            serviceConnection: Options.AcrServiceConnection);
```

nit: formatting

src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateEolAnnotationDataForAllImagesCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+    }
+
+    protected override string Description => "Generate EOL annotation data for all images in the registry";
+
+    protected override async Task<IEnumerable<EolDigestData>> GetDigestsToAnnotateAsync()
+    {
+        // All images in all repos of the registry are marked as unsupported.
+        Dictionary<string, string?> registryTagsByDigest = await GetAllImageDigestsFromRegistryAsync();
+        return GetDigestsWithoutExistingAnnotation(GetUnsupportedDigests(registryTagsByDigest, []));
```
lbussell: Can we skip the `Where(...)` call incurred by `GetUnsupportedDigests()`? Then that method isn't needed on the base abstract class either.

Edit: Something's bugged with the suggestion here, but you get the idea.

```suggestion
        IEnumerable<EolDigestData> digests = registryTagsByDigest
            .Select(kvp => new EolDigestData(kvp.Key) { Tag = kvp.Value });
        return GetDigestsWithoutExistingAnnotation(digests);
```

mthalman: Fixed with https://github.com/dotnet/docker-tools/pull/1778/commits/9323a04fa6e94ca37ff433e6c8eee195d0e9efd9.

I also found that the call to `GetDigestsWithoutExistingAnnotation` can be moved to the base.

lbussell (approved):

PR #1821, opened 2025-09-25
Fix ARG pattern to exclude invalid hyphen character

mthalman:
- Hyphens are not valid characters in Docker ARG names per Dockerfile specification
- Updated regex to only allow valid characters
- Added unit test to verify ARG pattern correctly validates character classes

### DIFF
--- a/src/ImageBuilder.Tests/PlatformInfoTests.cs
+++ b/src/ImageBuilder.Tests/PlatformInfoTests.cs
@@ -45,15 +45,16 @@ namespace Microsoft.DotNet.ImageBuilder.Tests
         }
 
         [Theory]
-        [InlineData("VALID", "ubuntu:latest", "ubuntu:latest")]     // Full match and substitution
-        [InlineData("VALID_123", "alpine:latest", "alpine:latest")]     // Full match and substitution
+        [InlineData("VALID", "ubuntu:latest", "$VALID", "ubuntu:latest")]
+        [InlineData("VALID_123", "alpine:latest", "$VALID_123", "alpine:latest")]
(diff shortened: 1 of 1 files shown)


src/ImageBuilder/ViewModel/PlatformInfo.cs
```diff
@@ -22,7 +22,7 @@ public class PlatformInfo
 
         private static Regex FromRegex { get; } = new Regex($@"FROM\s+(--platform=.*?\s+)?(?<{FromImageMatchName}>\S+)(\s+AS\s+(?<{StageIdMatchName}>\S+))?");
 
-        private static readonly string s_argPattern = $"\\$(?<{ArgGroupName}>[\\w\\d-_]+)";
+        private static readonly string s_argPattern = $"\\$(?<{ArgGroupName}>[\\w\\d_]+)";
```
lbussell: While we're here can you convert this to a source generated partial property? https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators

I think there are examples elsewhere in the codebase.

lbussell: I also think it would be valuable to link to the Dockerfile documentation/spec behind this pattern.

mthalman: Fixed https://github.com/dotnet/docker-tools/pull/1821/commits/1260bd8901dd66717b0839f9038589519764ad88

mthalman: Docker doesn't have a hard spec on this that can be referenced, other than the source code of the parser.

lbussell (approved):

PR #1880, opened 2025-12-09
Use Microsoft.Extensions.Configuration to load ACR Service Connection info at Runtime

### LINKED ISSUE #1747: ImageBuilder should consume a configuration file for build/publishing operations

lbussell:
Fixes #1747

This PR introduces a configuration-based approach for managing ACR references and service connections. The changeset is big, but mostly straightforward.

- `Microsoft.Extensions.Configuration` is used to load `appsettings.json` at runtime. Configuration can be loaded or overridden from [many different sources](https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration#configuration-providers).
- The `PublishConfiguration` property in `appsettings.json` is serialized to a [`PublishConfiguration` object](https://github.com/lbussell/docker-tools/blob/3f27a6f46eb0ca4999585a3fdb91ae9bbfeb8502/src/ImageBuilder/Configuration/PublishConfiguration.cs#L10-L34) when ImageBuilder starts up: https://github.com/lbussell/docker-tools/blob/3f27a6f46eb0ca4999585a3fdb91ae9bbfeb8502/src/ImageBuilder/Configuration/ConfigurationExtensions.cs#L16-L18
- Then, any service will be able to resolve an `IOptions<PublishConfiguration>` via dependency injection: https://github.com/lbussell/docker-tools/blob/3f27a6f46eb0ca4999585a3fdb91ae9bbfeb8502/src/ImageBuilder/RegistryResolver.cs#L15
- As a result, for the ACRs defined in `PublishConfiguration`, we no longer need to pass them as command line arguments nor do we need to pass them between functions.

All of the other changes follow logically from there.

- Removed `--acr-service-connection` CLI option from BuildOptions, CopyImagesOptions, CleanAcrImagesOptions, GenerateBuildMatrixOptions, AnnotateEolDigestsOptions, PublishManifestOptions.
- No longer need to pass service connections or registries around between methods.
- No more concept of "owned registries". We can just [query the PublishConfiguration itself](https://github.com/lbussell/docker-tools/blob/3f27a6f46eb0ca4999585a3fdb91ae9bbfeb8502/src/ImageBuilder/RegistryResolver.cs#L35-L41) about what registries we have service connections for.
- Added RegistryResolver to centralize logic around determining whether we have a service connection to a given registry.
- Added a couple of additional abstractions ([ImageName](https://github.com/lbussell/docker-tools/blob/3f27a6f46eb0ca4999585a3fdb91ae9bbfeb8502/src/ImageBuilder/ImageName.cs#L7), [Acr](https://github.com/lbussell/docker-tools/blob/3f27a6f46eb0ca4999585a3fdb91ae9bbfeb8502/src/ImageBuilder/Configuration/Acr.cs#L13)) to make reasoning about various strings easier. They handle parsing and normalizing references to those items.

Pipeline changes will come in a follow-up because otherwise this PR wouldn't pass validation. They have already been prepared and tested. They exist on this branch: https://dev.azure.com/dnceng/internal/_git/dotnet-docker-tools?version=GBdev/loganbussell/me-configuration&_a=history. Successful pipeline run: [part 1](https://dev.azure.com/dnceng/internal/_build/results?buildId=2854891), [part 2](https://dev.azure.com/dnceng/internal/_build/results?buildId=2854919)

### DIFF
--- a/eng/Version.Details.props
+++ b/eng/Version.Details.props
@@ -6,7 +6,7 @@ This file should be imported by eng/Versions.props
 <Project>
   <PropertyGroup>
     <!-- dotnet/arcade dependencies -->
-    <MicrosoftDotNetArcadeSdkPackageVersion>11.0.0-beta.25610.3</MicrosoftDotNetArcadeSdkPackageVersion>
+    <MicrosoftDotNetArcadeSdkPackageVersion>11.0.0-beta.25612.6</MicrosoftDotNetArcadeSdkPackageVersion>
   </PropertyGroup>
   <!--Property group for alternate package version names-->
   <PropertyGroup>
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -3,9 +3,9 @@
   <ProductDependencies>
   </ProductDependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="11.0.0-beta.25610.3">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="11.0.0-beta.25612.6">
       <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>27e190e2a8053738859c082e2f70df62e01ff524</Sha>
+      <Sha>8adf115288fa51feaa30d063b946478054c7f7b4</Sha>
     </Dependency>
   </ToolsetDependencies>
 </Dependencies>
--- a/eng/common/tools.ps1
+++ b/eng/common/tools.ps1
@@ -556,19 +556,26 @@ function LocateVisualStudio([object]$vsRequirements = $null){
     })
   }
 
-  if (!$vsRequirements) { $vsRequirements = $GlobalJson.tools.vs }
+  if (!$vsRequirements) {
+    if (Get-Member -InputObject $GlobalJson.tools -Name 'vs' -ErrorAction SilentlyContinue) {
+      $vsRequirements = $GlobalJson.tools.vs
+    } else {
+      $vsRequirements = $null
+    }
+  }
+
   $args = @('-latest', '-format', 'json', '-requires', 'Microsoft.Component.MSBuild', '-products', '*')
 
   if (!$excludePrereleaseVS) {
     $args += '-prerelease'
   }
 
-  if (Get-Member -InputObject $vsRequirements -Name 'version') {
+  if ($vsRequirements -and (Get-Member -InputObject $vsRequirements -Name 'version' -ErrorAction SilentlyContinue)) {
     $args += '-version'
     $args += $vsRequirements.version
   }
 
-  if (Get-Member -InputObject $vsRequirements -Name 'components') {
+  if ($vsRequirements -and (Get-Member -InputObject $vsRequirements -Name 'components' -ErrorAction SilentlyContinue)) {
     foreach ($component in $vsRequirements.components) {
       $args += '-requires'
       $args += $component
--- a/eng/docker-tools/DEV-GUIDE.md
+++ b/eng/docker-tools/DEV-GUIDE.md
@@ -1,0 +1,387 @@
+# Developer Guide: Using the docker-tools Infrastructure
+
+This guide walks you through the practical scenarios and workflows for using the docker-tools infrastructure. The `eng/docker-tools` directory is a **shared infrastructure layer** used across all .NET Docker repositories (dotnet-docker, dotnet-buildtools-prereqs-docker, dotnet-framework-docker). It solves a fundamental challenge: building, testing, and publishing Docker images across multiple operating systems (Alpine, Ubuntu, Azure Linux, Windows Server variants), multiple CPU architectures (amd64, arm64, arm32), and multiple .NET versions—all while maintaining consistency and reliability.
+
+At its core, the infrastructure provides:
+
+- **PowerShell scripts** for local image building and Docker operations—so you can test Dockerfile changes on your machine before committing
+- **Azure Pipelines templates** for CI/CD (build, test, publish)—a composable template system that orchestrates builds across dozens of OS/architecture combinations in parallel
+- **ImageBuilder orchestration**—a specialized .NET tool that understands manifest files, manages image dependencies, handles multi-arch manifest creation, and coordinates the entire build process
+- **Caching and optimization**—intelligent systems that skip unchanged images and minimize redundant work
+- **SBOM generation**—automatic Software Bill of Materials creation for supply chain security
+
+The infrastructure handles complexity that would otherwise be overwhelming: a single commit to a repo can trigger builds of hundreds of image variants across Linux and Windows agents, each requiring proper build sequencing, testing, and eventual publication to Microsoft Artifact Registry (MAR).
+
+**Important:** Files in `eng/docker-tools/` are synchronized across repositories by automation in the [dotnet/docker-tools](https://github.com/dotnet/docker-tools) repository. If you need to make changes to this infrastructure, submit them there—changes made directly in consuming repos will be overwritten.
+
+---
+
+## Local Development Scenarios
+
+### Scenario: Building Docker Images Locally
+
+The most common local task is building images to test Dockerfile changes before pushing.
+
+**Quick Build - All Images:**
+```powershell
+./eng/docker-tools/build.ps1
+```
+
+**Filter by OS:**
+```powershell
+# Build only Alpine images
+./eng/docker-tools/build.ps1 -OS "alpine"
+
+# Build Ubuntu 24.04 images
+./eng/docker-tools/build.ps1 -OS "noble"
+```
+
+**Filter by Architecture:**
+```powershell
+# Build arm64 images only
+./eng/docker-tools/build.ps1 -Architecture "arm64"
+```
+
+**Filter by Path:**
+```powershell
+# Build images from a specific directory
+./eng/docker-tools/build.ps1 -Paths "src/runtime/8.0/alpine3.20"
+
+# Build all 8.0 runtime images using glob pattern
+./eng/docker-tools/build.ps1 -Paths "*runtime*8.0*"
+```
+
+**Combine Filters:**
+```powershell
+# Build .NET 8.0 Alpine arm64 images
+./eng/docker-tools/build.ps1 -Version "8.0" -OS "alpine" -Architecture "arm64"
+```
+
+**Filter by Product Version (if applicable):**
+```powershell
+# Build only .NET 8.0 images
+./eng/docker-tools/build.ps1 -Version "8.0"
+
+# Build .NET 6.0 and 8.0 images
+./eng/docker-tools/build.ps1 -Version "6.0","8.0"
+```
+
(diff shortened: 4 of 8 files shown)


src/ImageBuilder/Configuration/AcrConfiguration.cs
```diff
@@ -0,0 +1,31 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#nullable enable
+
+namespace Microsoft.DotNet.ImageBuilder.Configuration;
+
+public sealed record AcrConfiguration
+{
+    // TODO: This was temporarily renamed to Server to match existing YAML configuration.
+    // See publish-config-prod.yml and publish-config-prod.yml.
```
mthalman: `publish-config-prod.yml` is listed twice here.

src/ImageBuilder/Configuration/PublishConfiguration.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        if (InternalMirrorAcr is not null)
+        {
+            yield return InternalMirrorAcr;
+        }
+
+        if (PublicMirrorAcr is not null)
+        {
+            yield return PublicMirrorAcr;
+        }
```
mthalman:
```suggestion
        return [ BuildAcr, PublishAcr, InternalMirrorAcr, PublicMirrorAcr].Where(acr => acr is not null);
```

src/ImageBuilder/Configuration/ServiceConnection.cs
```diff
@@ -0,0 +1,21 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#nullable enable
+
+namespace Microsoft.DotNet.ImageBuilder.Configuration;
+
+public record ServiceConnection : IServiceConnection
+{
+    /// <inheritdoc/>
+    public string Name { get; set; } = "";
```
mthalman, line 15: Doesn't look like this property is used.

src/ImageBuilder/Configuration/AcrConfiguration.cs
```diff
@@ -0,0 +1,31 @@
...
+    // It should be renamed back to Repo once the YAML files are updated.
+    public string? Server { get; set; } = null;
+    public string? ResourceGroup { get; set; } = null;
+    public string? Subscription { get; set; } = null;
+    public string? RepoPrefix { get; set; } = null;
+    public ServiceConnection? ServiceConnection { get; set; } = null;
+}
+
+public static class AcrConfigurationExtensions
```
mthalman: What's the need for this? Why not define the `ToAcr` method directly in `AcrConfiguration`?

src/ImageBuilder/Configuration/PublishConfiguration.cs
lbussell: The one-liner does not work because `There is no target type for the collection expression (CS9176)`. Closest you can get is:

```cs
AcrConfiguration?[] acrs = [BuildAcr, PublishAcr, InternalMirrorAcr, PublicMirrorAcr];
return acrs.Where(a => a is not null)!; // don't like the null-forgiving operator here
```

or:

```cs
AcrConfiguration?[] acrs = [BuildAcr, PublishAcr, InternalMirrorAcr, PublicMirrorAcr];
return acrs.OfType<AcrConfiguration>(); // better
```

mthalman: I like the second option. It would need a comment explaining that `OfType` is used to filter out nulls.

src/ImageBuilder/Commands/AnnotateEolDigestsCommand.cs
```diff
@@ -55,12 +55,9 @@ await _registryCredentialsProvider.ExecuteWithCredentialsAsync(
                 () =>
                 {
                     Parallel.ForEach(eolAnnotations.EolDigests, digestData => AnnotateDigest(digestData, globalEolDate));
```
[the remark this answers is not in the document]
lbussell: This uses a different overload that takes `Action` instead of `Func<Task>`

src/ImageBuilder/Configuration/RegistryConfiguration.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// If the registry is an ACR, this is the Azure subscription that the ACR belongs to.
+    /// </summary>
+    public string? Subscription { get; set; } = null;
+
+    /// <summary>
+    /// E.g. "public/" or "prefix-". A delimiter (slash/dash/etc.) is not automatically added to
+    /// the end of this prefix. It should be included if necessary.
+    /// </summary>
+    public string? RepoPrefix { get; set; } = null;
```
lbussell: So this `RepoPrefix` is defined both in `RegistryConfiguration` and also in `ManifestOptions`. This one is currently ignored. That is pretty bad because it could lead to some unexpected behavior like pushing to the wrong prefix. Ideally we would use one or the other, but ManifestOptions.RepoPrefix is really tightly coupled to the manifest filtering. I think the right thing to do is probably remove this property for now, and eventually replace ManifestCommand/ManifestOptions with some kind of service for resolving/querying and filtering the manifest.json.

mthalman: I agree with that.

lbussell: Done in https://github.com/dotnet/docker-tools/pull/1880/commits/5fd62db7c89e27940a1b373abd9b61453b9dc44c

src/ImageBuilder/Configuration/ServiceConnection.cs
lbussell: I added a comment explaining why it is included in https://github.com/dotnet/docker-tools/pull/1880/commits/b5fc9c182db24d624443a8691e7de1c59399a6b4

src/ImageBuilder/Configuration/PublishConfiguration.cs
lbussell: Done in https://github.com/dotnet/docker-tools/pull/1880/commits/a94cf389cfeded94c4325aaac68c19275e543520

src/ImageBuilder/Configuration/AcrConfiguration.cs
lbussell: I didn't see a compelling reason to include it in the record itself. If the record definition only includes the properties, then it's easier to glance at and reason about what kinds of configurations are possible/acceptable via appsettings.json. I could see an argument for moving this to `ConfigurationExtensions`.

I would have done the same for `PublishConfiguration.GetKnownAcrConfigurations()`, but I decided that method was better suited in the class since it needs to stay in sync with all of the AcrConfiguration properties.

lbussell:
Not sure why the Windows build is still failing, because I added to the `InterceptorsNamespaces` property in the csproj exactly as the error message says. Same build error still. Windows only.

PR #1899, opened 2025-12-19
Add tests for Manifest serialization and deserialization

lbussell:
This PR adds tests that assert all of the functionality around ImageBuilder's `Manifest` models. They make sure required properties are actually required, ensure that string-enum conversion works correctly, and ensure that serialization/deserialization is consistent in both directions as well as round-trip serialization.

With these tests in place, we can be more confident with the following types of changes in the future:
- Migrating from Newtonsoft.Json to System.Text.Json
- Moving the Manifest models to the ImageBuilder.Models package
- Enabling nullable annotations
- Changing to immutable models/records

Also included some minor formatting changes:
- The file header template in .editorconfig didn't match up with lots of files. I counted the occurrences of each header and changed the .editorconfig to match the more common one.
- Fixed the file headers in ImageBuilder Manifest models
- Converted ImageBuilder Manifest models to file-scoped namespaces

### DIFF
--- a/src/ImageBuilder.Tests/Models/ImageSerializationTests.cs
+++ b/src/ImageBuilder.Tests/Models/ImageSerializationTests.cs
@@ -28,13 +28,11 @@ public class ImageSerializationTests
     }
 
     [Fact]
-    public void FullyPopulatedImage_Serialization()
+    public void FullyPopulatedImage_Bidirectional()
     {
-        // Bidirectional test not possible: empty Platforms array is required for deserialization
-        // but omitted during serialization
         Image image = new()
         {
-            Platforms = [], // Leave sub-model arrays empty per instructions
+            Platforms = [],
             SharedTags = new Dictionary<string, Tag>
             {
                 ["8.0"] = new Tag(),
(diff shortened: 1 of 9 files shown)


src/ImageBuilder.Tests/Models/ImageSerializationTests.cs
```diff
@@ -0,0 +1,180 @@
...
+        Image image = new();
+
+        AssertSerializationFails(image, nameof(Image.Platforms));
+    }
+
+    [Fact]
+    public void FullyPopulatedImage_Serialization()
+    {
+        // Bidirectional test not possible: empty Platforms array is required for deserialization
+        // but omitted during serialization
```
mthalman: Can't it be updated to force being serialized?

lbussell: I updated the serialization behavior. So now, it has the following behaviors:

- Required properties are *always* serialized no matter what.
- Otherwise, do not serialize default values, null values, and empty lists.

Tests have been updated accordingly.

PR #1900, opened 2025-12-19
Update ImageBuilder and update pipelines to use appsettings.json for configuration

lbussell:
This PR is a follow-up to #1880. It updates the pipelines to use appsettings.json to configure service connections instead of passing them as command line arguments.

It includes the ImageBuilder update from #1894. I am submitting this as a separate PR since I don't want automated builds to unexpectedly overwrite my progress.

Test runs:
- ~~build#2864259 - https://github.com/dotnet/docker-tools/pull/1900/commits/5b86d60f3ba3182df8ff4717e532d357c41cfa62~~
- ~~build#2864324 - https://github.com/dotnet/docker-tools/pull/1900/commits/00c96077dc2b07256cac3be41de49ee8cde1852b~~
- build#2865810 - https://github.com/dotnet/docker-tools/pull/1900/commits/d85e9143c8f059b3602bd777d7eafe8697800ae5 - passing

### DIFF
--- a/eng/docker-tools/templates/steps/init-docker-linux.yml
+++ b/eng/docker-tools/templates/steps/init-docker-linux.yml
@@ -42,11 +42,10 @@ steps:
     displayName: Pull Image Builder
     condition: and(succeeded(), ${{ parameters.condition }})
 
-  - ${{ if and(parameters.publishConfig, ne(variables['Build.Reason'], 'PullRequest')) }}:
-    - template: /eng/docker-tools/templates/steps/generate-appsettings.yml@self
-      parameters:
-        publishConfig: ${{ parameters.publishConfig }}
-        condition: ${{ parameters.condition }}
+  - template: /eng/docker-tools/templates/steps/generate-appsettings.yml@self
+    parameters:
+      publishConfig: ${{ parameters.publishConfig }}
+      condition: ${{ parameters.condition }}
 
   - script: >-
       docker build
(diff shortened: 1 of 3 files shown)


eng/docker-tools/templates/steps/init-docker-windows.yml
```diff
@@ -37,6 +38,21 @@ steps:
...
+    - powershell: |-
+        $appsettingsJsonContent = @"
+        {
+          "PublishConfiguration": ${{ convertToJson(parameters.publishConfig) }}
+        }
+        "@
+        Set-Content -Path "appsettings.json" -Value $appsettingsJsonContent
+        Get-Content -Path "appsettings.json"
+      displayName: Output publish configuration
+      condition: and(succeeded(), ${{ parameters.condition }})
```
mthalman: It looks like this has the same implementation as init-docker-linux.yml. This can be factored out into a step template file instead.

lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/1900/commits/b6fc71a87f69c7c103d306177d0c52b1d0a097c5, testing in build#2872503

eng/docker-tools/templates/steps/init-docker-linux.yml
```diff
@@ -29,6 +42,12 @@ steps:
     displayName: Pull Image Builder
     condition: and(succeeded(), ${{ parameters.condition }})
 
+  - ${{ if and(parameters.publishConfig, ne(variables['Build.Reason'], 'PullRequest')) }}:
```
mthalman: How about including this condition in the template too?

Issue #1905, opened 2025-12-22
`GetAllImageDigestsFromRegistryAsync` is suddenly extremely slow


lbussell:
In [this dotnet-docker-nightly pipeline run ](https://dev.azure.com/dnceng/internal/internal%20Team/_build/results?buildId=2863803&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&s=380ab8c6-4139-55ff-15c4-fb48260f4dca&t=ed10bff5-3e0f-5624-de44-7c9d29af573e&l=118)it took 1hr15min, causing the pipeline to time out later in the run:

```
2025-12-19T01:01:15.0763844Z Querying registry for all image digests...
2025-12-19T02:15:43.0489446Z Checking digest for existing annotation: ...
```

Previously ([in this pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2858424&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=a08adfce-16dd-580a-541c-89ea6031d6c4)), it took less than one minute:

```
2025-12-10T21:56:14.9556032Z Querying registry for all image digests...
2025-12-10T22:19:31.3493702Z Checking digest for existing annotation: ...
```

The earliest occurrence of this behavior I could find is in https://dev.azure.com/dnceng/internal/_build/results?buildId=2861673&view=logs&j=739bb7ec-7296-5b40-7dff-ee3297bfdaa4&t=a08adfce-16dd-580a-541c-89ea6031d6c4 where it took 23 minutes. That is dotnet-docker commit db838f02ef8fb99154680b938054748dc856d8b9. 

The cause is likely #1879.

PR #1906, opened 2025-12-22
Add timed logging to GetAllImageDigestsFromRegistryAsync

lbussell:
This PR adds better logging to GetAllImageDigestsFromRegistryAsync, in an attempt to diagnose #1905, which is causing publishing failures due to pipeline timeouts (https://github.com/dotnet/dotnet-docker-internal/issues/9694)


lbussell:
This should wait until we implement #1908

PR #1932, opened 2026-01-14
Make copilot instructions more concise

lbussell:
This PR is targeting https://github.com/dotnet/docker-tools/pull/1893, because I wanted to propose changes without completely taking over the PR...

I ran the copilot-instructions file through my [create-skill](https://github.com/lbussell/agent-skills/blob/main/skills/create-skill/SKILL.md) skill, which is based on a bunch of official recommendations from Anthropic. I think what came out the other end is much more clear and concise. I also made a few manual tweaks to it afterwards.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,165 +1,35 @@
 # GitHub Copilot Instructions for dotnet/docker-tools
 
-This repository contains common tools for .NET Docker repositories, with the primary tool being **ImageBuilder** - a specialized .NET application for building, testing, and publishing Docker images across multiple platforms and architectures.
+Common tools for .NET Docker repositories, primarily **ImageBuilder** - a CLI tool for building, testing, and publishing Docker images across multiple platforms and architectures.
 
-## Repository Structure
+## Key Documentation
 
-- **`src/`** - Contains the ImageBuilder tool and its components
-  - `ImageBuilder/` - Main tool source code (C# .NET)
-  - `ImageBuilder.Models/` - Data models and manifest schema
-  - `ImageBuilder.Tests/` - Unit tests
-  - `Dockerfile.linux` and `Dockerfile.windows` - ImageBuilder container images
-- **`eng/docker-tools/`** - Shared infrastructure synchronized across all .NET Docker repositories
-  - **`DEV-GUIDE.md`** - **CRITICAL**: Comprehensive developer guide for using the docker-tools infrastructure (see Developer Guide section below)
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -1,165 +1,35 @@
...
-  - **`DEV-GUIDE.md`** - **CRITICAL**: Comprehensive developer guide for using the docker-tools infrastructure (see Developer Guide section below)
-  - PowerShell scripts for local development (`build.ps1`, `Invoke-ImageBuilder.ps1`)
-  - Azure Pipelines templates for CI/CD
-- **`documentation/`** - User-facing documentation
-  - `manifest-file.md` - Manifest schema documentation
-  - `base-image-dependency-flow.md` - Base image update automation
+- **[README.md](../README.md)** - Building the repo locally
+- **[src/README.md](../src/README.md)** - Building ImageBuilder container image, available commands
+- **[eng/docker-tools/DEV-GUIDE.md](../eng/docker-tools/DEV-GUIDE.md)** - Local development workflows, pipeline architecture, CI/CD patterns
+- **[documentation/manifest-file.md](../documentation/manifest-file.md)** - Manifest schema documentation
```
mthalman, line 10: I'm curious how an agent handles relative paths. Does it keep the context of knowing which directory it got the prompt from so that it can compute these paths relative to that directory? Or is it better to make these paths relative from the root?

.github/copilot-instructions.md
```diff
@@ -1,165 +1,35 @@
...
-- **Stages control**: Use `stages` variable to run specific pipeline stages
-
-## Key Files to Keep in Sync
-
-When updating functionality, these files often need coordinated changes:
-
-1. **ImageBuilder source code** (`src/ImageBuilder/`)
-2. **Pipeline templates** (`eng/docker-tools/templates/`)
-3. **PowerShell scripts** (`eng/docker-tools/*.ps1`)
-4. **DEV-GUIDE.md** - Document changes here!
```
mthalman, line 148: The new changes don't mention anything about keeping this file in sync.

.github/copilot-instructions.md
lbussell: According to the VSCode GitHub Copilot "Chat Debug" view - yes, it does. It includes them all as attachments directly beneath the instructions. That could be both a good or bad thing depending on what behavior you want.

lbussell: image

PR #1947, opened 2026-01-23
Add "bootstrap" option to ImageBuilder unofficial pipeline

lbussell:
**Background**: Testing ImageBuilder changes end-to-end (code + pipelines) requires building and pushing an image to a registry and manually updating the ImageBuilder code reference for every code change. This limits how quickly you can iterate on larger ImageBuilder changes.

**This PR**: adds a bootstrapImageBuilder parameter to the imagebuilder-unofficial pipeline. When enabled, ImageBuilder is built from source at the start of every job. This enables validating ImageBuilder source changes and pipeline template changes together in a single pipeline run.

Linux builds the container directly using `docker build`, tags the image, and overrides the imagebuilder image variable so that it's used for later steps. Windows builds directly using .NET, and puts the output in the place that subsequent steps expect. In this sense, it's different from the pipeline build, since it's not built *using* ImageBuilder. That's why it's a "bootstrap". I determined that this would be OK because the pipeline itself will then proceed to validate that build ImageBuilder can build ImageBuilder.

**Other options I considered:**
- Conditionally adding a build stage before the first build stage to build ImageBuilder just in time. This won't work because *that* stage might be broken due to pipeline changes.
- Separate pipeline that runs the ImageBuilder image build and triggers another pipeline to validate the changes with that version of ImageBuilder. It's the same problem though - if you make breaking changes to the pipeline, then one of the two will fail.

**Other changes:**
- Combined `init-docker-linux` and `init-docker-windows` templates into one.
- The bootstrap code and options are all only in this repo and won't affect consuming repos. This is through a new parameter that allows overriding the imagebuilder initialization steps.

**Testing:**
- Build, Post-Build, Test: build#2886215
- Publish: build#2886405

### DIFF
--- a/eng/docker-tools/templates/jobs/build-images.yml
+++ b/eng/docker-tools/templates/jobs/build-images.yml
@@ -116,7 +116,7 @@ jobs:
     - template: /eng/docker-tools/templates/jobs/${{ format('../steps/test-images-{0}-client.yml', parameters.dockerClientOS) }}@self
       parameters:
         condition: ne(variables.testScriptPath, '')
-        skipInit: true
+        skipCommonInit: true
   - ${{ if and(eq(variables['System.TeamProject'], parameters.internalProjectName), ne(variables['Build.Reason'], 'PullRequest'), eq(parameters.dockerClientOS, 'linux')) }}:
     - template: /eng/docker-tools/templates/steps/publish-artifact.yml@self
       parameters:
--- a/eng/docker-tools/templates/steps/init-common.yml
+++ b/eng/docker-tools/templates/steps/init-common.yml
@@ -16,11 +16,6 @@ parameters:
   type: boolean
   default: true
 
-# Whether to set up the Test Runner environment
-- name: setupTestRunner
-  type: boolean
-  default: false
-
 # Whether to delete existing Docker images
 - name: cleanupDocker
   type: boolean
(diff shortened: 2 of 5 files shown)


eng/docker-tools/templates/jobs/build-images.yml
```diff
@@ -6,6 +6,7 @@ parameters:
   buildJobTimeout: 60
   commonInitStepsForMatrixAndBuild: []
   customInitSteps: []
```
mthalman, line 7: Since we already have `customInitSteps`, can we make use of that instead of introducing `customImageBuilderSetupSteps`? In that case `customInitSteps` could be defaulted to just call `init-docker.yml`.

lbussell: Yeah. This would be a breaking change, but after doing some searching, I don't see any usage of `customInitSteps` by consuming repos. So that sounds reasonable to me.

lbussell: Okay, that was awful.
- e4d8de9d implemented this change, but CopyBaseImages was broken because our ImageBuilder init step relied on the `repoRoot` variable. That was defined in `init-matrix-build-publish.yml`, which wasn't used in CopyBaseImages. It used only `init-docker.yml`.
- `init-docker.yml` didn't have any way to reason about whether or not we checked out another repo. It directly relied on variables set by `init-matrix-build-publish.yml`.
- I iterated on this several times, but the only way that I could really reason about everything together was to combine all of the init templates into one. That's the next three commits. I believe this really just unveils a lot of complexity that was hiding behind lots of templates relying on each others' global variables.
- Feel free to reject this change on the premise that `init-common.yml` is too large and complicated.

eng/docker-tools/templates/steps/init-common.yml
```diff
@@ -1,9 +1,356 @@
...
+# - Pull the ImageBuilder image from MCR
+# - Build a "withrepo" image that layers the source code into the container
+# - Generate docker run commands that mount Docker socket and artifact paths
+# Windows: Runs as extracted executable
+# - Pull the ImageBuilder image, create a temporary container, and copy out
+#   the executable to Build.BinariesDirectory
+# - Run directly as .exe (no container) because Windows containers have
+#   limitations with Docker-in-Docker and volume mounts
+# Custom setup (customInitSteps) overrides both Linux and Windows setup steps.
+- ${{ if eq(parameters.setupImageBuilder, true) }}:
```
mthalman, line 241: Since this is a long conditional set of steps, I think it would benefit by being factored out into a separate template.

eng/docker-tools/templates/jobs/build-images.yml
mthalman: I don't mind these changes at all. You organized and commented the init-common file very well. I'm in favor of this approach.

eng/docker-tools/templates/steps/init-common.yml
lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/1947/commits/c12c9f16416111318f7306a68ec68d964b060f1e

Test run: build#2888758

mthalman (approved):

PR #1950, opened 2026-01-28
Add C# GitHub Copilot instructions

lbussell:
Loosely adapted from:
- https://github.com/dotnet/dotnet-docker/blob/main/.github/instructions/csharp.instructions.md
- https://github.com/github/awesome-copilot/blob/main/agents/CSharpExpert.agent.md
- https://github.com/lbussell/agent-skills/blob/main/skills/csharp-review/SKILL.md

### DIFF
--- a/.github/instructions/csharp.instructions.md
+++ b/.github/instructions/csharp.instructions.md
@@ -9,9 +9,6 @@ applyTo: "**/*.cs"
 For all new C# code:
 
 - Use file-scoped namespaces.
-- Use nullable reference types
-    - Use `<Nullable>enable</Nullable>` in project files for new projects.
-    - When adding code to existing files without nullable reference types enabled, add `#nullable enable` followed by `#nullable restore` in source files.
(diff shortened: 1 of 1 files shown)


.github/instructions/csharp.instructions.md
```diff
@@ -0,0 +1,60 @@
...
+# C#/.NET Guidelines
+
+## Coding style
+
+For all new C# code:
+
+- Use file-scoped namespaces.
+- Use nullable reference types
+    - Use `<Nullable>enable</Nullable>` in project files for new projects.
+    - When adding code to existing files without nullable reference types enabled, add `#nullable enable` followed by `#nullable restore` in source files.
```
mthalman: How about we set all existing projects to be NRT-enabled. For all of the errors that pop up, add `#nullable disable` for those instances. This would be a good job for Copilot to work on. Then we don't need any special instructions around NRT.

lbussell: Sure. I've started https://github.com/dotnet/docker-tools/pull/1951, and I will simplify this part of the instructions.

PR #1967, opened 2026-02-13
Install ORAS directly from Azure Linux package feed

lbussell:
The version is still 1.3.0.

@mthalman do we need to adjust anything about the renovate config since we removed the env var and renovate comment?


mthalman:
> @mthalman do we need to adjust anything about the renovate config since we removed the env var and renovate comment?

It should just no-op now that that stuff is removed.

PR #1972, opened 2026-02-18
Re-apply dry-run fix for CopyImageService publish config lookup

lbussell:
## Problem

PR #1966 fixed `CopyImageService.ImportImageAsync` to skip the `PublishConfiguration` lookup in dry-run mode. This fix was inadvertently reverted by the logging migration in #1968 due to a bad merge.

This causes `CopyBaseImages` and `CopyAcrImages` to fail in PR validation pipelines with:
```
System.InvalidOperationException: No authentication details found for source registry '$(acr-staging-test.server)'.
Ensure the registry is configured in the publish configuration.
```

In PR builds, `appsettings.json` is intentionally not generated (the `generate-appsettings.yml` step skips for PRs), so `PublishConfiguration.RegistryAuthentication` is empty. The `GetRegistryResource` call was happening before the `isDryRun` check, so even dry-run mode required a fully populated config.

## Changes

1. **Test** — Added `CopyImageServiceTests` with a test that exercises `ImportImageAsync` in dry-run mode with an empty `PublishConfiguration`, verifying it does not throw.
2. **Fix** — Moved `GetRegistryResource` calls and all Azure import logic after an early return for `isDryRun`, so dry-run mode only logs what would happen without requiring registry authentication details.

Fixes pipeline failures in #1963.

### DIFF
--- a/src/ImageBuilder/CopyImageService.cs
+++ b/src/ImageBuilder/CopyImageService.cs
@@ -53,12 +53,27 @@ public class CopyImageService : ICopyImageService
         ContainerRegistryImportSourceCredentials? sourceCredentials = null,
         bool isDryRun = false)
     {
+        Acr destAcr = Acr.Parse(destAcrName);
+
+        string action = isDryRun ? "(Dry run) Would have imported" : "Importing";
+        string sourceImageName = DockerHelper.GetImageName(srcRegistryName, srcTagName);
+        var destinationImageNames = destTagNames
+            .Select(tag => $"'{DockerHelper.GetImageName(destAcr.Name, tag)}'")
+            .ToList();
+        string formattedDestinationImages = string.Join(", ", destinationImageNames);
+        _logger.LogInformation("{Action} {DestinationImages} from '{SourceImage}'",
+            action, formattedDestinationImages, sourceImageName);
+
+        if (isDryRun)
+        {
+            _logger.LogInformation("Importing skipped due to dry run.");
+            return;
+        }
+
         var destResourceId = _publishConfig.GetRegistryResource(destAcrName);
         var srcResourceId = srcRegistryName is not null
             ? _publishConfig.GetRegistryResource(srcRegistryName)
             : null;
-
-        Acr destAcr = Acr.Parse(destAcrName);
 
         // Azure ACR import only supports one source identifier. Use ResourceId for ACR-to-ACR
         // imports (same tenant), or RegistryAddress for external registries.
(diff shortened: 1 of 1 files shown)


src/ImageBuilder/CopyImageService.cs
```diff
@@ -53,13 +53,28 @@ public async Task ImportImageAsync(
         ContainerRegistryImportSourceCredentials? sourceCredentials = null,
         bool isDryRun = false)
     {
+        Acr destAcr = Acr.Parse(destAcrName);
+
+        string action = isDryRun ? "(Dry run) Would have imported" : "Importing";
```
nagilson, line 58: nit: does this need localization?

src/ImageBuilder/CopyImageService.cs
```diff
@@ -53,13 +53,28 @@ public async Task ImportImageAsync(
...
+        string sourceImageName = DockerHelper.GetImageName(srcRegistryName, srcTagName);
+        var destinationImageNames = destTagNames
+            .Select(tag => $"'{DockerHelper.GetImageName(destAcr.Name, tag)}'")
+            .ToList();
+        string formattedDestinationImages = string.Join(", ", destinationImageNames);
+        _logger.LogInformation("{Action} {DestinationImages} from '{SourceImage}'",
```
nagilson, line 64: The other PR uses _loggerService but this uses _logger. What is the difference?

nagilson, line 64: `($"{action} {formattedDestinationImages} from '{sourceImageName}'");` <-- I think this approach would be cleaner

nagilson (approved):
Seems good, but left a few comments worth addressing. Thanks!

src/ImageBuilder/CopyImageService.cs
lbussell: We migrated to M.E.Logging: https://github.com/dotnet/docker-tools/pull/1968

lbussell: That is an anti-pattern for `Microsoft.Extensions.Logging`. If you format the string directly in the call to the logger:
1. You will not correctly capture structured logging information.
1. You will get an analyzer/warning for it.

src/ImageBuilder/CopyImageService.cs
lbussell: Negative, internal only tool.

src/ImageBuilder/CopyImageService.cs
nagilson: I see, thanks ;) Learned something new

PR #1991, opened 2026-03-02
Add signing pipeline implementation

### LINKED ISSUE #1376: Enable Image Signing
## Overview

This issue will cover the implementation details to enable https://github.com/dotnet/dotnet-docker/issues/4589

I have created a "vertical slice" of the signing infrastructure to verify that everything works. This is in the form of a pipeline that takes a pre-built image, signs it using our signing service, and verifies the signature. The pipeline code is [here](https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet-docker?path=eng/pipelines/dotnet-core-nightly.yml&version=GC3068026075529d734c854d044cb3b78d4a672490&_a=contents) [internal MSFT link].

### How images are signed

1. **Build the Images**
2. **Generate Signing Payloads**: Once images are built, generate a **signing payload** for each manifest and manifest list we intend to sign. The payload contains the OCI descriptor for the image to be signed. The descriptor can be retrieved by running the ORAS CLI.
3. **Sign Payloads**: Send all of the payloads off to the signing service at once. Our signing service will update the payloads in-place with the signed versions.
4. **Attach Signatures**: Next, the signed payload must be attached to the image as an artifact. This is done with the ORAS CLI.
    1. Using the signed payload, read the [COSE signature](https://github.com/notaryproject/specifications/blob/main/specs/signature-envelope-cose.md) to compute the x509 certificate chain. This is added to the signature artifact as an annotation. There is a reference implementation for this [here](https://github.com/Two-Hearts/generate-cert-chain-thumbprint) in python, that we'll need to port over to .NET.

### How images are verified

1. Download the root and/or issuer certificates corresponding to the signing key used for the images. All official and test certificates are accessible from public download links.
2. Add the certs to the Notation CLI in a new trust store.
3. Define and import a Notation Trust Policy that trusts all of the certs in the aforementioned trust store.
4. Then, you can verify the image with the notation CLI.

### Prerequisites:
- [x] https://github.com/dotnet/docker-tools/pull/1311
- [ ] Add Notation CLI to ImageBuilder Image

### Pipeline Changes

Signing must take place after all images are built. We'll need a list of all of the digests we built in order to know what to sign. Our usual image-info.json file is a natural fit. That means signing will need to wait until after the [Post-Build stage](https://github.com/dotnet/dotnet-docker/blob/41a2f2766ff71601aef369a7b7ddc2d9bacaaebc/eng/common/templates/stages/build-test-publish-repo.yml#L164-L172), which assembles the image info file describing all of the images we built. Furthermore, if we wish to sign manifest lists, we'll need to wait until after the [Publish Manifest stage](https://github.com/dotnet/dotnet-docker/blob/41a2f2766ff71601aef369a7b7ddc2d9bacaaebc/eng/common/templates/jobs/publish.yml#L74-L88) (not the most descriptive name), which creates and pushes all of the docker manifest lists for multi-platform tags. It also seems like we could consider moving the manifest list creation into the post-build stage as well.

The pipeline changes should allow signing with either test keys or production keys, and also the provide option to skip signing altogether.

### ImageBuilder Changes

Since we need to send signing payloads off to an external service via a pipeline task, the signing implementation in ImageBuilder must be split into at least two separate parts: before and after sending signing payloads to the signing service.

#### New Command: `GenerateSigningPayloads`

- Inputs: image-info.json file, output directory for signing payloads.
- Outputs: places one signing payload per image in the output directory. We may also find it useful to output an info file mapping each digest to its payload file's path, which we could pass to the next stage.

#### New Command: `AttachSignatures`

- Inputs: One of either (path to directory containing signing payloads) or (output file from `GenerateSigningPayloads` command)
- Outputs: List of signature digests.

This will be similar to `ApplyEolDigestAnnotations` from the current [image lifecycle annotation work](https://github.com/dotnet/docker-tools/issues/1201). There is opportunity to share parts of the implementation here.

In the case of failure, the command should output a list of digests and/or payloads that did not get their signature attached. We should also investigate what happens when there are multiple attempts to attach signatures to the same digest. There should not be two signatures referring to the same digest. One solution could be to remove any old signature annotations and re-attach newly created annotations on a re-run of the pipeline.

#### Verifying signatures

This is relatively straightforward - this could be done during our test leg, or immediately after attaching signatures. Notation CLI and ORAS only interact with registries and not the local Docker installation. This means verifying signatures does not require pulling any images, so this should be a relatively lightweight process. Performing this check outside of ordinary test infrastructure also means that this check could run on all repos without any test infrastructure changes.

### Does ACR import between ACRs leave artifacts and referrers intact?

We may consider [using ORAS to move images between ACRs](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-manage-artifact#promoting-the-artifact-graph) instead of `acr import`, to keep signatures intact.

### Effect on our ACR and clean-up schedule

In order to maintain integrity of the verified content, signatures exist in the registry just like an image, and simply refer to the digest of the image to be verified. Since signatures are stored in the registry just like other artifacts, they also may need to be cleaned up (needs more investigation) along with the images they refer to. In our cleanup pipelines, we may consider using ORAS to check for and delete associated artifacts when cleaning up old images in our ACR.

lbussell:
Fixes #1376

### DIFF
--- a/eng/docker-tools/templates/steps/init-signing-linux.yml
+++ b/eng/docker-tools/templates/steps/init-signing-linux.yml
@@ -81,16 +81,9 @@ steps:
     inputs:
       targetType: 'inline'
       script: |
-        $imageBuilderDockerRunExtraOptions = ""
-
-        if ($env:MBSIGN_APPFOLDER) {
-          # Mount the MicroBuild signing plugin directory (contains DDSignFiles.dll and esrpcli.dll).
-          $imageBuilderDockerRunExtraOptions = "-v $env:MBSIGN_APPFOLDER`:/microbuild"
-          Write-Host "MicroBuild signing enabled, mounting $env:MBSIGN_APPFOLDER to /microbuild"
-        } else {
-          Write-Warning "MBSIGN_APPFOLDER not set - signing will not work"
-        }
-
+        # Mount the MicroBuild signing plugin directory (contains DDSignFiles.dll and esrpcli.dll).
+        $imageBuilderDockerRunExtraOptions = "-v $env:MBSIGN_APPFOLDER`:/microbuild"
+        Write-Host "MicroBuild signing enabled, mounting $env:MBSIGN_APPFOLDER to /microbuild"
         Write-Host "##vso[task.setvariable variable=imageBuilderDockerRunExtraOptions]$imageBuilderDockerRunExtraOptions"
 
         # Write the signing env file for docker --env-file.


eng/docker-tools/templates/stages/publish.yml
```diff
@@ -34,12 +34,25 @@ stages:
     dependsOn: []
   ${{ else }}:
     ${{ if and(eq(variables['System.TeamProject'], parameters.internalProjectName), ne(variables['Build.Reason'], 'PullRequest')) }}:
-      dependsOn: Test
+      ${{ if eq(parameters.publishConfig.Signing.Enabled, true) }}:
+        dependsOn:
+        - Test
+        - Sign
+      ${{ else }}:
+        dependsOn: Test
```
mthalman: Is this possible instead? If so, you could use this pattern for Post_Build too.

```suggestion
      dependsOn:
      - Test
      - ${{ if eq(parameters.publishConfig.Signing.Enabled, true) }}:
        - Sign
```

eng/docker-tools/templates/steps/init-imagebuilder.yml
```diff
@@ -96,18 +97,65 @@ steps:
...
-          "$(imageNames.imageBuilder.withrepo)"
         )
 
         $authedDockerRunArgs = @(
-          '-e'
-          'SYSTEM_ACCESSTOKEN=$env:SYSTEM_ACCESSTOKEN'
-          '-e'
-          'SYSTEM_OIDCREQUESTURI=$env:SYSTEM_OIDCREQUESTURI'
+          "-e", 'SYSTEM_ACCESSTOKEN=$env:SYSTEM_ACCESSTOKEN'
+          "-e", 'SYSTEM_OIDCREQUESTURI=$env:SYSTEM_OIDCREQUESTURI'
```
mthalman: I think this can be simplified further. By just declaring the name, Docker will retrieve the value with that name. This applies to a lot of cases in this file.


```suggestion
          "-e", 'SYSTEM_ACCESSTOKEN'
          "-e", 'SYSTEM_OIDCREQUESTURI'
```

eng/docker-tools/templates/stages/publish.yml
lbussell: Probably! I'll take a look

eng/docker-tools/templates/steps/init-imagebuilder.yml
lbussell: I wonder if we could even use an env file. https://docs.docker.com/reference/cli/docker/container/run/#env

mthalman: I'm a fan of env files.

lbussell: Addressed in https://github.com/dotnet/docker-tools/pull/1991/commits/8d6fe2736ba2ee3599758bc2379b511f72a9305d and then https://github.com/dotnet/docker-tools/pull/1991/commits/f3226a0a45cab13179babad275738be6ea3b419e.

eng/docker-tools/templates/stages/publish.yml
lbussell: https://github.com/dotnet/docker-tools/pull/1991/commits/c48fab432ac0e0361f8edb690b1d1161bf374006

lbussell:
Validated internally: build#2917345

Issue #2006, opened 2026-03-06
Windows image build times are unnecessarily slow


lbussell:
On Windows, ImageBuilder is not ran inside a container. We use containers only as a means to shuttle around the ImageBuilder executable, extract it, and run natively on Windows.

### Analysis of a recent ImageBuilder build stage

~ | Win1809 | Win2022 | Combined
-|-:|-:|-:
Image pull      | `7m 25s`  | `31s`     | `7m 56s`
Image build     | `1m 28s`  | `1m 48s`  | `3m 16s`
Image push      | `4m 41s`  | `16s`     | `4m 58s`
Total           | `13m 35s` | `2m 35s`  | `16m 10s`

Only 20% of combined Windows ImageBuilder build time is the actual docker build. The other 80% is network I/O — pulls (49%) and pushes (31%) — dominated almost entirely by the Win1809 leg's `servercore:ltsc2016` image dependency.

lbussell:
[Triage] We spend a lot of build time shuttling around the servercore:ltsc2016 image in the Win1809 build leg (it builds both ltsc2016 and 1809 images in the same build). 1809 images and onwards use nano server, but server 2016 predates nano server.

We can never use docker-in-docker on server 2016, so even if using imagebuilder as a dotnet tool doesn't work out (#2007), we should find a way to avoid pulling the entire server 2016 image.

Unfortunately there is no such thing as a `FROM scratch` image without BuildKit on Windows. Maybe there's another, smaller alternative we could use.

PR #2007, opened 2026-03-06
Pack ImageBuilder as dotnet tool

lbussell:
This is related to https://github.com/dotnet/docker-tools/issues/2006. Since ImageBuilder is only used natively on Windows (and not ran from inside a container image), I would like to explore distributing ImageBuilder as a tool instead of a container image for that platform. The central hypothesis here is that we could potentially remove the Windows build leg of ImageBuilder entirely, speeding up build and iteration time.

This PR only adds a new acquisition channel for ImageBuilder. It does not remove Windows container images or anything like that.

### DIFF
--- a/eng/docker-tools/templates/variables/dotnet/common.yml
+++ b/eng/docker-tools/templates/variables/dotnet/common.yml
@@ -14,7 +14,7 @@ variables:
   value: --registry-creds 'docker.io=placeholder;placeholder'
 
 - name: linuxAmd64InternalPoolImage
-  value: 1es-ubuntu-2204
+  value: build.azurelinux.3.amd64
 - name: linuxAmd64PublicPoolImage
   value: build.azurelinux.3.amd64.open
 - name: linuxAmd64PublicPoolName


mthalman:
> The central hypothesis here is that we could potentially remove the Windows build leg of ImageBuilder entirely, drastically speeding up build and iteration time.

In January of 2027, Windows Server 2016 will be EOL. Dropping support for that would allow us to run ImageBuilder in a container just like we do for Linux because 2016 is the version that doesn't allow us to run nested containers. So that's only 9 months away. I'd rather keep the status quo until that point and then move forward with running ImageBuilder in a container in Windows. That will allow us to finally have a common pattern across both platforms.

lbussell:
> In January of 2027, Windows Server 2016 will be EOL. Dropping support for that would allow us to run ImageBuilder in a container just like we do for Linux because 2016 is the version that doesn't allow us to run nested containers. So that's only 9 months away. I'd rather keep the status quo until that point and then move forward with running ImageBuilder in a container in Windows. That will allow us to finally have a common pattern across both platforms.

Trust me, I too am counting the days until Server 2016 EOL.

However, the [latest ImageBuilder build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2919499&view=results) took 1 hour and 27 minutes. Whereas it takes *literally 40 seconds* to do a fully clean restore + build + test + pack of this repo on my _laptop_. It is not an exaggeration to say that we could ship changes 10x or even 20x faster by ditching Docker-in-Docker.

src/ImageBuilder/Microsoft.DotNet.ImageBuilder.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
+    <AssemblyName>imagebuilder</AssemblyName>
```
mthalman: This will impact other things:

https://github.com/dotnet/docker-tools/blob/71f34e00e2546b0b34c2ee6361513511be9e9fec/eng/docker-tools/Invoke-ImageBuilder.ps1#L79

https://github.com/dotnet/docker-tools/blob/71f34e00e2546b0b34c2ee6361513511be9e9fec/eng/docker-tools/templates/steps/init-imagebuilder.yml#L138

lbussell: I have a separate branch with those changes here: https://github.com/lbussell/docker-tools/commit/ab7ec6e886ed60c2f477d9d664a87b094299663d

These changes would break this PR since it is not using the updated version of ImageBuilder yet.

I think that we could also use `<ToolCommandName/>` here instead though.

lbussell: https://github.com/dotnet/docker-tools/pull/2007/commits/0a7dc1b9e0d7132628798efd98e75df3be50541c

Issue #2002, opened 2026-03-04
Move manifest list creation from publish stage to post-build stage


lbussell:
## Problem

Manifest lists are currently created and pushed during the **Publish** stage by the `publishManifest` command. This causes several problems:

1. **Manifest lists cannot be signed.** The Sign stage runs before Publish, so manifest list digests don't exist yet at signing time. See https://github.com/dotnet/docker-tools/issues/2001.
2. **The Publish stage creates new digests.** Ideally, the Publish stage should only copy and publish artifacts that were already built and verified - not create new ones.
3. **Signing manifest lists would require the Publish stage to sign.** If we tried to sign manifest lists without moving their creation earlier, we'd need signing infrastructure in the Publish stage, adding complexity and deviating from the clean separation of concerns between stages.

## Current Pipeline Flow

1. **Build** - creates platform images, produces per-job image-info fragments
2. **Post_Build** - merges image-info fragments into a single file
3. **Sign** - signs all digests in image-info (platform images only, no manifest lists yet)
4. **Test** - tests images
5. **Publish** - copies images, creates manifest lists, pushes manifests, records manifest digests

## Proposed Change

Move manifest list creation from the Publish stage to the Post_Build stage:

1. **Build** - creates platform images, produces per-job image-info fragments
2. **Post_Build** - merges image-info, **creates manifest lists**, records manifest digests in image-info
3. **Sign** - signs all digests in image-info (**platform images AND manifest lists**)
4. **Test** - tests images
5. **Publish** - copies images, pushes already-created manifest lists (**no new digests created**)

## Benefits

- **Enables manifest list signing** - manifest list digests are present in `image-info.json` before the Sign stage
- **Cleaner Publish stage** - Publish only copies and pushes existing artifacts, no new content is created
- **Simpler signing model** - all signing happens in one place (the Sign stage), no need for signing infrastructure in Publish

## Considerations

- The `publishManifest` command currently both creates manifest lists *and* pushes them to the registry. These two operations will need to be separated - creation in Post_Build, pushing in Publish.
- Manifest lists created in Post_Build will reference platform images in the build registry. During Publish, after images are copied to the publish registry, manifest lists may need to be re-tagged or re-associated with the publish registry.
- The `SaveTagInfoToImageInfoFileAsync` logic in `PublishManifestCommand` (which records manifest digests back to `image-info.json`) will need to move or be refactored to run at the appropriate stage.

PR #2030, opened 2026-03-17
Move manifest list creation to Post_Build, copy via ACR import

lbussell:
This PR makes the following changes:
- Add CreateManifestListCommand that is inteded to run in Post_Build after mergeImageInfo to create Docker manifest lists, push them to the staging registry, and record their digests in image-info.json. This enables manifest list signing by ensuring digests are present before the Sign stage.
- Extend CopyAcrImagesCommand to copy manifest list shared tags alongside platform tags via ACR import.
- Extract ManifestListHelper from PublishManifestCommand, essentially unchanged. It determines which manifest lists to create based on which platforms were actually built, preventing manifest lists from referencing unbuilt images.
- Remove PublishManifestCommand - no longer needed. The Publish stage will now only copy images/manifest lists.

This is part 1/2 of https://github.com/dotnet/docker-tools/issues/2002

The [move-manifest-list-pipelines](https://github.com/lbussell/docker-tools/tree/move-manifest-lists-pipelines) branch is companion to this one, and contains the pipeline changes necessary for these changes.

I validated these changes internally using the imagebuilder unofficial pipeline (build#2928726), and confirmed that manifest lists were signed (with test signatures) in the staging (test) ACR and that the signatures were copied across to the prod (test) ACR.


src/ImageBuilder/Commands/CopyAcrImagesCommand.cs
```diff
@@ -124,6 +132,58 @@ public override async Task ExecuteAsync()
...
+                                .FirstOrDefault(t => t.Name == sharedTag);
+
+                            if (tagInfo?.SyndicatedRepo is not null)
+                            {
+                                foreach (string syndicatedDestinationTagName in tagInfo.SyndicatedDestinationTags)
+                                {
+                                    destinationTag = TagInfo.GetFullyQualifiedName(
+                                        $"{Manifest.Registry}/{Options.RepoPrefix}{tagInfo.SyndicatedRepo}",
+                                        syndicatedDestinationTagName);
+                                    yield return (sourceTag, destinationTag);
```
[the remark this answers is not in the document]
lbussell: Fixed in e5b274aa

src/ImageBuilder/ManifestListHelper.cs
```diff
@@ -0,0 +1,166 @@
...
+        RepoInfo repo,
+        ImageInfo image,
+        ImageArtifactDetails imageArtifactDetails,
+        IEnumerable<string> tags,
+        Func<string, string> getImageName,
+        Func<PlatformInfo, TagInfo?> getTagRepresentative)
+    {
+        return tags
+            .Select(tag => BuildManifestListInfo(repo, image, imageArtifactDetails, tag, getImageName, getTagRepresentative))
+            .Where(manifestListInfo => manifestListInfo is not null)!;
```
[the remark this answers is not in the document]
lbussell: Fixed in 837b7256

src/ImageBuilder/Commands/CreateManifestListCommand.cs
```diff
@@ -0,0 +1,143 @@
...
+        foreach (ImageData image in images)
+        {
+            image.Manifest.Created = createdDate;
+
+            TagInfo sharedTag = image.ManifestImage.SharedTags.First();
+
+            image.Manifest.Digest = DockerHelper.GetDigestString(
+                image.ManifestRepo.FullModelName,
+                await _manifestService.Value.GetManifestDigestShaAsync(
+                    sharedTag.FullyQualifiedName, Options.IsDryRun));
```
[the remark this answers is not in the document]
lbussell: Fixed in 8a85be08

src/ImageBuilder.Tests/ManifestListHelperTests.cs
```diff
@@ -0,0 +1,547 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.ImageBuilder.Commands;
```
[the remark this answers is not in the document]
lbussell: Fixed in 9e1e84b6

src/ImageBuilder/Commands/CreateManifestListCommand.cs
```diff
@@ -0,0 +1,148 @@
...
+            registryName: Manifest.Registry);
+    }
+
+    private async Task SaveTagInfoToImageInfoFileAsync(DateTime createdDate, ImageArtifactDetails imageArtifactDetails)
+    {
+        _logger.LogInformation("SETTING TAG INFO");
+
+        IEnumerable<ImageData> images = imageArtifactDetails.Repos
+            .SelectMany(repo => repo.Images)
+            .Where(image => image.Manifest != null);
```
mthalman, line 94: This will end up including cached images as well. So if all the images that make up a manifest list were cached, then the manifest list won't exist at the staging location. And that's because we only push the manifest list tag if there's at least one non-cached image (see `ManifestListHelper.GetManifestListsForChangedImages`). So if we process such an image, it will process its shared tags below and attempt to get its digest from the staging location which won't exist. We either need to compensate for that here or push the manifest list tag to staging. I kind of like the latter because it's consistent with what we do for images and allows tests to rely on those tags existing, regardless of the cached state.

This also implies missing test coverage for this class for image caching scenarios.

src/ImageBuilder/ManifestListHelper.cs
```diff
@@ -0,0 +1,166 @@
...
+                                imagePlatform.Platform.Tags.Any()))
+                    .FirstOrDefault()
+                    .Platform;
+
+                if (platformInfo is null)
+                {
+                    throw new InvalidOperationException(
+                        $"Could not find a platform with concrete tags for '{platform.DockerfilePathRelativeToManifest}'.");
+                }
+
```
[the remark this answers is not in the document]
mthalman: +1

src/ImageBuilder/Commands/CreateManifestListCommand.cs
lbussell: > push the manifest list tag to staging

Are you suggesting only copying the fully-cached manifest list tags to staging, or all the tags?

I think that for either of those to work, we would need to skip signing for images/manifest lists that already have signatures, otherwise we might sign them twice.

Skipping signing images that already have signatures sounds like a good idea either way.

Does this align with what you were thinking?

mthalman: > Are you suggesting only copying the fully-cached manifest list tags to staging, or all the tags?

All concrete image tags are already being pushed to staging (regardless of cache status). So I'm saying we treat manifest list tags the same way.

> Skipping signing images that already have signatures sounds like a good idea either way.

Yes, this would imply the need to skip signing in this case.

src/ImageBuilder/Commands/CreateManifestListCommand.cs
```diff
@@ -0,0 +1,148 @@
...
+/// Designed to run in the Post_Build stage after image info fragments have been merged.
+/// </summary>
+public class CreateManifestListCommand : ManifestCommand<CreateManifestListOptions, CreateManifestListOptionsBuilder>
+{
+    private readonly Lazy<IManifestService> _manifestService;
+    private readonly IDockerService _dockerService;
+    private readonly ILogger<CreateManifestListCommand> _logger;
+    private readonly IDateTimeService _dateTimeService;
+    private readonly IRegistryCredentialsProvider _registryCredentialsProvider;
+    private readonly IAzureTokenCredentialProvider _tokenCredentialProvider;
```
mthalman: Unused field.

lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/2030/commits/6991eecf6ceab52089d0b81eb3f18603fe448bf8, https://github.com/dotnet/docker-tools/pull/2030/commits/65da6c9a43552de454e42e9094fcc45ad1fe0ec7

src/ImageBuilder/Commands/CreateManifestListCommand.cs
lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/2030/commits/8064469e993bb26d858c81948455ccdbd6057c9a and https://github.com/dotnet/docker-tools/pull/2030/commits/df35732ae18e9f243144aba9e92508926e8ecc68

PR #2032, opened 2026-03-18
Extract pool/signing variables into reusable template

gdams:
Move pool/image configuration and MicroBuild signing variables from dotnet/common.yml into a new pool-and-signing.yml template.

dotnet/common.yml currently bundles these shared variables together with the DotNet-Docker-Common-2 variable group, which forces downstream repos (e.g. go-images) to copy-paste ~80 lines of pool and signing config rather than importing dotnet/common.yml directly.

By extracting pool-and-signing.yml, downstream repos can import just the shared variables without pulling in dotnet-specific variable groups or re-importing base templates that would override their customizations.

See https://github.com/microsoft/go-images/blob/microsoft/main/eng/pipeline/variables/go-common.yml as an example of how much downstream duplication we need right now simply because we can't import the `DotNet-Docker-Common-2` variable group from `eng/docker-tools/templates/variables/dotnet/common.yml`

### DIFF
--- a/eng/docker-tools/templates/variables/dnceng-project-names.yml
+++ b/eng/docker-tools/templates/variables/dnceng-project-names.yml
@@ -1,5 +1,5 @@
 # Azure DevOps project name variables for dnceng pipelines.
-# Shared across .NET and Go repos that build in dnceng.
+# Reference this template in pipelines that run in the dnceng or dnceng-public Azure DevOps orgs.
 
 variables:
 - name: publicProjectName
--- a/eng/docker-tools/templates/variables/dnceng-signing.yml
+++ b/eng/docker-tools/templates/variables/dnceng-signing.yml
@@ -1,5 +1,5 @@
 # MicroBuild signing variables for dnceng pipelines.
-# Shared across .NET and Go repos that build in dnceng.
+# Reference this template in pipelines that run in the dnceng or dnceng-public Azure DevOps orgs.
 
 variables:
 - name: TeamName


eng/docker-tools/templates/variables/dotnet/common.yml
```diff
@@ -1,82 +1,5 @@
...
-# Team name required for MicroBuild signing plugin
-- name: TeamName
-  value: DotNetCore
-# MicroBuild signing plugin feed source
-- name: MicroBuildFeedSource
-  value: https://dnceng.pkgs.visualstudio.com/_packaging/MicroBuildToolset/nuget/v3/index.json
-# MicroBuild signing plugin version
-- name: MicroBuildPluginVersion
-  value: latest
+- template: /eng/docker-tools/templates/variables/pool-and-signing.yml@self
```
lbussell: Instead of putting this all into one big variable template, can we split it up? I think that would be a more meaningful improvement.

Additionally, all of these variables are **dnceng-specific**, which is something that the .NET and golang repos share. The new templates should be named as such:

```yaml
- template: /eng/docker-tools/templates/variables/dnceng-build-pools.yml@self
- template: /eng/docker-tools/templates/variables/dnceng-signing.yml@self
- template: /eng/docker-tools/templates/variables/dnceng-project-names.yml@self
```

For `dockerHubRegistryCreds`, not sure exactly where to put that. I can look at addressing it in the copy-base-images step.

gdams: agreed that it makes sense to split them up. Updated PTAL

dagood: FWIW, I'd be happy to leave `dockerHubRegistryCreds` in `eng/docker-tools/templates/variables/dotnet/common.yml` and keep duplicating it in the Go images repos. We have to stub out a handful of variables already, and this would fit in fine. (And unlike the pools/images, I don't imagine it will change much.)

lbussell: Yes, let's keep it in `dotnet/common.yml` for now since it's blocking the PR validation. I took a brief look at fixing this elsewhere in the pipeline templates and it wasn't trivial/should be outside the scope of this PR.

gdams: > Yes, let's keep it in `dotnet/common.yml` for now since it's blocking the PR validation. I took a brief look at fixing this elsewhere in the pipeline templates and it wasn't trivial/should be outside the scope of this PR.

added it back, let me know what you think :)

eng/docker-tools/templates/variables/dnceng-build-pools.yml
```diff
@@ -0,0 +1,58 @@
+# Build pool and image variables for dnceng pipelines.
+# Shared across .NET and Go repos that build in dnceng.
```
lbussell: **Nit**: this comment could be clearer.

```suggestion
# Reference this template in pipelines that run in the dnceng or dnceng-public Azure DevOps orgs.
```

Also applies to the other new templates.

PR #2119, opened 2026-05-23
Check base images using mirror registry instead of public DockerHub

lbussell:
This PR responds to new network isolation requirements that broke the check base images pipeline. Outbound connections to DockerHub are blocked by 1ES Network Isolation policies ("CFSClean") in internal Azure Pipelines runs.

Previously the pipeline worked like this:
1. **Copy Base Images** - imports new/updated external images to the internal mirror registry. Since this process runs in Azure, it isn't subject to the same Network Isolation policies as 1ES Pipelines.
2. **Get Stale Images** - compares base image digests from two sources:
    - From dotnet/versions / image-info.json files.
    - From the upstream/DockerHub registry.
3. **Queue a build** for all images that have mis-matched digests.

Obviously, comparing base image digests from DockerHub is a problem if the network blocks the connection.

This PR updates the `getStaleImages` command to use the mirrored images that we had already just copied anyways. For simplicity/consistency's sake, I also updated `getStaleImages` to use the exact same mirror/override arguments as `copyBaseImages`, that way we know we're referencing the images the exact same way.

### DIFF
--- a/src/ImageBuilder/Commands/GetStaleImagesCommand.cs
+++ b/src/ImageBuilder/Commands/GetStaleImagesCommand.cs
@@ -55,7 +55,11 @@ namespace Microsoft.DotNet.ImageBuilder.Commands
 
             IEnumerable<Task<SubscriptionImagePaths>> getPathResults =
                 SubscriptionHelper.GetSubscriptionManifests(
-                    Options.SubscriptionOptions.SubscriptionsPath, Options.FilterOptions, _gitService, _manifestJsonService)
+                    Options.SubscriptionOptions.SubscriptionsPath,
+                    Options.FilterOptions,
+                    _gitService,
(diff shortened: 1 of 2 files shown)


lbussell:
This PR contains ImageBuilder source code changes only. Pipeline changes are in https://github.com/dotnet/docker-tools/pull/2123.

mthalman:
> This process runs in Azure, and uses DockerHub credentials stored in Azure KeyVault, so it isn't subject to rate limiting or network isolation.

I don't understand why it's not subject to network isolation but other operations are.

lbussell:
> I don't understand why it's not subject to network isolation but other operations are.

I updated the description to hopefully be a little more clear.

mthalman (approved):

PR #2123, opened 2026-05-26
Check base images using mirror registry instead of public DockerHub (+ pipeline changes)

lbussell:
This PR is based on https://github.com/dotnet/docker-tools/pull/2119 and also includes the necessary pipeline changes that should be applied after the ImageBuilder reference is updated.


lbussell:
Superseded by https://github.com/dotnet/docker-tools/pull/2129.

PR #2126, opened 2026-05-26
Add platform validation for multi-arch manifests in Post_Build stage

### LINKED ISSUE #2127: Add pre-publish validation that ensures shared tag manifests include all expected platforms defined by manifest.json
Currently, an official build might publish a partial manifest for a few reasons:

* https://github.com/dotnet/docker-tools/issues/2107
* Historical: https://github.com/dotnet/docker-tools/pull/2125
  * Likely cause of outage: https://github.com/microsoft/go/issues/2313

I can't think of a reason an official build should ever publish a shared tag that doesn't include all its platforms defined in the `manifest.json`. As far as I can tell, this always indicates a problem. To protect against regressions, it makes sense to me to explicitly check rather than relying on the logic along the way. Failing the build is vastly preferable to pushing out bad tags.

A dev build, on the other hand, might want to publish a shared tag that only includes some platforms, for avoiding waste, improving end-to-end performance, or some specific test. So, I imagine it'd have to be conditional, or simply opt-in.

dagood:
Change `createManifestList` to validate the list of platforms in the manifest list against the expectations in `manifest.json`. More information in doc changes.

* Fixes https://github.com/dotnet/docker-tools/issues/2127


dagood:
This is an implementation by Copilot for:

* https://github.com/dotnet/docker-tools/issues/2127

The doc and test changes look reasonable to me, but I would need more time to understand everything happening in `src/ImageBuilder/ManifestListHelper.cs`. I can't tell at first glance if Copilot's self-criticism is valid, for example.

src/ImageBuilder/Commands/CreateManifestListOptions.cs
```diff
@@ -13,17 +13,24 @@ public class CreateManifestListOptions : ManifestOptions, IFilterableOptions
...
     public RegistryCredentialsOptions CredentialsOptions { get; set; } = new RegistryCredentialsOptions();
     public string ImageInfoPath { get; set; } = string.Empty;
+    public bool ValidateManifestListPlatforms { get; set; }
 
     private static readonly Argument<string> ImageInfoPathArgument = new(nameof(ImageInfoPath))
     {
         Description = "Path to the image info file to read and update with manifest list digests"
     };
 
+    private static readonly Option<bool> ValidateManifestListPlatformsOption = new("--validate-manifest-list-platforms")
```
mthalman: Why have an option at all? I would think this should just be the default, and only, behavior.

lbussell: +1, even in the "dev" scenario mentioned in #2127, you would want to simulate the real publishing behavior, I think. It will likely require #2128 to function without this option, however.

dagood: I imagined that someone might want to be able to run a filtered dev build that produces a manifest tag that only includes the outputs from that specific build, to tinker with.

With harvesting existing images to put in the manifest, maybe someone could end up being confused about which of the images pointed to by the manifest came from the dev build vs. some other build?

For me/Go, it doesn't matter either way, just trying to imagine odd dev cases and keep this change low-impact so it can be merged ASAP. (We're holding off on automated builds until validation is in.)

I'll push another commit that makes it the only behavior to see how that looks.

src/ImageBuilder/ManifestListHelper.cs
```diff
@@ -122,56 +228,101 @@ private static IEnumerable<ManifestListInfo> GetManifestListsForTags(
...
+                    candidateImage.AllPlatforms
                         .Select(candidatePlatform => (Image: candidateImage, Platform: candidatePlatform))
                         .Where(candidate =>
-                            // Exclude the current platform itself
                             platform != candidate.Platform
-                            // Must be the same Dockerfile, OS, and architecture
                             && PlatformInfo.AreMatchingPlatforms(
-                                image1: candidateImage,
+                                image1: image,
                                 platform1: platform,
```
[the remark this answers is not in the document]
lbussell: This bug is like super niche and we'll likely never hit it, but since the behavior depended on the ordering of the manfiest, I fixed it and added a regression test in https://github.com/dotnet/docker-tools/pull/2126/commits/480effebc1d4074e90341ea451c7838887b0b49c.

Honestly, I really don't like how the Manifest and ImageArtifactDetails are structured and accessed in ImageBuilder. The complexity of this file really exemplifies why. But that all predates this PR of course. I am just calling out that I mostly hope that this code is temporary and that I can greatly simplify this soon.

lbussell (approved):

dagood:
* Fixed logic-level conflict with https://github.com/dotnet/docker-tools/pull/2135.

PR #2135, opened 2026-06-04
Convert ImageBuilder tests from xUnit to MSTest

lbussell:
Converts `ImageBuilder.Tests` from xUnit to MSTest.

A few notes on decisions that aren't obvious from the diff:

- The test runner is selected via the `<TestRunnerName>MSTest</TestRunnerName>` MSBuild property, which tells Arcade to bring in the MSTest packages.
- Assertions were converted to Shouldly rather than `Assert.*`, to match the existing convention in this project.
- `TestParallelization.cs` sets class-level parallelism (`Workers = 0, Scope = ExecutionScope.ClassLevel`) to match xUnit's default behavior, so tests within a class still run serially.
- xUnit's `ITestOutputHelper` is replaced with MSTest's `TestContext.WriteLine`. This keeps the per-test diagnostic output correctly attributed under parallel execution, which `Console.WriteLine` would not guarantee.

Test bodies are otherwise left as-is.

### DIFF
--- a/src/ImageBuilder.Tests/IngestKustoImageInfoCommandTests.cs
+++ b/src/ImageBuilder.Tests/IngestKustoImageInfoCommandTests.cs
@@ -287,11 +287,11 @@ namespace Microsoft.DotNet.ImageBuilder.Tests
             command.LoadManifest();
             await command.ExecuteAsync();
 
-            TestContext?.WriteLine("{0}", $"Expected Image Data: {Environment.NewLine}{expectedImageData}");
-            TestContext?.WriteLine("{0}", $"Actual Image Data: {Environment.NewLine}{ingestedData[command.Options.ImageTable]}");
+            TestContext?.WriteLine($"Expected Image Data: {Environment.NewLine}{expectedImageData}");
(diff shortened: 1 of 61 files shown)


JeremyKuhne (approved):

src/ImageBuilder.Tests/IngestKustoImageInfoCommandTests.cs
```diff
@@ -292,20 +287,20 @@ private async Task ValidateExecuteAsync(
             command.LoadManifest();
             await command.ExecuteAsync();
 
-            _outputHelper.WriteLine($"Expected Image Data: {Environment.NewLine}{expectedImageData}");
-            _outputHelper.WriteLine($"Actual Image Data: {Environment.NewLine}{ingestedData[command.Options.ImageTable]}");
+            TestContext?.WriteLine("{0}", $"Expected Image Data: {Environment.NewLine}{expectedImageData}");
```
mthalman: Can you explain the need for `{0}` here?

lbussell: It was not necessary. I removed it.

mthalman (approved):

PR #2137, opened 2026-06-05
Rate limit HTTP requests to Azure Container Registry APIs

lbussell:
Ever since #2050, checking for image lifecycle annotations has sped up. Checking for existing EOL annotations can exceed rate ACR's per-identity rate limit.

The original LifecycleAnnotationService assumed that error while checking for annotations/referrers means that the annotation does not exist. If an HTTP 429 error occured while checking for lifecycle annotations, then it would report that the annotation doesn't exist. This caused issues in the EOL annotation step of the golang image publishing pipeline.

This PR adds:
- A regression test that ensures that the LifecycleAnnotationService doesn't swallow errors.
- Global ACR rate-limiting via standard HttpClient configuration.

For the rate limiting, I decided to migrate away from our ad-hoc `HttpClientProvider`. It existed since before we adopted `Microsoft.Extensions.Hosting` in #1860. `M.E.Hosting` contains an `IHttpClientProvider` by default. It already includes logging as well. Now, we configure `HttpClient` configuration and rate limiting through standard means.


src/ImageBuilder/LifecycleMetadataService.cs
mthalman, line 30: This was previously catching any exception and returning null in that case. Now there is a chance for it to throw so we need to re-evaluate the callers of this method to determine the impact. For example, there is this call path `AnnotateEolDigestsCommand.ExecuteAsync -> AnnotateEolDigestsCommand.AnnotateDigestAsync -> LifecycleMetadataService.IsDigestAnnotatedForEolAsync`. Because `AnnotateDigestAsync` gets invoked within a `Parallel.ForEachAsync`, any exception there can blow up the whole batch which defeats the purpose of the iterate nature of that workflow.

src/ImageBuilder/RateLimiting/AcrRateLimitingHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        if (!isAcrRequest)
+            return await base.SendAsync(request, cancellationToken);
+
+        using RateLimitLease lease = await rateLimiter.AcquireAsync(cancellationToken);
+
+        if (!lease.IsAcquired)
+            throw new InvalidOperationException("Unable to acquire a permit from the ACR rate limiter.");
+
+        return await base.SendAsync(request, cancellationToken);
```
mthalman, line 37: Consider adding logging for exceptions thrown from here. This was lost from the removal of `HttpClientProvider.cs`.

mthalman (commented):
Note that `ContainerRegistryClient` and `ContainerRegistryContentClient` are not getting the benefit of these changes to support rate limiting. Consider updating their respective factory classes to configure an options instance for the client.

```csharp
var options = new ContainerRegistryClientOptions
{
    Transport = new HttpClientTransport(httpClientFactory.CreateClient())
};
```

src/ImageBuilder/RateLimiting/AcrRateLimitingHandler.cs
lbussell: It is logged twice already. Once by the HttpClient's handler, and once by the (just added) resilience policy. Both include the stack trace.

src/ImageBuilder/LifecycleMetadataService.cs
lbussell: Added custom resilience handler in https://github.com/dotnet/docker-tools/pull/2137/commits/0a2c2b6ae465313b7787252e43b27107ccd52754 so as to not blow up over transient network errors.

src/ImageBuilder.Tests/AcrRateLimitingTests.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        public int RequestCount => _requestCount;
+
+        protected override Task<HttpResponseMessage> SendAsync(
+            HttpRequestMessage request,
+            CancellationToken cancellationToken
+        )
+        {
+            Interlocked.Increment(ref _requestCount);
+            return Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK));
```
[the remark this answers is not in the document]
lbussell: This suggestion doesn't compile. Also it looks like the code quality bot can't track ownership through `Task.FromResult`? This is just in tests, so even if I'm wrong and the response is not disposed, the impact is negligible. The response is disposed by the caller.

mthalman (approved):

PR #2143, opened 2026-06-12
Only rate limit ACR referrer lookup requests

### LINKED ISSUE #2141: Copying images during publishing is extremely slow
#2139 introduced bounded concurrency during the CopyAcrImagesCommand:

https://github.com/dotnet/docker-tools/blob/4a187e7d352125a67e938c4cc78a023530340be8/src/ImageBuilder/Commands/CopyAcrImagesCommand.cs#L62-L73

That PR was based on the premise that bounding concurrency while making ACR HTTP Requests (through ORAS) would result in a higher overall throughput than saturating the new ACR rate limiting (#2137).

This appeared to be true in my local testing that simply hammered the ACR as quickly as possible. However, a [recent pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2996512&view=logs&j=046815c1-7653-5dfd-0c09-c18aa9ed7f03&t=4a52a1c7-d280-57d4-c862-bdb046440835&s=5ac7342a-e4de-5b90-7317-19a97f311866) averaged about 26 requests per 60 seconds, roughly 10 times lower than the lowest rate limit.

This is because `ImportImageAsync`, which runs inside the same restricted parallelism, waits for each import to complete. In the previously linked pipeline, each import operation took anywhere from 0.2 to 10.0 seconds.

The actual pipeline took 72 minutes to copy all images. I have calculated some hypothetical numbers on what would happen if we ran at the rate limit with unbounded imports:

- **Total ACR requests:** 3,290 (1,638 manifest requests + 1,624 referrer requests + 28 oauth requests)
- **Permit rate:** 240 / 60s = **4 req/s**

`3290 requests / 4 rps = 822.5 seconds = 13.7 minutes`

## Additional note on ACR rate limits

From some _experimentation_, I learned that ACR's rate limits are different per-operation. Manifest requests and referrer lookups appear to have different rate limits. Referrer lookups have a much lower rate limits than other operations.

If we rate limit only the referrer lookups, then the theoretical floor drops to ~6.5 minutes.

Furthermore, it appears that different tiers of ACR have different rate limits. I have personally observed that Premium ACRs have a referrer rate limit of 1000/60s, while Basic ACRs have a limit of 250/60s.

## Proposed solution

The changes from #2137 and #2139 should be significantly dialed back. We should:

- Revert unnecessary changes to bounded parallelism until we notice problems happening in other commands.
- Only rate limit referrer lookups. Other operations can rely on Microsoft.Extensions.Http's default resilience pipeline behavior which retries on 429s and respects Retry-After headers. The likelihood of hitting the other rate limits is highly unlikely.
- Referrer rate limiting should default to 250/60s, but allow setting a higher limit via configuration.

lbussell:
This PR fixes #2141. The issue text contains the detailed explanation of why these changes are necessary, but summarizing here:

- The previous rate limiting changes, #2137 and #2139, were too restrictive and slowed down publishing too much - it was an order of magnitude slower than the theoretical lowest ACR rate limit due to unnecessarily tight parallelism limits.
- ACR referrer lookups have a much lower rate limit than the rest of the API.
- Therefore, this PR reverts parallelism limits and applies rate limiting to *only* the ACR referrer lookup through OrasDotnetService using a named HttpClient.
- Http resilience changes from the previous PRs remain in place.


mthalman (approved):

lbussell:
Current code has a problem where unbounded concurrency sends too many oauth requests to the registry up-front during image copying, causing more 429 errors. The default HTTP client has retries enabled, but oauth requests are POST requests, so they aren't retried. I will have a fix for this pushed shortly.

lbussell:
I tested this internally with a publish run of the dotnet-docker pipeline: build#3002401

Here are some metrics from that run:

| Metric | Value |
|---|---:|
| Copy Images task duration | 470.7 s (~7.85 min) |
| Active request window | 372.6 s |
| Tags to copy/import | 1,706 |

Requests/responses by endpoint:

| Endpoint / layer | Total | Success | 401 | 429 | Retries | Avg req/60s | Peak req/60s |
|---|---:|---:|---:|---:|---:|---:|---:|
| manifests (HEAD) | 3,412 | 1,706 | 1,706 | 0 | 0 | 551 | 3,412 |
| oauth2/token | 2,151 | 1,706 | 0 | 445 | 445 | 347 | 2,151 |
| referrers (GET) | 1,706 | 1,706 | 0 | 0 | 0 | 275 | 500 |
| ImportImage (ARM) | 1,807 | 1,706 | — | 101 | 101 | 285 | 500 |
| **Total** | **9,076** | **6,824** | **1,706** | **546** | **546** | — | — |

The high number of 401 responses is normal. ORAS only tries to get a token for the registry if it gets a 401 first.

429 responses for oauth requests still happen due to unbounded parallelism, but now they are correctly handled by the Azure SDK.

If we see further issues or failures, then we can revisit bounded parallelism with higher limits.

The high number of oauth token requests is not expected, so I'm investigating.

mthalman (approved):

Issue #2142, opened 2026-06-12
Publish image-info as OCI artifacts


lbussell:
Currently, [image-info.json](https://github.com/dotnet/versions/blob/main/build-info/docker/image-info.dotnet-dotnet-docker-main.json) is published by committing directly to the [dotnet/versions](https://github.com/dotnet/versions) repo. Branch protection or rulesets can prevent pushes to the repo, leaving automation dependent on stale image-info and causing unnecessary re-builds.

ImageBuilder should create the image-info artifact when the final image set is produced during the Post-Build stage of the build pipeline, then push/copy it to the registry as part of the same Publish flow as the images. A successful image publish should include both the images and matching image-info artifact.

For .NET itself, the OCI artifact should remain private until we decide on a public name and destination. Syncing image-info back to `dotnet/versions` should be handled separately by a scheduled pipeline that opens automated pull requests.

lbussell:
After discussing with the team, I think that these should be public from the get-go to maintain parity with the https://github.com/dotnet/versions repo.

lbussell:
Superseded by #2166

PR #2155, opened 2026-06-19
Publish image info as OCI artifact

lbussell:
This is part 1/2 of #2142. This contains the publishing side of things. Follow-up PR will be the consumption side.

This PR is focused on being as small as possible while adding the OCI artifact publishing.

Contents:
- New command, `publishImageInfoArtifact`, that can be used in the exact same place as the existing `publishImageInfo` command.
- New service, `ImageInfoService`, that handles publishing the image info artifact. It will also handle pulling the artifact later, in the consumption side of things.
- New field in `manifest.json` for specifying where the image info artifact will be published to.
- The artifact is of type `application/vnd.microsoft.dotnet.image-info.v1+json`.

Other notes:
- `RepoPrefix` is read directly from publishConfig into `publishImageInfoArtifact` without going through a CLI option (first command that does this, it should be the pattern going forwards).

### DIFF
--- a/src/ImageBuilder/Commands/PublishImageInfoArtifactCommand.cs
+++ b/src/ImageBuilder/Commands/PublishImageInfoArtifactCommand.cs
@@ -37,7 +37,7 @@ public class PublishImageInfoArtifactCommand(
         if (string.IsNullOrWhiteSpace(publishRegistry?.Server))
         {
             throw new InvalidOperationException(
-                "No publish registry is configured. Skipping image-info artifact push."
+                "No publish registry is configured. Cannot publish image-info artifact."
             );
         }
 
--- a/src/ImageBuilder/Commands/PublishImageInfoArtifactCommand.cs
+++ b/src/ImageBuilder/Commands/PublishImageInfoArtifactCommand.cs
@@ -1,0 +1,57 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.DotNet.ImageBuilder.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.DotNet.ImageBuilder.Commands;
+
+public class PublishImageInfoArtifactCommand(
+    IManifestJsonService manifestJsonService,
+    IImageInfoService imageInfoService,
+    IOptions<PublishConfiguration> publishConfigOptions,
+    ILogger<PublishImageInfoArtifactCommand> logger
+) : ManifestCommand<PublishImageInfoArtifactOptions>(manifestJsonService)
+{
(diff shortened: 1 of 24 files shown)


src/ImageBuilder/ImageInfoService.cs
```diff
@@ -0,0 +1,74 @@
...
+        if (isDryRun)
+        {
+            _logger.LogInformation("Skipping image info artifact push due to dry run.");
+            return;
+        }
+
+        string repo = $"{repoPrefix}{imageInfoArtifact.Repo}";
+        _logger.LogInformation(
+            "Image info will be published to registry={registry}, repo={Repo}, dryRun={DryRun}",
+            registry, repo, isDryRun);
```
mthalman: Move this to be above the dry-run exit above.

src/ImageBuilder/Commands/PublishImageInfoArtifactCommand.cs
```diff
@@ -0,0 +1,57 @@
...
+    public override async Task ExecuteAsync()
+    {
+        _logger.LogInformation("Publishing image info as OCI artifact");
+
+        RegistryEndpoint? publishRegistry = _publishConfiguration.PublishRegistry;
+
+        if (string.IsNullOrWhiteSpace(publishRegistry?.Server))
+        {
+            throw new InvalidOperationException(
+                "No publish registry is configured. Skipping image-info artifact push."
```
mthalman: It's not so much _skipping_ as it's _failing_.

src/ImageBuilder/Oras/IOrasService.cs
```diff
@@ -71,4 +71,25 @@ Task<string> AttachArtifactAsync(
         string artifactType,
         IDictionary<string, string> annotations,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Pushes text content to the registry as a single standalone OCI artifact and applies one or
```
mthalman: It's not specific to text, right? It seems any set of bytes can be provided.

```suggestion
    /// Pushes content to the registry as a single standalone OCI artifact and applies one or
```

src/ImageBuilder/ImageInfoService.cs
lbussell: Accepted. I moved the destination log before the dry-run return so dry-run logs still show where the artifact would be published.

src/ImageBuilder/Commands/PublishImageInfoArtifactCommand.cs
lbussell: Accepted. I updated the exception message to say the command cannot publish the image-info artifact because this is a failure, not a skip.

src/ImageBuilder/Oras/IOrasService.cs
lbussell: Accepted. The API accepts arbitrary bytes, so I changed the summary from text content to content.

mthalman (approved):

PR #2157, opened 2026-06-22
Consume OCI image info in `build`, `generateBuildMatrix`, `getStaleImages`, `mergeImageInfo` commands 

lbussell:
Part 2/3 of #2142. Part 3/3 will be documentation/pipeline changes.

Each commit is independently reviewable. I am taking a slightly different/refined approach compared to https://github.com/dotnet/docker-tools/pull/2155. Here are the invariants I followed:

- OCI image info is not optional.
- OCI image info is always the source of truth.
- Commands that previously relied on a file for image info input now pull the artifact directly using ORAS.
- During builds, image info is always pulled from the publish repo, not a public repo (unless doing a PR build, etc.). This means we don't have to rely on MAR for correctness. It also unlocks internal/private image info tracking for free.
- Individual image publishers decide whether they want to make OCI image info public or not by choosing the repo destination (one that's onboarded to MAR or not).
- Publishing to GitHub is downstream of the OCI image info and optional + disabled by default. Eventually we will disable/remove this feature since the automation would be extremely simple to write in a separate GitHub workflow or Azure pipeline something. It doesn't need to be in the publish pipeline.

Out of scope for this MVP:

- Publishing image infos for individual build legs. These stay as pipeline artifacts for now.
- Publishing un-merged image info to the build staging registry. Stays as pipeline artifact.
- Commands that take multiple image infos. Since individual image infos are pipeline artifacts these are still read from files.

Pipeline changes (part 3/3) are currently WIP, pushed here: https://github.com/dotnet/docker-tools/tree/lbussell/oci-image-info-pipelines

### DIFF
--- a/src/ImageBuilder/Oras/OrasDotNetService.cs
+++ b/src/ImageBuilder/Oras/OrasDotNetService.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
+using System.Text.Json;
 using System.Text.Json.Nodes;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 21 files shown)


src/ImageBuilder/Oras/OrasDotNetService.cs
```diff
@@ -283,6 +284,64 @@ await Packer.PackManifestAsync(
...
+        TimeSpan elapsed = Stopwatch.GetElapsedTime(startTime);
+        _logger.LogInformation(
+            "Pulled OCI artifact from {Registry}/{Repository}:{Tag}: manifestDigest={Digest}, artifactType={ArtifactType}, blobs={BlobCount} in {Elapsed}",
+            registry, repository, tag, manifestDescriptor.Digest, manifest.ArtifactType, blobs.Count, elapsed);
+
+        return new OciArtifact(manifestDescriptor, manifest, blobs);
+    }
+
+    /// <inheritdoc/>
+    public Task<Stream> FetchBlobAsync(
```
mthalman: This method is unused.

mthalman (approved):

src/ImageBuilder/Oras/OrasDotNetService.cs
lbussell: Removed in https://github.com/dotnet/docker-tools/pull/2157/commits/6e4ca7af1b75d0aad417588ef2651d16178eba77

lbussell:
Dismissing the github-code-quality comments about `Path.Combine` because the inputs are trusted and it only commented on test projects.

mthalman (approved):

Issue #1602, opened 2025-02-21
Use OCI image annotations to drive automated image re-builds


lbussell:
Related: https://github.com/dotnet/dotnet-docker/issues/5531

If we added the following [OCI image annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md), we could use this information to drive our automatic image rebuilds instead of keeping track of and updating image-info files.

- **`org.opencontainers.image.source`**: URL to get source code for building the image (string). This would be equivalent to image info's [commitUrl](https://github.com/dotnet/versions/blob/f3829a5549a0a62e295503b1c2c44f4af5c3bed4/build-info/docker/image-info.dotnet-dotnet-docker-main.json#L33)
- **`org.opencontainers.image.base.digest`**: [Digest](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests) that the image is based on (string). Equivalent to image-info's [baseImageDigest](https://github.com/dotnet/versions/blob/f3829a5549a0a62e295503b1c2c44f4af5c3bed4/build-info/docker/image-info.dotnet-dotnet-docker-main.json#L28).
- **`org.opencontainers.image.base.name`** (optional?): fully qualified base image name. I don't think this is completely necessary since we can query the Dockerfile for this information. However, we have this information, and adding it would make it easier for users to determine if our images are out of date without checking the Dockerfile.

To implement this, we'd need:

- [ ] Add the option to set annotations during image builds
  - This is done using the `--annotation "foo=bar"` argument for `docker build` [[source](https://docs.docker.com/build/metadata/annotations/#add-annotations)]
  - Could be turned on/off via the manifest or a build command option for ImageBuilder.
- [ ] Add the capability to use image annotations in [GetStaleImagesCommand](https://github.com/dotnet/docker-tools/blob/main/src/Microsoft.DotNet.ImageBuilder/src/Commands/GetStaleImagesCommand.cs)
  - This could be either automatic or configurable in [check-base-image-subscriptions.json](https://github.com/dotnet/docker-tools/blob/main/eng/check-base-image-subscriptions.json)

lbussell:
Since annotations are an OCI-only thing, this would require https://github.com/dotnet/dotnet-docker/issues/6148.

However, we can use image labels (on the image config) for the same purpose, even with the same keys/values. Even if we decided to move to OCI + annotations, keeping labels around with the same values is apparently common practice.

PR #2161, opened 2026-06-25
Source OCI image-info from PublishConfiguration and don't pull unless necessary

lbussell:
Previously, ImageBuilder pulled OCI image info using using CLI options. It also wasn't consistent across commands. This change resolves the artifact's location from `PublishConfiguration` (which is already injected via `appsettings.json`), so every command pulls from the registry where image-info was last published, falling back to the public registry when no publish registry is configured. The pulls are also gated to the scenarios that actually consume the image info, so commands don't need publish registry access when they don't use it.

These are ImageBuilder code-only changes; the corresponding pipeline template updates follow as a separate step per the two-step change process.

- Resolve build-time image-info source from `PublishConfiguration`.
- Resolve `getStaleImages` image-info source from `PublishConfiguration`, removing the override options.
- Resolve `generateBuildMatrix` image-info source from `PublishConfiguration`, removing the override options.
- Resolve `mergeImageInfo` initial image-info source from `PublishConfiguration`, removing the override options.
- Only pull image-info in `generateBuildMatrix` when `--trim-cached-images` is set.
- Only pull the initial image-info in `mergeImageInfo` when publishing.

### DIFF
--- a/src/ImageBuilder/Commands/MergeImageInfoCommand.cs
+++ b/src/ImageBuilder/Commands/MergeImageInfoCommand.cs
@@ -8,18 +8,25 @@ using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using Microsoft.DotNet.ImageBuilder.Configuration;
 using Microsoft.DotNet.ImageBuilder.Models.Image;
 using Microsoft.DotNet.ImageBuilder.ViewModel;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.DotNet.ImageBuilder.Commands
 {
     public partial class MergeImageInfoCommand : ManifestCommand<MergeImageInfoOptions>
     {
         private readonly IImageInfoService _imageInfoService;
+        private readonly PublishConfiguration _publishConfig;
 
-        public MergeImageInfoCommand(IManifestJsonService manifestJsonService, IImageInfoService imageInfoService) : base(manifestJsonService)
+        public MergeImageInfoCommand(
+            IManifestJsonService manifestJsonService,
+            IImageInfoService imageInfoService,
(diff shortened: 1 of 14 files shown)


src/ImageBuilder/Commands/MergeImageInfoCommand.cs
```diff
@@ -61,13 +68,19 @@ public override async Task ExecuteAsync()
...
                     .ImageArtifactDetails;
             }
-            else if (!string.IsNullOrWhiteSpace(Options.InitialImageInfoRegistryOverride) ||
-                !string.IsNullOrWhiteSpace(Options.InitialImageInfoRepoPrefix))
+            else if (Options.IsPublishScenario && !string.IsNullOrWhiteSpace(_publishConfig.PublishRegistry?.Server))
             {
+                // When publishing, the previously-published image info (the merge target) is pulled from the
+                // publish registry's OCI artifact. Fall back to the manifest registry when no publish registry
+                // is configured.
+                string imageInfoRegistry = string.IsNullOrWhiteSpace(_publishConfig.PublishRegistry.Server)
```
mthalman: This condition will always evaluate to false because the Server property has already been checked in this `if` block.

lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/2161/commits/97862b06877b0a9990201be744d39d536848e565

mthalman (approved):

lbussell:
In the scenario where OCI image info is not on a public registry, this makes public/PR validation builds and internal builds have a totally different workflow. One pulls from OCI and one needs to use a file. That really complicates the pipelines.

Before committing fully to this approach, I'm going to spend just a little bit of time seeing how much work it would be to implement https://github.com/dotnet/docker-tools/issues/1602 instead.

lbussell:
Closing this for now - See #2167 for context.

PR #2163, opened 2026-07-01
Add referrer with annotations for image source and base image name + digest

lbussell:
Part of #1602.

This PR attaches ImageBuilder build metadata to each newly built and pushed platform image as an OCI referrer artifact. The metadata is stored as annotations on the referrer manifest rather than as Docker image labels, so it does not get inherited by downstream images.

The referrer uses this artifact type:

`application/vnd.microsoft.imagebuilder.image-info.v1`

| Annotation | Example value |
|------------|---------------|
| `vnd.microsoft.imagebuilder.source` | `https://github.com/dotnet/docker-tools` |
| `vnd.microsoft.imagebuilder.revision` | `c0ffeec0ffeec0ffee...` |
| `vnd.microsoft.imagebuilder.base.name` | `mcr.microsoft.com/dotnet/runtime:9.0-azurelinux3.0` |
| `vnd.microsoft.imagebuilder.base.digest` | `sha256:abcdef123456...` |
| `vnd.microsoft.imagebuilder.dockerfile` | `src/Dockerfile.linux` |

The Dockerfile annotation contains the path to the Dockerfile that the image was built from, relative to the repo root. This maintains parity with what's tracked in ImageArtifactDetails/image-info.json without baking the metadata into the image config.

### Change to the build command's execution flow

While the git diff doesn't show it nicely, this PR rearranges the steps in the build command. Previously, there were multiple loops over all images. Now, each image is pushed immediately after it is built so ImageBuilder can attach the metadata referrer to the exact pushed image digest. Here is a high-level before/after comparison:

**Before**:
- Build:
  - Pull base images
  - For each image:
    - Build image
  - For each image:
    - Push image
  - For each image:
    - Compute image metadata (base image, layers, URL, etc.)
    - Populate image info
  - Write image-info.json

**After**:
- Build:
  - Pull base images
  - For each image:
    - Build image
    - Push image
    - Compute image metadata (base image, layers, URL, etc.)
    - Attach metadata referrer
    - Populate image info
  - Write image-info.json

**Note**: referrer annotations are not inherited by downstream images.


src/ImageBuilder/Commands/BuildCommand.cs
```diff
@@ -98,131 +85,282 @@ public override async Task ExecuteAsync()
...
+
+                            BuildImage(platform, allTags, labels);
+                            builtPlatforms.Add(platformData);
+
+                            if (Options.IsPushEnabled)
+                            {
+                                IEnumerable<TagInfo> tagsForDigest = imageArtifactDetails is null ? [] : concreteTags;
+
+                                // Log in again to refresh token as it may have expired from a long build
+                                await ExecuteWithDockerCredentialsAsync(
```
mthalman, line 211: What are the perf consequences of running this for every single image that gets built?

src/ImageBuilder/Commands/BuildCommand.cs
```diff
@@ -726,23 +770,40 @@ await Parallel.ForEachAsync(finalStageExternalFromImages, async (fromImage, canc
...
+                _dockerService.PushImage(tag.FullyQualifiedName, Options.IsDryRun);
 
-                foreach (TagInfo tag in _processedTags)
+                if (digestTagNames.Contains(tag.FullyQualifiedName))
                 {
-                    _dockerService.PushImage(tag.FullyQualifiedName, Options.IsDryRun);
+                    string? digest = null;
+                    for (int attempt = 0; attempt <= RetryHelper.MaxRetries && digest is null; attempt++)
+                    {
+                        digest = await imageDigestCache.GetLocalImageDigestAsync(tag.FullyQualifiedName, Options.IsDryRun);
```
mthalman, line 794: Include some sort of timing backoff for each attempt.

src/ImageBuilder/Commands/BuildCommand.cs
```diff
@@ -98,131 +85,282 @@ public override async Task ExecuteAsync()
...
+                            if (!string.IsNullOrEmpty(Options.SourceRepoUrl))
+                            {
+                                labels[OciAnnotations.Source] = Options.SourceRepoUrl;
+                                labels[OciAnnotations.Revision] = _gitService.GetCommitSha(platform.DockerfilePath, useFullHash: true);
+
+                                // Record which Dockerfile the image was built from, relative to the repo root.
+                                // The repo root must be discovered via Git rather than assuming it's the manifest's directory.
+                                string repoRoot = _gitService.GetRepoRoot(platform.DockerfilePath);
+                                labels[ImageBuilderLabels.Dockerfile] =
+                                    PathHelper.NormalizePath(Path.GetRelativePath(repoRoot, platform.DockerfilePath));
```
mthalman: Use `GitServiceExtensions.GetDockerfileCommitUrl` instead?

lbussell: This is a workaround for https://github.com/dotnet/docker-tools/issues/2165. Additionally, since the source URL is already present in the `org.opencontainers.image.source` label, including the repo base URL a second time in another label is not necessary. That's why I chose only the Dockerfile path here.

mthalman: If we already have `org.opencontainers.image.source` then why do we need this label?

src/ImageBuilder/Commands/BuildCommand.cs
lbussell: I'm going to take a second pass at this method. This is not how we should do retries. Thanks for calling it out.

src/ImageBuilder/Commands/BuildCommand.cs
lbussell: I'm testing it internally and I'll get back to you.

mthalman: This comment is still relevant.

src/ImageBuilder/Commands/BuildCommand.cs
mthalman: This comment is still relevant.

PR #2164, opened 2026-07-01
Continue on docker build errors

dkurepa:


dkurepa:
@mthalman @lbussell we're not sure if this file is consumed by some other repository, let us know if you think this change is acceptable

eng/docker-tools/templates/stages/build-and-test.yml
```diff
@@ -212,7 +212,7 @@ stages:
 ################################################################################
 - stage: Post_Build
   dependsOn: Build
-  condition: and(succeeded(), contains(variables['stages'], 'build'))
+  condition: and(succeededOrFailed(), contains(variables['stages'], 'build'))
```
mthalman: There would need to be some kind of switch here to enable this just for the buildtools repo as this is not the behavior we would want for any of the other consuming repos.

mthalman (changes requested):

eng/docker-tools/templates/stages/build-and-test.yml
dkurepa: added a parameter to the yaml templates to control this

mthalman (approved):

dkurepa:
@mthalman looks like I can't merge this PR, could you merge it for me please? Also it might be good if the product construction team had merge access in this repo in general

lbussell (approved):

PR #2173, opened 2026-07-13
Retry transient signature push failures

### LINKED ISSUE #2168: Sign stage times out uploading signatures when signing many images
The `Sign` stage intermittently fails when signing many images because an in-flight signature `PUT` is canceled by a 10s per-request timeout. It only passes when the whole stage is re-run.

Failing run: https://dev.azure.com/dnceng/internal/_build/results?buildId=3010858 (`Sign` stage)

```text
Polly.Timeout.TimeoutRejectedException: The operation didn't complete within the allowed timeout of '00:00:10'.   Execution Time: 10005ms
HTTP PUT .../dotnet/aspnet/manifests/sha256:abc123...
  at OrasDotNetService.PushSignatureAsync ... line 132
  at ImageSigningService.SignImagesAsync ... line 110
```

The uploads themselves during signing are small. Payload size is not the cause. The failure is a transient/slow registry response under load: the `image-builder-oras-timeout` handler in `src/ImageBuilder/ImageBuilder.cs` applies `AddTimeout(TimeSpan.FromSeconds(10))`, a total per-attempt timeout, and `ImageSigningService.SignImagesAsync` pushes all signatures concurrently via `Parallel.ForEachAsync`. When one `PUT` exceeds 10s, it is canceled.

The Polly retry does not recover it because `retryOptions.DisableForUnsafeHttpMethods()` disables retries for `PUT`/`POST`, so the request is logged as `Handled: 'False'` and the command fails outright, requiring manual re-run.

We should make signature uploads resilient to transient registry slowness so the stage does not depend on a re-run. Possible directions worth investigating (not prescriptive):

- bound parallelism during signing
- enabling retries for idempotent digest-addressed writes
- inactivity-based request timeout (allow unlimited timeout as long as bytes are actively being uploaded)

lbussell:
This PR fixes #2168 by adding a targeted retry handler to the push operation for signatures. 

Related to #2137, #2139, and #2143.

### DIFF
--- a/src/ImageBuilder/Oras/OrasDotNetService.cs
+++ b/src/ImageBuilder/Oras/OrasDotNetService.cs
@@ -37,8 +37,6 @@ public class OrasDotNetService(
     IRegistryCredentialsHost? credentialsHost = null)
         : IOrasService
 {
-    private const int PushSignatureMaxRetryAttempts = 2;
-
     /// <summary>
     /// Media type for COSE signature envelopes per the Notary v2 spec.
     /// </summary>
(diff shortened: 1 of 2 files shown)


src/ImageBuilder/Oras/OrasDotNetService.cs
```diff
@@ -245,4 +266,35 @@ private Repository CreateRepository(string reference)
...
+                            PushSignatureMaxRetryAttempts);
+                        return default;
+                    },
+                }
+            )
+            .Build();
+
+    private static bool IsTransientHttpException(HttpRequestException exception) =>
+        exception.StatusCode is null
+        || exception.StatusCode is HttpStatusCode.RequestTimeout or HttpStatusCode.TooManyRequests
```
mthalman: On 429/TooManyRequests status, the logic is ignoring `Retry-After` from the response and using its own backoff time. Ideally it would honor `Retry-After`.

mthalman (approved):

src/ImageBuilder/Oras/OrasDotNetService.cs
lbussell: I updated the PR to treat the immediate symptom only. There is potential for simplification around how HTTP resilience is handled in ImageBuilder, but this PR is intended to be a band-aid fix to unblock publishing large amounts of images.

PR #2179, opened 2026-07-17
Preserve file modes for bundled infrastructure

lbussell:
PR #2159 embedded infrastructure file contents and paths in ImageBuilder, but .NET embedded resources do not carry Unix file modes. As a result, the `update` command recreated executable helper scripts as `0644`, producing the mode-only changes in #2177.

Generate `ContentFileModes.txt` from Git's index and embed it alongside the infrastructure resources. ImageBuilder applies the captured mode for every file it writes on Unix, without maintaining a runtime path allowlist. A build check rejects stale generated metadata when building from a Git checkout.

Validated by building the Linux ImageBuilder container and running `update` inside it: the four affected scripts are recorded by Git as `100755`, while regular helper scripts remain `100644`.


lbussell:
Back to the drawing board for a bit...

PR #2188, opened 2026-07-24
Add GitHub App authentication to GitAutomation library using Azure Key Vault

lbussell:
Related: https://github.com/dotnet/dotnet-docker-internal/issues/7468

GitHub App authentication goes like this:
- Sign a JWT with your app's private key and send it to GitHub in order to get an app access token.
- Exchange the app's access token for an access token for individual installations. That's the only thing the app's access token is good for.
- Use installation tokens for actual operations.

This PR adds App authentication support to the GitAutomation library:
- The GitHub App client and Installation clients are managed by two separate `ICredentialStore` implementations. `ICredentialStore` is from OctoKit. By implementing our own credential stores, we can have automatic refresh of both the app tokens and multiple installation tokens. The installation clients use the single app client internally to get and refresh their installation tokens.
- When using GitHub App authentication, the automation identity is resolved automatically. All you need is the client ID and the private key. The "TokenProvider" abstraction is expanded to a general "AccessProvider" to allow for this.
- The App's private key is signed remotely using Azure Key Vault. It never leaves the Key Vault. There are no plans to add support for local private keys.


src/GitAutomation/GitHub/GitHubAppCredentialStore.cs
```diff
@@ -0,0 +1,74 @@
...
+        try
+        {
+            if (_credentials is not null && _expiresAt > DateTimeOffset.UtcNow.AddMinutes(1))
+            {
+                return _credentials;
+            }
+
+            DateTimeOffset now = DateTimeOffset.UtcNow;
+            // GitHub's maximum expiration time is 10 minutes, stay a little under that.
+            _expiresAt = now.AddMinutes(9);
```
mthalman: This should only be set at the field-level after the signing of the JWT has succeeded. If there happens to be any kind of failure, the expiration will already have been advanced causing repeated 401 statuses on further authenticated requests during that period.

lbussell: Accepted. The fix now keeps expiresAt local while signing and only updates _expiresAt after SignDataAsync succeeds, so a failed signing attempt will not extend the cached credential lifetime.

mthalman (approved):

Issue #2056, opened 2026-03-30
`cleanup-acr-images-official` pipeline is broken


lbussell:
The `cleanup-acr-images-official` pipeline has been broken for some time. This issue should be closed only once the pipeline runs successfully.

Current failure modes:

1. Ghost manifest crashes `GenerateEolAnnotationDataForAllImagesCommand` - https://github.com/dotnet/docker-tools/pull/2055.
~~`GetAllManifestPropertiesAsync` lists a manifest in `build-staging/2881074/dotnet/nightly/aspnet` that no longer exists - `GetManifestAsync` returns 404 every time. Because the exception is unhandled, it crashes the entire Annotations job. This has been blocking the cleanup pipeline from completing at all.~~
2. `oras` CLI is not authenticated in the Clean job - https://github.com/dotnet/docker-tools/pull/2052.
~~The `pruneEol` action calls `IsDigestAnnotatedForEol`, which shells out to `oras discover`. The oras CLI has no credentials in the Clean job - every call returns 401 Unauthorized.~~
3. Duplicate annotations keep staging repos perpetually fresh.
https://github.com/dotnet/docker-tools/issues/2045 caused ImageBuilder to think that no image digests had EOL lifecycle annotations. This causes ImageBuilder to attach lots of duplicate annotations (See https://github.com/dotnet/dotnet-docker/issues/7121). Furthermore, staging image repos are pruned based on their `LastUpdatedOn` date. Months of duplicate annotation writes have updated `LastUpdatedOn` on staging repos, so the `Delete build-staging/*` step's 15-day age check never passes. 1,840 staging repos have accumulated. This also inflates the publish pipeline's `generateEolAnnotationDataForPublish` to 78 minutes and 5M log lines, as it enumerates and checks all the accumulated digests and their bloated referrer lists.
   - https://github.com/dotnet/docker-tools/pull/2057 adds logging to confirm whether using the created date instead of the `LastUpdatedOn` date is the correct solution for this.
4. `Clean public/dotnet* (PruneEol > 15d)` step times out. - https://github.com/dotnet/docker-tools/pull/2050.
[...]

PR #2195, opened 2026-07-30
Clean up manifests in batches

lbussell:
**Problem**: ACR cleanup currently happens in one shot. All manifests are queried from each repo all at once, and then they are deleted all at once. If there are too many manifests in a repo, then the cleanup pipeline can time out while listing the manifests before even deleting anything. As a result, manifests keep piling up and nothing gets deleted. This problem is exacerbated by recent ACR rate limiting changes.

This PR makes the following changes:
- Manifests are queried and processed / deleted in batches of 250. This means that even if cleanup doesn't finish, progress will be made incrementally over multiple runs.
- You can specify an overall time limit for the operation, after which the cleanup will stop gracefully, even if all repos/manifests were not processed.
- Repos are processed sequentially rather than running them all in parallel. This is just to keep things simple / easier to understand after this refactor.

One could imagine running cleanup with less intensity and more frequency after these changes (perhaps running it for 30 minutes every hour?). We'll see after these changes.

This PR is part of #2056

### DIFF
--- a/src/ImageBuilder/Commands/CleanAcrImagesOptions.cs
+++ b/src/ImageBuilder/Commands/CleanAcrImagesOptions.cs
@@ -17,6 +17,7 @@ public class CleanAcrImagesOptions : Options
     public int Age { get; set; }
     public string RegistryName { get; set; }
     public string[] ImagesToExclude { get; set; } = [];
+    public ushort? TimeLimitMinutes { get; set; }
 
     private const CleanAcrImagesAction DefaultCleanAcrImagesAction = CleanAcrImagesAction.PruneDangling;
     private const int DefaultAge = 30;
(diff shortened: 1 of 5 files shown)


src/ImageBuilder/Commands/CleanAcrImagesOptions.cs
```diff
@@ -52,32 +51,36 @@ public class CleanAcrImagesOptions : Options
         AllowMultipleArgumentsPerToken = false
     };
 
+    private static readonly Option<ushort?> TimeLimitMinutesOption = new("--time-limit-minutes")
+    {
+        Description = "Maximum cleanup duration before stopping after the current manifest batch (default: unlimited)"
```
mthalman: This doesn't accurately describe the implementation. The implementation will can produce a timeout while fetching manifests. So an in-progress batch may never get to the deletion phase, which is contrary to this description.

lbussell: Updated description in https://github.com/dotnet/docker-tools/pull/2195/commits/8301d90418ef434ef34eff3bff0a20ae07238005

mthalman (approved):

PR #2196, opened 2026-07-31
Simplify automated rebuild failure gating

mthalman:
Automated rebuilds currently only count pipeline runs tagged `autobuilder`, which makes recovery depend on manually tagging a successful run. This change bases the safeguard solely on pipeline history.

The rebuild workflow now blocks only when the three most recent runs all failed. Any successful run breaks the failure streak, and the recovery guidance and focused tests reflect the new behavior.


lbussell (approved):
Thanks, this greatly simplifies things and should cut down on failure notification noise.

Issue #2166, opened 2026-07-01
Rework container image rebuilds


lbussell:
## Current state

```mermaid
graph TD

subgraph docker-tools repo
check-base-images["check base images pipeline \n (multiple repos)"]
subscriptions[subscriptions.json]
end

subgraph consuming["consuming repo"]
build-pipeline[build-pipeline]
end

subgraph versions["versions repo"]
image-info[image-info.json]
end

subgraph acr
public["public images"]
end

check-base-images -->|reads| subscriptions
check-base-images -->|"reads (many)"| image-info
check-base-images -->|"queues (many)"| build-pipeline
subscriptions -->|"references (many)"| consuming
build-pipeline <-->|updates and reads| image-info
build-pipeline -->|pushes images| public
```

### Problems with current setup

- Requires a separate repo to store `image-info.json` metadata file. Git is in the critical path for builds.
- Only the docker-tools repo can queue base image re-bulids, and the docker-tools repo must know about all other consuming repos in order to queue their builds.
- The docker-tools repo and the consuming repos can go out-of-sync. For example, if the manifest.json or image-info formats change, then the docker-tools check-base-images pipeline can fail until all consuming repos have adopted the updated image-builder version.

## Desired state

**Key ideas**:
- Consuming repos should own their own base image update pipeline + schedule.
- The docker-tools repo should provide sane defaults/pipeline templates for base image re-builds.
- Base image re-builds should work both in public and internal scenarios without extra work.
- There should not be a dependency on external repos or storage locations for the critical base image re-build path. `image-info.json`, if it still needs to exist, should not be a source of truth. It should be computed separately, strictly downstream of image build pipelines.

```mermaid
graph TD

subgraph consuming[consuming repo]
check-base-images["check base images pipeline \n (one repo)"]
build-pipeline[build pipeline]
end

subgraph acr
public["public images \n with labels/annotations"]
end

public -->|reads| check-base-images
public -->|reads| build-pipeline
check-base-images -->|queues| build-pipeline
build-pipeline -->|pushes images| public
```

## Future work

Here's the work that I see required to reach this desired state:

- [ ] #1602
  - [ ] Attach metadata labels to images during build - #2163
- [ ] Make rebuild decisions repo-local
  - [ ] Implement a new version of the `getStaleImages` command which reads these labels to determine which images are out of date.
  - [ ] Implement a new pipeline which runs the new command and queues a build for stale images.
- [ ] Backwards compatibility with existing image-info.json publishing
[...]

PR #2207, opened 2026-08-05
[feature/caching-rework] Centralize build planning

lbussell:
This PR centralizes build planning and policies across multiple separate commands. The design is not perfect as-is but I would like to work on improving it incrementally.

Part of https://github.com/dotnet/docker-tools/issues/2166

## Design

The design centers around three new types: `BuildGraph`, `BuildPlanner`, and `IBuildPolicy`.

A `BuildGraph` represents the relationships between the images in the manifest.
The `BuildPlanner` takes the `BuildGraph` in combination with an `IBuildPolicy` and uses it to determine what actions need to be taken.

```mermaid
classDiagram
direction LR

class BuildCommand {
}
class GenerateBuildMatrixCommand {
}
class GetStaleImagesCommand {
}
class BuildTarget {
}
class BuildGraph {
  +Targets
  +Parents
  +Children
  +SharedBuildTargets
}
class BuildPlanner {
  +CreatePlanAsync(graph, imageInfo, policy)
}
class IBuildPolicy {
}
class BuildPlanItem {
  +Target
  +Action
  +Reasons
  +PublishedImage
}
class BuildAction {
  NoAction
  UsePublishedImage
  PublishExistingImage
  BuildImage
}
class BuildReason {
  +Message
  +Cause
}
class PublishedImage {
  +Source
  +Image
  +SharedTags
}

BuildCommand --> BuildGraph : creates
GenerateBuildMatrixCommand --> BuildGraph : creates
GetStaleImagesCommand --> BuildGraph : creates

BuildCommand --> BuildPlanner : executes build plan
GenerateBuildMatrixCommand --> BuildPlanner : uses for trimming
GetStaleImagesCommand --> BuildPlanner : checks for stale images

BuildGraph *-- BuildTarget
BuildPlanner --> BuildGraph
BuildPlanner --> IBuildPolicy : evaluates
BuildPlanner --> BuildPlanItem : produces

BuildPlanItem --> BuildTarget
BuildPlanItem --> BuildAction
BuildPlanItem *-- BuildReason
BuildPlanItem --> PublishedImage
PublishedImage --> BuildTarget : metadata source
```

### Build policies

There are multiple "policies" that we check images against to determine what actions to take. Previously, all these checks were strewn about the codebase and not managed centrally anywhere.

The `CompositeBuildPolicy` combines multiple different build policies into one that we can pass to the build planner.

```mermaid
classDiagram
direction LR

class IBuildPolicy {
  +EvaluateAsync(context) BuildPolicyResult
}
class BuildPolicyContext {
  +BuildGraph Graph
  +BuildTarget Target
  +PublishedImages
}
class BuildPolicyResult {
  +BuildAction Action
  +BuildReason[] Reasons
}
class CompositeBuildPolicy {
}
class AlwaysBuildPolicy {
}
class MissingPublishedImagePolicy {
}
class TagSetChangedPolicy {
}
class DockerfileChangedPolicy {
}
class BaseImageChangedPolicy {
}
class ImageDigestCache {
}
class IGitService {
}

IBuildPolicy <|.. CompositeBuildPolicy
IBuildPolicy <|.. AlwaysBuildPolicy
IBuildPolicy <|.. MissingPublishedImagePolicy
IBuildPolicy <|.. TagSetChangedPolicy
IBuildPolicy <|.. DockerfileChangedPolicy
IBuildPolicy <|.. BaseImageChangedPolicy

IBuildPolicy --> BuildPolicyContext : input
IBuildPolicy --> BuildPolicyResult : output
CompositeBuildPolicy o-- IBuildPolicy : combines

BuildPolicyResult --> BuildAction
BuildPolicyResult *-- BuildReason
BuildPolicyContext --> BuildGraph
BuildPolicyContext --> BuildTarget
BuildPolicyContext --> PublishedImage

BaseImageChangedPolicy --> ImageDigestCache
DockerfileChangedPolicy --> IGitService
```


src/ImageBuilder/Commands/GetStaleImagesCommand.cs
```diff
@@ -97,110 +100,38 @@ private async Task<IEnumerable<string>> GetPathsToRebuildAsync(Models.Subscripti
...
-            if (shouldRebuildImage)
-            {
-                IEnumerable<PlatformInfo> dependentPlatforms = GetDescendants(platform, manifest);
-                return dependentPlatforms.Select(p => p.Model.Dockerfile).ToList();
-            }
-
-            return [];
+            BuildGraph graph = BuildGraph.CreateFiltered(manifest);
+
+            IBuildPolicy policy = new CompositeBuildPolicy(
```
mthalman, line 105: Are we able to have some kind of common logic for constructing the composite policy so that it's shared amongst the commands? I know there are differences but the overall structure is similar.

src/ImageBuilder/Build/BuildPlan.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using System.Collections.Generic;
+using Microsoft.DotNet.ImageBuilder.Models.Image;
+
+namespace Microsoft.DotNet.ImageBuilder.Build;
+
+/// <summary>
+/// Work that ImageBuilder must perform for a target.
+/// </summary>
+public enum BuildAction
```
mthalman, line 13: Having the enum value represent its priority is a hidden contract that can be avoided by just have an explicit Action/Priority pair instead.

src/ImageBuilder/Commands/GetStaleImagesCommand.cs
```diff
@@ -97,110 +100,38 @@ private async Task<IEnumerable<string>> GetPathsToRebuildAsync(Models.Subscripti
...
+                    new MissingPublishedImagePolicy(),
+
+                    // Rebuild when the registry digest for the base image has changed.
+                    BaseImageChangedPolicy.FromRegistry(
+                        _imageDigestCache,
+                        imageNameResolver,
+                        Options.IsDryRun),
+
+                    // Rebuild when the Dockerfile has changed.
+                    new DockerfileChangedPolicy(
```
mthalman, line 120: `GetStaleImages` had never checked Dockerfile commits before. This policy will cause a failure when attempting to get a Dockerfile commit because the repo clone that had been done in `SubscriptionHelper` deletes the clone after loading the manifest, causing the Dockerfile to no longer exist on disk.

src/ImageBuilder/Build/BuildPlanner.cs
```diff
@@ -0,0 +1,343 @@
...
+        {
+            _logger.LogInformation(
+                "Build plan for {DockerfilePath}: {Action}. {Reasons}",
+                GetName(item.Target),
+                item.Action,
+                string.Join(" ", item.Reasons.Select(FormatReason)));
+        }
+    }
+
+    private static BuildReason GetCause(BuildPlanItem item) => item.Reasons.Last();
```
mthalman, line 334: Just selecting last can be misleading, right? It would seem better to determine the winning action and use its reason.

src/ImageBuilder/Commands/GetStaleImagesCommand.cs
lbussell: Good catch, for now the best approach is probably to only check for base image updates in `GetStaleImages`.

src/ImageBuilder/Build/BuildPlanner.cs
lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/2207/commits/42a376d4308209916426294084f99e79f52025e2 by only capturing the winning decision, and logging all others.

src/ImageBuilder/Build/BuildPlan.cs
lbussell: Added explicit priority mapping in https://github.com/dotnet/docker-tools/pull/2207/commits/31328ecb012edf20d49136e50ad413158dba0367.

src/ImageBuilder/Commands/GetStaleImagesCommand.cs
lbussell: Added in https://github.com/dotnet/docker-tools/pull/2207/commits/29742fb5142b44bcb855e1331cb155745373f169

src/ImageBuilder/Build/BuildPlanner.cs
```diff
@@ -0,0 +1,279 @@
...
+                    // Only select platforms that are part of the current build graph.
+                    && targetsByPlatform.ContainsKey(item.Platform.PlatformInfo)
+                )
+                .ToDictionary(
+                    // SAFETY: PlatformInfo is checked in the Where clause above.
+                    item => targetsByPlatform[item.Platform.PlatformInfo!],
+                    item => new PublishedImage(
+                        targetsByPlatform[item.Platform.PlatformInfo!],
+                        item.Platform,
+                        item.SharedTags)
```
mthalman, line 86: This would break if there are multiple platforms across multiple images that point to the same Dockerfile. This would only happen in a scenario like we once had when we support multi-platform manifest lists across Windows and Linux. In that scenario, we had one manifest image for that multi-platform and another for the multi-arch Debian tag, both tags pointing to the same Debian Dockerfile. That would cause a duplicate key exception to occur here. We don't currently have such a configuration but wanted to point it out.

src/ImageBuilder/Build/BuildPlanner.cs
```diff
@@ -0,0 +1,279 @@
...
+
+    private void LogPlan(IEnumerable<BuildPlanItem> plan)
+    {
+        foreach (BuildPlanItem item in plan)
+        {
+            logger.LogInformation(
+                "Build plan for {BuildTarget}: {Action}. {Reason}",
+                item.Target.DisplayName,
+                item.Decision.Action,
+                item.Decision.Reason);
```
mthalman, line 276: Because `BuildReason` is recursively nested and we're relying on the logger's default implementation of calling `ToString`, this results in bunch of nested output, not flattened.

Example:

`
Build plan for runtime (...): BuildImage. BuildReason { Message = "Dependency 'runtime-deps (...)' must build., CausedBy = BuildReason { Message = Base image 'mcr.microsoft.com/dotnet/runtime:8.0' changed from 'sha256:old' to 'sha256:new'., CausedBy =  } }
`

src/ImageBuilder/Build/BuildPlanner.cs
lbussell: A previous implementation did check for this, but then I evaluated all of the consuming repos and did not find an instance of that pattern, so I removed it (in [this diff](https://github.com/dotnet/docker-tools/commit/5d31ab9ccca72a9f729be10389bdad8c7a3d482f#diff-dfecdebc7786642ad95ddf1f148212f922987b43b93e4065473f65ac081569a8R74-R87)).

src/ImageBuilder/Build/BuildPlanner.cs
lbussell: This is intentional in order to keep the code simpler for now. If ends up looking really bad, I'll change it later.

mthalman (approved):

PR #2208, opened 2026-08-07
Simplify publishing output paths

lbussell:
The pipeline artifacts from the publish stage aren't consumed anywhere (that I could find). Yet, we upload them in three separate publish artifact tasks. That's a waste of time.

- All commands use `BuildConfiguration.ArtifactStagingDirectory` for build artifact publishing and consumption. All input/output paths are relative to this directory. This is through a new abstraction, `ArtifactService`.
- Publishes all artifacts just one time at the very end of the publish stage.


src/ImageBuilder/ArtifactService.cs
```diff
@@ -0,0 +1,55 @@
...
+        _fileSystem.CreateDirectory(outputDirectory);
+        _fileSystem.WriteAllText(outputPath, contents);
+    }
+
+    /// <inheritdoc />
+    public string ResolvePath(string artifactPath)
+    {
+        if (string.IsNullOrWhiteSpace(_buildConfig.ArtifactStagingDirectory))
+        {
+            throw new InvalidOperationException(
```
mthalman, line 34: Any call to `ResolvePath` will fail in the context of a PR build because `generate-appsettings.yml` skips outputting an appsettings.json file for PRs: https://github.com/dotnet/docker-tools/blob/ffecd625b2acdfdcfd85d1005140906b89598b84/eng/docker-tools/templates/steps/generate-appsettings.yml#L34

So no `ArtifactStagingDirectory` value will be set, causing this exception to throw.

lbussell: Fixed in 653d46d

mthalman (approved):

PR #2213, opened 2026-08-10
Allow steps to be injected before ImageBuilder runs

premun:
Backport of https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1729 where we needed to copy `eng/common` in the right time but there was no hook for it yet

### DIFF
--- a/src/Infrastructure/Content/CHANGELOG.md
+++ b/src/Infrastructure/Content/CHANGELOG.md
@@ -4,6 +4,15 @@ All breaking changes and new features in `eng/docker-tools` will be documented i
 
 ---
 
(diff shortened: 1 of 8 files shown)


pavel-purma (approved):

lbussell:
@premun the addition of this hook is probably fine, but would https://github.com/dotnet/docker-tools/issues/1592 have eliminated the need for this? It would allow you to set build context to the root of the repo, for example.

premun:
@lbussell probably yes, the original problem was we could not copy the root-level `eng/common`.

lbussell (approved):

PR #2218, opened 2026-08-28
Use central package management for source projects

lbussell:
Package versions under `src` were declared independently in each project, which made shared dependency updates easy to miss.

This PR enables NuGet central package management for all source projects through `src/Directory.Packages.props`.

- Centralizes every direct package version under `src`
- Updates the shared Microsoft.Extensions and System package versions from 10.0.10 to 10.0.11
- Keeps packages on independent release tracks at their existing versions
- Adds package source mappings for the public and engineering feeds
- Copies the central package file into the Linux and Windows ImageBuilder container restore layers

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -36,8 +36,13 @@
     <PackageVersion Include="Polly.RateLimiting" Version="8.7.0" />
     <PackageVersion Include="Shouldly" Version="4.3.0" />
     <PackageVersion Include="System.CommandLine" Version="2.0.10" />
+
+    <!-- Resolve diamond dependency: Microsoft.TeamFoundationServer.Client transitively requires >= 8.0.0 -->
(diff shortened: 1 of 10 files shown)


mthalman (commented):
Wouldn't CPM be the more appropriate solution for this?

lbussell:
> Wouldn't CPM be the more appropriate solution for this?

1. Central package version management was historically blocked by https://github.com/dotnet/docker-tools/issues/1592. Since we wanted to have a Directory.Packages.props at the root of the repo, the Dockerfile in src/ couldn't resolve it. We could probably revisit it though, since all C# files and projects were recently rearranged to be under the `src/` directory.
2. This orthogonal to CPM. We would still have this property even if we adopted CPM. I'm trying to force all the Microsoft.Extensions.* and System.* packages to version together.

mthalman (approved):

lbussell:
Added CPM.

mthalman (approved):

Issue #1658, opened 2025-03-31
Remove dependency on Microsoft.DotNet.VersionTools


mthalman:
The GitHub functionality in Microsoft.DotNet.VersionTools is not being maintained past 9.0. We also run into issues that have required changes to this package for Docker scenarios. It would be best to remove this dependency for these reasons.

This package is being used by ImageBuilder and the update-dependencies tools of dotnet-docker and dotnet-framework-docker.

lbussell:
Also used in FilePusher: https://github.com/dotnet/docker-tools/blob/3e6ff3e63b45a28263a1b596b2a0e6f71c588f4d/eng/src/file-pusher/FilePusher.cs#L17

PR #2222, opened 2026-09-01
Add GitAutomation.AzureDevOps library

lbussell:
This PR adds a minimal Azure DevOps HTTP client library. The purpose is to make adding Azure DevOps support to the GitAutomation library as simple as possible.

I chose to implement a new HTTP client library instead of using `Microsoft.TeamFoundationServer.Client` because that library has a huge dependency tree, when we would only need to use a very small part of it. Additionally, the Azure DevOps API is extremely stable.

Part of:

- dotnet/dotnet-docker#6345
- #1658

### DIFF
--- a/src/GitAutomation.AzureDevOps/AzureDevOpsClient.cs
+++ b/src/GitAutomation.AzureDevOps/AzureDevOpsClient.cs
@@ -1,0 +1,210 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Net.Http.Json;
+
+namespace Microsoft.DotNet.GitAutomation.AzureDevOps;
+
+/// <summary>Calls the Azure DevOps Git HTTP API.</summary>
+/// <remarks>The caller owns <paramref name="httpClient"/>.</remarks>
+/// <param name="httpClient">The configured HTTP client.</param>
+public sealed class AzureDevOpsClient(HttpClient httpClient) : IDisposable
+{
+    private readonly HttpClient _httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+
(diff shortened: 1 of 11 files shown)


src/GitAutomation.AzureDevOps/AzureDevOpsClient.cs
```diff
@@ -0,0 +1,214 @@
...
+    public async Task<AzureDevOpsPullRequest[]> ListActivePullRequestsAsync(
+        string repository,
+        string sourceRefName,
+        CancellationToken cancellationToken)
+    {
+        string uri = new RequestUriBuilder(repository)
+            .AppendPath("pullrequests")
+            .AddQueryParameter("searchCriteria.status", "active")
+            .AddQueryParameter("searchCriteria.sourceRefName", sourceRefName)
+            .Build();
```
mthalman, line 67: The model types being used here are assuming that the response is fully populated but that's not the case for this endpoint. It only returns partial data. The main problem here is the `AzureDevOpsRepository.RemoteUrl` and `AzureDevOpsRepository.WebUrl` properties. These are marked non-nullable but won't be populated from this call. You can see these fields don't exist in the output of this, for example: https://dev.azure.com/dnceng/internal/_apis/git/pullrequests?searchCriteria.status=active&$top=100&api-version=7.1

lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/2222/commits/4b24572a38170605c161a27bf72800e2fcde3288

lbussell:
I removed the `AzureDevOps` prefix from the model types, because 1) some names were getting unreasonably long, and 2) the `GitAutomation.AzureDevOps` namespace prevents conflicts downstream.

After double-checking the rest of the API models, I also found that committer email address can be null if 1) the commit doesn't have an email address, or 2) if the http requests are unauthenticated. I updated the model accordingly.

mthalman (approved):

PR #2223, opened 2026-09-01
Make GitAutomation interface GitHub agnostic

lbussell:
This PR reworks the GitAutomation API to be provider neutral before adding Azure DevOps support. Trying to retrofit the provider-agnostic requirements on to the previous implementation resulted in a lot of tension because we support forks on GitHub but not on Azure DevOps. The boundary had to be reworked a little bit.

All remote operations are now behind a new `IPullRequestEndpoint` interface. `Git` and `GitContext` are now also behind one `GitWorkingCopy` implementation.

Part of:
- dotnet/dotnet-docker#6345
- #1658


src/GitAutomation/GitHub/OctokitExtensions.cs
```diff
@@ -12,4 +12,10 @@ internal static class OctokitExtensions
     public static bool IsValid([NotNullWhen(true)] this AccessToken? accessToken) =>
         accessToken is not null
         && accessToken.ExpiresAt > DateTimeOffset.UtcNow.AddMinutes(1);
+
+    extension(CommitInfo)
+    {
+        public static CommitInfo FromPullRequestCommit(PullRequestCommit prCommit) =>
+            new CommitInfo(prCommit.Sha, prCommit.Author.Name, prCommit.Author.Email);
```
mthalman: Apparently the recommendation is to use `prCommit.Commit.Author` instead of `prCommit.Author`. You can see in Octokit's own tests that they don't populate Name/Email for `prCommit.Author`: https://github.com/octokit/octokit.net/blob/main/Octokit.Tests/Models/PullRequestCommitTests.cs.

FWIW, the previous implementation of this code used the recommended way: https://github.com/dotnet/docker-tools/blob/add1f10eaba9fece6e2d97c746f5cf439205b71b/src/GitAutomation/GitHub/GitHubRepoHost.cs#L64

lbussell: Yep. Thought I could simplify it! Guess not. I will make sure this gotcha is captured in a comment.

mthalman (approved):

mthalman (approved):

PR #2224, opened 2026-09-01
Add Azure DevOps pull request support to GitAutomation library

lbussell:
The prior two changes in this stack exist to make this implementation as simple as possible. Most of the changes here are tests and a README rewrite.

Part of:

- dotnet/dotnet-docker#6345
- #1658


src/GitAutomation/AzureDevOps/AzureDevOpsPullRequestEndpoint.cs
```diff
@@ -0,0 +1,184 @@
...
+            force,
+            cancellationToken);
+    }
+
+    /// <inheritdoc/>
+    public async Task<ExistingPullRequest?> FindPullRequestAsync(string key, CancellationToken cancellationToken)
+    {
+        string sourceRefName = GetRefName(key);
+        AzureDevOpsPullRequest[] pullRequests =
+            await _client.ListActivePullRequestsAsync(_repositoryIdOrName, sourceRefName, cancellationToken);
```
mthalman, line 79: The `list` endpoint shouldn't be relied upon for the PR description. See https://learn.microsoft.com/rest/api/azure/devops/git/pull-requests/get-pull-requests?view=azure-devops-rest-7.1&tabs=HTTP:

> Please note that description field will be truncated up to 400 symbols in the result.

If the description is over 400 chars, it will cause the planner to update it because it doesn't match the actual value: https://github.com/dotnet/docker-tools/blob/463db626faef12e50e6942ede06eac9d60afa95d/src/GitAutomation/Planner.cs#L62-L66

lbussell: Fixed in https://github.com/dotnet/docker-tools/pull/2224/commits/721e7885719d81bf2a3e0f226dfdc45552476304 and https://github.com/dotnet/docker-tools/pull/2222/commits/b4520a083f26ef257704fcf1312fb7022a458efa (further down the stack)

mthalman (approved):

mthalman (approved):

mthalman (approved):

PR #2238, opened 2026-09-15
Thread CancellationToken through all async methods

lbussell:
While doing some other work, I got annoyed that pressing Ctrl-C didn't properly stop ImageBuilder.

This PR addresses the issue comprehensively by passing CancellationToken through *all* commands and async methods.


src/ImageBuilder/AcrImageImporter.cs
```diff
@@ -37,15 +37,18 @@ public AcrImageImporter(
...
         var armClient = GetArmClientForAcr(destAcrName);
         var registryResource = armClient.GetContainerRegistryResource(destResourceId);
 
         try
         {
             await RetryHelper.GetWaitAndRetryPolicy<Exception>(_logger)
-                .ExecuteAsync(() => registryResource.ImportImageAsync(WaitUntil.Completed, importContent));
+                .ExecuteAsync(
+                    ct => registryResource.ImportImageAsync(WaitUntil.Completed, importContent, ct),
+                    cancellationToken);
```
mthalman, line 51: I think there's a danger here that cancellation exception would cause a retry because of the policy.

lbussell: Polly cancels further retries when cancellation is requested for the token.

From [the Polly documentation](https://github.com/App-vNext/Polly/blob/8.7.0/src/Polly/AsyncPolicy.ExecuteOverloads.cs#L36-L44):

> [describing `ExecuteAsync`'s `cancellationToken` parameter]
> A cancellation token which can be used to cancel the action.  When a retry policy in use, also cancels any further retries.

mthalman (approved):

